PR #2, opened 2017-02-03
Add travis CI

feiskyer:
This PR adds initial verify scripts and travis CI.

nit: It will fail because there is no go files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,16 @@ go:
   - 1.8.x
   - tip
 
-install:
+
+before_install:
   - go get -u github.com/alecthomas/gometalinter
+
+
+install:
   - gometalinter --install
 
+
 script:
-  - hack/verify-all.sh -v
+  - hack/repo-infra/verify/verify-go-src.sh -r $(pwd) -v
+  - hack/repo-infra/verify/verify-boilerplate.sh 
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: go
+sudo: required
+dist: trusty
+
+go:
+  - 1.7.x
+  - 1.8.x
+  - tip
+
+install:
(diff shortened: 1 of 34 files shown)


xlgao-zju:
Do we need `golint` in CI?

feiskyer:
> Do we need golint in CI?

Of course. Added in the latest commit.

xlgao-zju:
Cool

hack/verify-all.sh
```diff
@@ -0,0 +1,88 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SILENT=true
+KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
```
xlgao-zju: Maybe `CRI_TOOLS_ROOT` rather than `KUBE_ROOT`?

feiskyer:
@Random-Liu @yujuhong Travis enabled. PTAL.

mrunalp:
LGTM

xlgao-zju:
Ping @Random-Liu

ideahitme:
@here please consider using https://github.com/kubernetes/repo-infra/tree/master/verify to avoid code duplication

feiskyer:
@ideahitme Thanks. Updated to repo-infra.

ideahitme:
@feiskyer normally you would do it via vendoring or git-submodule, would be easier for you to upgrade and manage it in the future. Take a look at https://github.com/kubernetes-incubator/external-dns/ for reference

feiskyer:
@ideahitme Yep, vendor is more easier to upgrade but godeps doesn't support manage none-golang packages. So this PR just get it via `git subtree`.

PR #7, opened 2017-02-28
Fix ginkgo not found, when we do not have ginkgo in GOPATH

xlgao-zju:
cc @feiskyer 


feiskyer:
Thanks. LGTM

PR #8, opened 2017-03-03
Add pod basic test case

xlgao-zju:

### DIFF
--- a/pkg/framework/util.go
+++ b/pkg/framework/util.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	internalapi "k8s.io/kubernetes/pkg/kubelet/api"
+	runtimeapi "k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
 	"k8s.io/kubernetes/pkg/kubelet/remote"
 
 	. "github.com/onsi/ginkgo"
@@ -30,7 +31,10 @@ import (
 )
 
 const (
-	defaultAPIVersion string = "0.1.0"
+	defaultAPIVersion      string = "0.1.0"
+	defaultUIDPrefix       string = "e2e-cri-uid"
+	defaultNamespacePrefix string = "e2e-cri-namespace"
+	defaultAttempt         uint32 = 2
 )
 
 var (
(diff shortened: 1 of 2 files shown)


xlgao-zju:
@feiskyer @Random-Liu

pkg/validate/pod.go
```diff
@@ -32,7 +33,46 @@ var _ = framework.KubeDescribe("Test CRI PodSandbox", func() {
 		c = f.CRIClient.CRIRuntimeClient
 	})
 
-	It("test get version info", func() {
+	It("test get version info [Conformance]", func() {
```
feiskyer: We should make this a complete sentence (and all followings), e.g. `runtime should return version info`. And `Version()` should be outside of sandbox context.

pkg/framework/util.go
```diff
@@ -105,7 +109,111 @@ func NewUUID() string {
 
 // TestGetVersion test if we can get runtime name.
 func TestGetVersion(c internalapi.RuntimeService) {
-	version, err := c.Version(defaultAPIVersion)
+	_, err := c.Version(defaultAPIVersion)
```
feiskyer: Should check the output, e.g. all response strings should not be null.

pkg/framework/util.go
```diff
@@ -105,7 +109,111 @@ func NewUUID() string {
...
+	ExpectNoError(err, "failed to create PodSandbox: %v", err)
+	Logf("Created PodSandbox %q\n", podID)
+	return podID
+}
+
+// RunDefaultPodSandbox runs a PodSandbox with default options.
+func RunDefaultPodSandbox(c internalapi.RuntimeService, prefix string) string {
+	podSandboxName := prefix + NewUUID()
+	config := &runtimeapi.PodSandboxConfig{
+		Metadata: buildPodSandboxMetadata(podSandboxName, defaultUID, defaultNamespace, defaultAttempt),
```
feiskyer: Add random to sandbox ID/name/namespace, so the test cases could be easily paralleled later.

xlgao-zju:
@feiskyer updated

feiskyer:
LGTM.

Issue #4, opened 2017-02-23
Umbrella issues for validation tests


xlgao-zju:
Here is the list of functionality test, based on https://github.com/kubernetes-incubator/cri-tools/issues/4#issue-209695062 and https://github.com/kubernetes-incubator/cri-tools/issues/4#issuecomment-282947338:
- [x] P0 Basic (Conformance):
  - [x] sandbox: run/stop/remove/list/status #8 
  - [x] container: create/start/execSync/remove/status #13 
  - [x] image: pull/list/status/remove #9 
  - [x] version #8 
  - [x] status #11 
  - [x] logs @Helen-Xie #28 
  - [x] networking: DNSOptions/PortMapping/IP @xlgao-zju #30 
  - [x] volumes #27 
- [x] P0 SecurityContext (only subset of this function are testable): 
  - [x] container with security context:
    - [x] NamespaceOption
      - [x] hostPID #56 
      - [x] hostIPC #82 
      - [x] hostNetwork #75 
    - [x] Priviledged #78 
    - [x] Capabilities #80 
    - [x] RunAsUser/RunAsUserName #41 
    - [x] ReadOnlyRootfs #53 
- [x] P1 Streaming (Conformance): 
  - [x] exec #58 
  - [x] attach #64 
  - [x] portfoward #71 
- [ ] P2 Resource Limits
- [ ] P2 updateRuntimeConfig
- [ ] P2 Experimental: apparmor/sysctls/seccomp
- [ ] P2 Devices
- [ ] P2 SELinuxOption

See [more details](https://docs.google.com/document/d/1usdMWxOXvKIQ00UPRYQWNKY4jeL66FpbHnL7dz0m6qc/edit?usp=sharing) about each test.

The list may not be sufficient enough. Please feel free to:
* Point out anything is missing or incorrect.
* Claim any test you'd like to help.

/cc @feiskyer @resouer @xlgao-zju @Helen-Xie @yujuhong @dchen1107 @mrunalp @lucab

resouer:
Move discussion to https://github.com/kubernetes-incubator/cri-tools/pull/6

xlgao-zju:
@resouer Yes, will keep this issue only for tracking PRs about test case.

Random-Liu:
@feiskyer @resouer @xlgao-zju @Helen-Xie
Here is a document for the CRI function validation test:
[...]

PR #9, opened 2017-03-06
Add image basic test case

Helen-Xie:
Please review the image basic test. @feiskyer @Random-Liu 

### DIFF
--- a/pkg/framework/util.go
+++ b/pkg/framework/util.go
@@ -216,7 +216,7 @@ func ListPodSanboxForID(c internalapi.RuntimeService, podID string) []*runtimeap
 	return ListPodSandbox(c, filter)
 }
 
-// ListPodSandbox list PodSandbox.
+// ListPodSandbox lists PodSandbox.
 func ListPodSandbox(c internalapi.RuntimeService, filter *runtimeapi.PodSandboxFilter) []*runtimeapi.PodSandbox {
 	By("List PodSandbox")
 	pods, err := c.ListPodSandbox(filter)
--- a/pkg/validate/image.go
+++ b/pkg/validate/image.go
@@ -1,0 +1,222 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validate
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/kubernetes-incubator/cri-tools/pkg/framework"
+	internalapi "k8s.io/kubernetes/pkg/kubelet/api"
+	runtimeapi "k8s.io/kubernetes/pkg/kubelet/api/v1alpha1/runtime"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var (
+	// image name for test image api
+	testImageName = "busybox"
+
+	// name-tagged reference for test image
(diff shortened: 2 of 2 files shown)


Helen-Xie:
I signed it!

Helen-Xie:
#4

pkg/framework/util.go
```diff
@@ -216,11 +217,94 @@ func ListPodSanboxForID(c internalapi.RuntimeService, podID string) []*runtimeap
...
+// ListPodSandbox lists PodSandbox.
 func ListPodSandbox(c internalapi.RuntimeService, filter *runtimeapi.PodSandboxFilter) []*runtimeapi.PodSandbox {
 	By("List PodSandbox")
 	pods, err := c.ListPodSandbox(filter)
 	ExpectNoError(err, "failed to list PodSandbox status: %v", err)
 	Logf("List PodSandbox succeed")
 	return pods
 }
+
+// PullPublicImage pulls the public image named imageName.
```
feiskyer: Could we split those func into separate files?

pkg/validate/image.go
```diff
@@ -0,0 +1,106 @@
...
+
+var (
+	// image name for test image api
+	testImageName = "busybox"
+
+	// name-tagged reference for latest test image
+	latestTestImageRef = testImageName + ":latest"
+
+	// Digested reference for test image
+	busyboxDigestRef = testImageName + "@sha256:a59906e33509d14c036c8678d687bd4eec81ed7c4b8ce907b888c607f6a1e0e6"
```
feiskyer: digest may change over time for image with latest tag. It's better to use a dedicated tag (e.g. 1.26.2) instead of latest.

pkg/validate/image.go
```diff
@@ -0,0 +1,106 @@
...
+
+		defer framework.RemoveImage(c, latestTestImageRef)
+
+		status := framework.ImageStatus(c, latestTestImageRef)
+		Expect(status.Id).NotTo(BeNil(), "image Id should not be nil")
+		Expect(len(status.RepoTags)).NotTo(Equal(0), "should have repoTags in image")
+		Expect(status.Size_).NotTo(BeNil(), "image Size should not be nil")
+	})
+
+	It("ListImage should get exactly 3 image in the result list [Conformance]", func() {
```
feiskyer: It may fail if runtime isn't a fresh install. I think it's better to check those images are included in the image list.

pkg/validate/image.go
```diff
@@ -0,0 +1,106 @@
...
+
+		framework.PullImageList(c, testImageList)
+
+		defer framework.RemoveImageList(c, testImageList)
+
+		images := framework.ListImages(c, testImageName)
+		Expect(len(images)).To(Equal(3), "should have exactly three images in list")
+	})
+
+	It("ListImage should get exactly 3 repoTags in the result image [Conformance]", func() {
```
feiskyer: same here.

Helen-Xie:
Update! @feiskyer

pkg/validate/image.go
```diff
@@ -0,0 +1,219 @@
...
+			"busybox:1-uclibc",
+			"busybox:1-musl",
+			"busybox:1-glibc",
+		}
+
+		pullImageList(c, testImageList)
+
+		defer removeImageList(c, testImageList)
+
+		images := listImages(c, testImageName)
```
feiskyer: This only list an image with `latest` tag, which will fail the test always.

Helen-Xie: I have test it, it can be pass, and it will list all the images named "busybox".

feiskyer: Maybe dockershim is implemented that way. But usually, an image without tag is treated as a `latest` one (e.g. containers are created by using latest tag ).

Helen-Xie: ah, should I specify the tag to be listed or list all the images?

feiskyer: I think list and filter out dedicated is ok (only one API call instead of three).

Helen-Xie: ok, I will update it later.

Helen-Xie:
Please review it again. Thanks. @feiskyer

feiskyer:
Thanks. LGTM.

PR #14, opened 2017-03-14
Mv version and status test into runtime_info.go

xlgao-zju:
cc @feiskyer 


pkg/validate/runtime_info.go
```diff
@@ -21,32 +21,24 @@ import (
 	internalapi "k8s.io/kubernetes/pkg/kubelet/api"
```
feiskyer: I think it's better to also move `TestGetRuntimeStatus ` and `TestGetVersion` into `runtime_info.go`

xlgao-zju:
@feiskyer Updated

PR #15, opened 2017-03-15
Add make lint

xlgao-zju:
cc @feiskyer 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,14 @@ install: check-gopath
 uninstall:
 	rm -f $(BINDIR)/critest
 
+lint:
+	./hack/repo-infra/verify/go-tools/verify-gometalinter.sh
+
+install.tools:
+	go get -u github.com/alecthomas/gometalinter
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -48,11 +48,23 @@ install: check-gopath
 uninstall:
 	rm -f $(BINDIR)/critest
 
+lint:
+	./hack/repo-infra/verify/go-tools/verify-gometalinter.sh
+
+install.tools: .install.gometalinter
+
+.install.gometalinter:
+	go get -u github.com/alecthomas/gometalinter
```
feiskyer: why not make these two commands under `install.tools` directly?

xlgao-zju: For adding other tools in the future. If you think we will not need any other tool, I'm fine with making these two commands under install.tools directly.

xlgao-zju:
@feiskyer @mrunalp Updated and make the commands under install.tools directly

PR #29, opened 2017-03-21
Initial code for crictl

mrunalp:
@feiskyer PTAL

This is adding basic code for crictl. I will follow up with adding support for other commands.

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -36,6 +36,10 @@ func main() {
 	app.Usage = "client for CRI"
 	app.Version = "0.0.1"
 
+	app.Commands = []cli.Command{
+		runtimeVersionCommand,
+	}
+
 	app.Flags = []cli.Flag{
 		cli.StringFlag{
 			Name:  "runtime-endpoint",
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/urfave/cli"
+	"google.golang.org/grpc"
+)
+
+const (
+	defaultTimeout = 10 * time.Second
+)
+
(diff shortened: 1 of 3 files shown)


cmd/crictl/main.go
```diff
@@ -0,0 +1,66 @@
+package main
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"time"
+
+	"github.com/Sirupsen/logrus"
```
feiskyer: replace with `github.com/golang/glog`? we used glog in other places.

cmd/crictl/version.go
```diff
@@ -0,0 +1,41 @@
...
+	},
+}
+
+// Version sends a VersionRequest to the server, and parses the returned VersionResponse.
+func Version(client pb.RuntimeServiceClient, version string) error {
+	r, err := client.Version(context.Background(), &pb.VersionRequest{Version: version})
+	if err != nil {
+		return err
+	}
+	fmt.Printf("VersionResponse: Version: %s, RuntimeName: %s, RuntimeVersion: %s, RuntimeApiVersion: %s\n", r.Version, r.RuntimeName, r.RuntimeVersion, r.RuntimeApiVersion)
```
feiskyer: Add a `\n` for each ?

cmd/crictl/main.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/urfave/cli"
+	"google.golang.org/grpc"
+)
+
+const (
+	DefaultTimeout = 10 * time.Second
```
feiskyer: I think `DefaultTimeout` doesn't need to be exported, it also breaks validations because of no comments.

mrunalp: Fixed.

cmd/crictl/version.go
mrunalp: Fixed.

cmd/crictl/main.go
mrunalp: Fixed.

feiskyer:
Cool. LGTM

PR #30, opened 2017-03-23
Add networking test

### LINKED ISSUE #17: critest: DNSOptions, PortMapping and IP

xlgao-zju:
close #17 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,9 @@ uninstall:
 lint:
 	./hack/repo-infra/verify/go-tools/verify-gometalinter.sh
 
+gofmt:
+	./hack/repo-infra/verify/go-tools/verify-gofmt.sh
+
 install.tools:
 	go get -u github.com/alecthomas/gometalinter
 	gometalinter --install
@@ -72,4 +75,5 @@ install.tools:
 	install \
 	uninstall \
 	lint \
+	gofmt \
 	install.tools
--- a/pkg/validate/container.go
+++ b/pkg/validate/container.go
@@ -257,6 +257,7 @@ func createDefaultContainer(rc internalapi.RuntimeService, ic internalapi.ImageM
 		Metadata: buildContainerMetadata(containerName, defaultAttempt),
 		Image:    &runtimeapi.ImageSpec{Image: defaultContainerImage},
 		Command:  []string{"sh", "-c", "top"},
+		Linux:    &runtimeapi.LinuxContainerConfig{},
 	}
 
 	return createContainer(rc, ic, containerConfig, podID, podConfig)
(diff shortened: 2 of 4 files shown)


Random-Liu:
I think the problem is that you should not create container with empty linux configuration.

We point the container to the infra container [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/security_context.go#L162).

However, if the LinuxConfiguration is empty, dockershim will skip all those config steps directly, see [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_container.go#L147-L166).

xlgao-zju:
Yes, I think empty LinuxConfiguration will cause creating pod to fail, when I use cni. 
The error like this.
```
NetworkPlugin cni failed to set up pod \"PodSandbox-for-create-test-77c938c8-13b4-11e7-b51d-4ccc6a538d8b_e2e-cri-namespace77c9390f-13b4-11e7-b51d-4ccc6a538d8b\" network: container veth name provided (eth0) already exists
``` 

I will try to figure out a solution.

xlgao-zju:
Do you have any idea on constructing the field `CgroupParent` in `LinuxConfiguration`. `kuberuntime` will do this, but if we test the shim directly, we should construct it before we send the request.

xlgao-zju:
File [k8s/#43890](https://github.com/kubernetes/kubernetes/pull/43890) in order to avoid panic in dockershim when we create pod with empty `NamespaceOption`.

xlgao-zju:
Because we always use default dns options [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker_manager.go#L1905), even if I send other options.

So, do not test dns options for now.

/cc @feiskyer @Random-Liu

xlgao-zju:
@feiskyer @Random-Liu This pr is ready for review now. :)

Random-Liu:
> Because we always use default dns options here, even if I send other options.
> So, do not test dns options for now.

I don't understand. Dockershim doesn't generate default dns option. It generates dns option based on `DNSOption` and `DNSSearch` passed, right?

Random-Liu:
Sorry, I get what you mean. Let me see.

Random-Liu:
Filed https://github.com/kubernetes/kubernetes/pull/43925 to fix this.

xlgao-zju:
@Random-Liu Cool

xlgao-zju:
@Random-Liu Add test about DNS options, review plz. :)

feiskyer:
LGTM

xlgao-zju:
@Random-Liu Would you like to take another look?

PR #43, opened 2017-04-10
tiny fix

xlgao-zju:


xlgao-zju:
Since this is a very tiny fix, so merge it now.

PR #48, opened 2017-04-11
Add lisence

xlgao-zju:
cc @feiskyer 


xlgao-zju:
Close this in favor of #49

PR #52, opened 2017-04-12
Use after and before to aviod duplicated code

xlgao-zju:
cc @feiskyer 


feiskyer:
Cool. LGTM

Issue #20, opened 2017-03-17
critest: container security context


feiskyer:
critest for container security context:

- [ ] NamespaceOption
  - [x] hostPID
  - [ ] hostIPC #82 
  - [x] hostNetwork #75 
- [x] Priviledged #78 
- [ ] Capabilities
- [x] RunAsUser/RunAsUserName #41 
- [x] ReadOnlyRootfs

feiskyer:
Ideas of how to test hostIPC by IPC shared memory

```sh
# create shared memory segment on the host
$ ipcmk -M 1M
Shared memory id: 98307



# the shared memory segment should be included in containers with --ipc=host
$ sudo docker run -it --rm --ipc=host busybox ipcs -m

------ Shared Memory Segments --------
key        shmid      owner      perms      bytes      nattch     status
0x00000000 0          root       644        80         2
0x00000000 32769      root       644        16384      2
0x00000000 65538      root       644        280        2
0xa3d06fda 98307      1000       644        1048576    0



# the shared memory segment should not be included in containers without --ipc=host
$ sudo docker run -it --rm busybox ipcs -m

------ Shared Memory Segments --------
key        shmid      owner      perms      bytes      nattch     status

```

xlgao-zju:
@feiskyer Create a memory segment and get the id `98307`. Then create a container with --ipc=host, exec in this container with cmd `ipcs -m`. Finally check if the output contains `98307`. 
WDYT?

feiskyer:
@xlgao-zju LGTM. Both positive and negative cases are required.

PR #78, opened 2017-04-25
add Privileged test case

Helen-Xie:
part of #20 

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -279,12 +279,49 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			By("Check whether rootfs is read-only")
 			checkRootfs(podConfig, logPath, readOnlyRootfs)
 		})
+
+		It("runtime should support Privileged is true", func() {
+			By("create pod")
+			isPrivileged := true
+			podID, podConfig = createPrivilegedPodSandbox(rc, isPrivileged)
+
+			By("create container for security context Privileged is true")
+			containerID, _ := createPrivilegedContainer(rc, ic, podID, podConfig, "container-with-isPrivileged-test-", isPrivileged)
+
+			By("start container")
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+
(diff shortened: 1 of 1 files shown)


Helen-Xie:
ping @feiskyer

pkg/validate/security_context.go
```diff
@@ -279,6 +279,42 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			By("Check whether rootfs is read-only")
 			checkRootfs(podConfig, logPath, readOnlyRootfs)
 		})
+
+		It("runtime should support Privileged is true", func() {
+			By("create pod")
+			podID, podConfig = createPodSandboxForContainer(rc)
```
feiskyer: Sandbox should also privileged for this case.

pkg/validate/security_context.go
```diff
@@ -428,3 +464,38 @@ func compareNetworkList(podConfig *runtimeapi.PodSandboxConfig, logPath string,
...
+	reverseCmd := []string{"ip", "link", "del", "dummy0"}
+
+	stdout, stderr, err := rc.ExecSync(containerID, cmd, time.Duration(defaultExecSyncTimeout)*time.Second)
+	msg := fmt.Sprintf("cmd %v, stdout %q, stderr %q", cmd, stdout, stderr)
+
+	if isPrivileged {
+		Expect(err).NotTo(HaveOccurred(), msg)
+
+		// We need to clean up the dummy link that was created.
+		execSyncContainer(rc, containerID, reverseCmd)
```
feiskyer: I think this is not required since we will remove the container/sandbox after each test case.

Helen-Xie: ok, I will update it.

feiskyer:
ReadOnlyRootfs and HostNetwork test failed randomly.

Helen-Xie:
update. ping @feiskyer 

the command execution time is too short, so func `Eventually()` of the containerStatus of the ReadOnlyRootfs and HostNetwork test case will failed randomly. I think there is not required `Eventually()`, because there is no need to keep the container always `running`.

xlgao-zju:
@feiskyer wired, exec test failed, and get passed when I restart it...
I think there is nothing wrong with exec test, the failure is [line](https://github.com/kubernetes-incubator/cri-tools/blob/master/pkg/validate/streaming.go#L148).

feiskyer:
LGTM

PR #85, opened 2017-05-03
Add performance test framework

xlgao-zju:
- Add performance test framework
- Add a few test cases about PodSandbox

One thing to be discussed: How many times of the operations are suitable? 20(current)? 30? Or more?

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,13 @@ go get github.com/kubernetes-incubator/cri-tools/cmd/critctl
 
 - `--image-endpoint`, `-i`:CRI server image endpoint, default same as runtime endpoint.
 
-- `--build-dependencies`, `-b`:If we don't need to build dependencies, we can add this flag `--build-dependencies=false` or `-b=false`.
+- `--compile`, `-c`:If we don't need to build dependencies, we can add this flag `--compile=false` or `-c=false`.
 
 - `--ginkgo-flags`,`-g`:Space-separated list of arguments to pass to Ginkgo test runner.
 
 - `--focus`,`-f`:CRI e2e test will only run the test that match the focus regular expression.
+
+- `--benchmark`,`-b`:If set, critest will only run benchmark.
 
 ### crictl
 
--- a/cmd/critest/util.go
+++ b/cmd/critest/util.go
@@ -29,23 +29,29 @@ import (
 
 var buildTargets = []string{
 	"github.com/kubernetes-incubator/cri-tools/pkg/validate",
+	"github.com/kubernetes-incubator/cri-tools/pkg/benchmark",
 	"github.com/kubernetes-incubator/cri-tools/vendor/github.com/onsi/ginkgo/ginkgo",
 }
 
 // build cri-tools binaries.
-func build() error {
+func build(benchmark bool) error {
 	glog.Infof("Building cri-tools binaries...")
 	outputDir, err := getBuildOutputDir()
 	if err != nil {
 		glog.Fatalf("Failed to get build output directory: %v", err)
 	}
 
-	err = runCommand("go", "test", "-c", "-v", "-o", filepath.Join(outputDir, "e2e.test"), buildTargets[0])
-	if err != nil {
-		return fmt.Errorf("failed to build e2e.test %v", err)
+	args := []string{"test", "-c", "-v", "-o", filepath.Join(outputDir, "e2e.test"), buildTargets[0]}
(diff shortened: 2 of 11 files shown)


xlgao-zju:
Ping @feiskyer

README.md
```diff
@@ -63,12 +63,14 @@ go get github.com/kubernetes-incubator/cri-tools/cmd/critctl
...
 - `--image-endpoint`, `-i`:CRI server image endpoint, default same as runtime endpoint.
 
-- `--build-dependencies`, `-b`:If we don't need to build dependencies, we can add this flag `--build-dependencies=false` or `-b=false`.
+- `--compile`, `-c`:If we don't need to build dependencies, we can add this flag `--compile=false` or `-c=false`.
 
 - `--ginkgo-flags`,`-g`:Space-separated list of arguments to pass to Ginkgo test runner.
 
 - `--focus`,`-f`:CRI e2e test will only run the test that match the focus regular expression.
 
+- `--benchmark`,`-b`:If set, critest will run benchmark.
```
feiskyer: only run benchmark?

cmd/critest/util.go
```diff
@@ -29,23 +29,31 @@ import (
...
 	glog.Infof("Building cri-tools binaries...")
 	outputDir, err := getBuildOutputDir()
 	if err != nil {
 		glog.Fatalf("Failed to get build output directory: %v", err)
 	}
 
-	err = runCommand("go", "test", "-c", "-v", "-o", filepath.Join(outputDir, "e2e.test"), buildTargets[0])
-	if err != nil {
-		return fmt.Errorf("failed to build e2e.test %v", err)
+	if benchmark {
```
feiskyer: It's better to add a variable of build targets (which is set depending on benchmark), and use same `runCommand` for it.

README.md
xlgao-zju: yes, I think it's unnessery to run both benchmark and e2e.

feiskyer: OK, then add `only` to flag docs.

pkg/benchmark/pod.go
```diff
@@ -0,0 +1,90 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package benchmark
+
+import (
+	"github.com/kubernetes-incubator/cri-tools/pkg/framework"
+	//"github.com/kubernetes-incubator/cri-tools/pkg/validate"
```
feiskyer: remove unused imports

README.md
xlgao-zju: will do

pkg/benchmark/pod.go
```diff
@@ -0,0 +1,90 @@
...
+
+			operation := b.Time("stop PodSandbox", func() {
+				err = c.StopPodSandbox(podID)
+			})
+
+			framework.ExpectNoError(err, "failed to stop PodSandbox: %v", err)
+			Expect(operation.Seconds()).Should(BeNumerically("<", 2), "stop PodSandbox shouldn't take too long.")
+
+			c.RemovePodSandbox(podID)
+		}, 20)
```
feiskyer: We could add a flag for this magic number

cmd/critest/util.go
xlgao-zju: `runCommand` is not only for building test, but also used to run other commands like building and running ginkgo. So, is it fine to add a parameter only for building test to a common function?

pkg/benchmark/pod.go
xlgao-zju: How many times of the operations are suitable?

feiskyer: I think defaulting 20 is ok

cmd/critest/util.go
feiskyer: No, I mean 

```go

args := []string{"test", "-c", "-v", "-o", filepath.Join(outputDir, "e2e.test"), buildTargets[0]}
if benchmark {
	args = []string{"test", "-c", "-v", "-o", filepath.Join(outputDir, "benchmark.test"), buildTargets[1]}
}

err = runCommand("go", args...)
if err != nil {
	return fmt.Errorf("build test suite failed: %v", err)
}
```

xlgao-zju:
Updated @feiskyer

feiskyer:
Thanks. LGTM.

PR #86, opened 2017-05-03
Add GinkgoRecover() to the goroutines to prevent panics

yujuhong:
Assertion in goroutines may cause panics that cannot be rescued by
ginkgo. Add GinkgoRecover() to prevent tests panicing.

### DIFF
--- a/pkg/validate/container.go
+++ b/pkg/validate/container.go
@@ -310,6 +310,7 @@ func stopContainer(c internalapi.RuntimeService, containerID string, timeout int
 	stopped := make(chan bool, 1)
 
 	go func() {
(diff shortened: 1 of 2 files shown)


yujuhong:
Failure message:
```
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.

But, if you make an assertion in a goroutine, Ginkgo can't capture the panic.
To circumvent this, you should call

	defer GinkgoRecover()

at the top of the goroutine that caused this panic.
```

feiskyer:
LGTM. Thanks.

Issue #25, opened 2017-03-17
Umbrella issues for benchmark tests


feiskyer:
Umbrella issues for benchmark tests.

- [ ] P0: Parallel Sandbox & containers operation
   - [x] Candidate Operation List (benchmark one operation in one test case):
        - [x] sandbox run #85 
        - [x] sandbox status
        - [x] remove sandbox
        - [x] stop sandbox
        - [x] list sandboxes
        - [x] create container #90 
        - [x] container status
        - [x] start container
        - [x] stop container
        - [x] remove container
        - [x] list container
    - [ ] Metrics:
        - [ ] Node: 
            - [ ] CPU usage (%)
            - [ ] MEM usage (%)
        - [ ] Runtime daemon perf: 
            - pod list latency (ms, 50%ile 90%ile 99%ile)
        - [ ] Core:
            - [ ] operation latency (ns, 50%ile 90%ile 99%ile)
    - [ ] Test Case Variable:
        - [ ] concurrency of operation
- [ ] p0: Image write operation (stick to specific image)
    - [ ] Candidate Operation List (benchmark one operation in one test case):
        - [ ] image pull
        - [ ] image remove
    - [ ] Metrics:
        - [ ] Node: 
            - [ ] CPU usage (%)
            - [ ] MEM usage (%)
            - [ ] DISK IO (%iowait)
            - [ ] NETWORK IO (bandwidth usage: bytes/sec)
    - [ ] Test Case Variable:
        - [ ] repeat number of operation
- [ ] p0: Image read operation (stick to specific image)
    - [ ] Candidate Operation List (benchmark one operation in one test case):
            - [ ] image list
            - [ ] image status
    - [ ] Metrics:
        - [ ] Node: 
            - [ ] CPU usage (%)
            - [ ] MEM usage (%)
        - [ ] Core:
            - [ ] operation latency (ns, 50%ile 90%ile 99%ile)
    - [ ] Test Case Variable:
        - [ ] repeat number of operation
- [ ] p0: Parallel image operation (stick to specific image)
    - [ ] Candidate Operation List (benchmark one operation in one test case):
        - [x] image pull
        - [x] image remove
        - [x] image list
    - [ ] Metrics:
        - [ ] Node: 
            - CPU usage (%)
            - MEM usage (%)
            - DISK IO (%iowait)
            - NETWORK IO (bandwidth usage: bytes/sec)
    - [ ] Test Case Variable:
        - concurrency of operation
[...]

PR #90, opened 2017-05-09
Add Container Operations Benchmark

Helen-Xie:
part of #25  Add a few benchmark test cases about Container

### DIFF
--- a/pkg/benchmark/container.go
+++ b/pkg/benchmark/container.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


Helen-Xie:
ping @feiskyer

pkg/benchmark/container.go
```diff
@@ -0,0 +1,119 @@
...
+		})
+
+		Measure("benchmark about creating Container", func(b Benchmarker) {
+			var containerID string
+
+			operation := b.Time("create Container", func() {
+				containerID = framework.CreateDefaultContainer(rc, ic, podID, podConfig, "Container-for-creating-benchmark-")
+			})
+
+			Expect(operation.Seconds()).Should(BeNumerically("<", 2), "create Container shouldn't take too long.")
```
feiskyer: Should we measure those by ms instead of second?

cc/ @xlgao-zju

pkg/benchmark/container.go
```diff
@@ -0,0 +1,119 @@
...
+
+			rc.StopContainer(containerID, framework.DefaultStopContainerTimeout)
+			rc.RemoveContainer(containerID)
```
feiskyer: create/start/stop/remove has already occurred in every measure cases. Could we merge all those into same Measure block (but different operation)?

Helen-Xie: update

pkg/benchmark/container.go
Helen-Xie: there is no func `Millisecond()` in operation, so I still use the `Seconds()`.

feiskyer:
LGTM

PR #92, opened 2017-05-12
Consolidate hostnetwork testing

feiskyer:
cc @xlgao-zju

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -18,6 +18,7 @@ package validate
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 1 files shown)


xlgao-zju:
@feiskyer 
How about creating a nginx container with `hostnetwork=true`, then reuse this [function](https://github.com/kubernetes-incubator/cri-tools/blob/master/pkg/validate/networking.go#L185) to check if we can get the main page of nginx via localhost:80.

feiskyer:
Fixed port is not a good idea. The port may be conflict with host's existing ports.

pkg/validate/security_context.go
```diff
@@ -616,3 +561,54 @@ func createCapabilityContainer(rc internalapi.RuntimeService, ic internalapi.Ima
...
+		HostNetwork: hostNetwork,
+	}
+	hostPath, podLogPath := createLogTempDir(podSandboxName)
+	podID, podConfig := createNamespacePodSandbox(rc, podSandboxNamespace, podSandboxName, podLogPath)
+
+	defer os.RemoveAll(hostPath) //clean up the TempDir
+
+	By("create a container in the sandbox")
+	command := []string{"sh", "-c", "netstat -ln"}
+	containerNamespace := &runtimeapi.NamespaceOption{
```
xlgao-zju: We can remove `containerNamespace` and rename `podSandboxNamespace`. Since they have same value.

pkg/validate/security_context.go
```diff
@@ -616,3 +561,54 @@ func createCapabilityContainer(rc internalapi.RuntimeService, ic internalapi.Ima
...
+		HostPid:     false,
+		HostIpc:     false,
+		HostNetwork: hostNetwork,
+	}
+
+	prefix := "container-with-HostNetwork-test-"
+	containerName := prefix + framework.NewUUID()
```
xlgao-zju: I think we can just let `containerName := "container-with-HostNetwork-test-" + framework.NewUUID()`

pkg/validate/security_context.go
```diff
@@ -616,3 +561,54 @@ func createCapabilityContainer(rc internalapi.RuntimeService, ic internalapi.Ima
...
+		log := parseLogLine(podConfig, logPath)
+		for _, msg := range log {
+			if strings.Contains(string(msg.log), fmt.Sprintf(":%s", hostNetworkPort)) {
+				if hostNetwork {
+					return nil
+				}
+
+				return fmt.Errorf("host port %s should not in container's port list", hostNetworkPort)
+			}
+
```
xlgao-zju: empty line.

feiskyer:
Updated. PTAL.

xlgao-zju:
LGTM

PR #98, opened 2017-05-19
Fix image test.

Random-Liu:
1. `ImageStatus` test should test both image id and image reference.
2. When using docker to pull image, no matter we use `busybox:latest`, `busybox`, `library/busybox`, `docker.io/library/busybox:latest`, only `busybox:latest` is shown in `RepoTags`. This is a docker specific behavior, we should not test that. In cri-containerd, we prefer always returning full path `docker.io/library/busybox:latest` to make it neutral and consistent.

Signed-off-by: Random-Liu <lantaol@google.com>

### DIFF
--- a/pkg/framework/util.go
+++ b/pkg/framework/util.go
@@ -192,8 +192,8 @@ func CreateContainer(rc internalapi.RuntimeService, ic internalapi.ImageManagerS
 		Logf("Use latest as default image tag.")
 	}
 
-	images := ListImageForImageName(ic, imageName)
-	if len(images) == 0 {
+	status := ImageStatus(ic, imageName)
+	if status == nil {
 		PullPublicImage(ic, imageName)
 	}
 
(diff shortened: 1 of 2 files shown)


feiskyer:
@Random-Liu CI failed:
image

repo is not included in docker's image ID. I think we should change `equal` to `contain` instead.

Random-Liu:
The current CRI image reference definition is not quite clear, and the current dockershim implementation is problematic.

Discussed this with @yujuhong today, and we'll fix it soon hopefully. After that this test will pass.

Random-Liu:
Updated the PR:
1) We should use `HasSuffix` to check whether image tag is in `RepoTags`.
2) We should not rely on list image filter for now, because: a) it's not used in kubelet; b) it will not work with docker 1.13 https://github.com/kubernetes/kubernetes/issues/34308#issuecomment-302833208

PR #101, opened 2017-05-29
cmd: crictl: add quiet option

runcom:
We're porting all of the CRI-O integration tests to cri-tools. This patch ensures a smooth transaction and add a useful feature to `ls` commands (think about batch removal).

This is also part of https://github.com/kubernetes-incubator/cri-o/issues/504


runcom:
@Random-Liu PTAL

feiskyer:
LGTM

PR #103, opened 2017-06-14
Use schema 2 image for HostPid test.

Random-Liu:
Use schema 2 image which doesn't affect test result. :)

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -34,7 +34,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


feiskyer:
@Random-Liu Is this because containerd only supports schema 2 images?

feiskyer:
LGTM

Random-Liu:
> @Random-Liu Is this because containerd only supports schema 2 images?

Yeah, they'll support schema 1 soon, but not yet finished.

PR #107, opened 2017-06-15
Set default config file to empty

feiskyer:
Set default config file to empty and don't log errors if it is not specified.


xlgao-zju:
@feiskyer Could you update the readme about this change?

feiskyer:
@xlgao-zju Updated.

PR #111, opened 2017-06-22
Add cri-tools docs

feiskyer:


docs/crictl.md
```diff
@@ -0,0 +1,122 @@
...
+### Run sandbox with config file
+
+```
+# cat sandbox-config.json
+{
+    "metadata": {
+        "name": "nginx-sandbox",
+        "namespace": "default",
+        "attempt": 1,
+        "uid": "hdishd83djaidwnduwk28bcsb"
```
xlgao-zju: Not sure about this line.

feiskyer: @xlgao-zju  What do you mean? this is the Pod UID.

xlgao-zju: @feiskyer I mean do we need UID when we create a pod?

feiskyer: It is a required field in kubelet, so it's better to set it in case some runtimes depends on the field.

docs/crictl.md
```diff
@@ -0,0 +1,122 @@
...
+- `image`: Manage images
+- `exec`: Exec(exec, syncexec) a command in a running container
+- `portforward`: Forword ports(localport:remoteport) from a sandbox
+- `help`, `h`: Shows a list of commands or help for one command
+
+crictl connects to `/var/run/dockershim.sock` by default. For other runtimes, the endpoint can be set in two ways:
+
+- By setting flags `--runtime-endpoint` and `--image-endpoint`
+- By setting environment variables `CRI_RUNTIME_ENDPOINT` and `CRI_IMAGE_ENDPOINT`
+
```
xlgao-zju: And we also support `By setting the endpoint in the config file`, put the usage of config file here?

docs/validation.md
```diff
@@ -0,0 +1,48 @@
...
+This will
+
+- Connect to the shim of CRI container runtime
+- Run the tests using `ginkgo`
+- Output the test results to STDOUT
+
+critest connects to `/var/run/dockershim.sock` by default. For other runtimes, the endpoint can be set in two ways:
+
+- By setting flags `--runtime-endpoint` and `--image-endpoint`
+- By setting environment variables `CRI_RUNTIME_ENDPOINT` and `CRI_IMAGE_ENDPOINT`
```
xlgao-zju: Also add the usage of config file here too?

feiskyer: tritest doesn't support config file

docs/crictl.md
feiskyer: ACK, added.

xlgao-zju:
@Random-Liu @yujuhong Would you like take a look at this PR?

xlgao-zju:
LGTM

PR #119, opened 2017-08-14
Split Travis CI into multiple stages

asifdxtreme:
Split the travis Build to different logical stages so that it's easy to understand what went wrong in CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,12 @@ before_install:
 install:
   - make install.tools
 
-
-script:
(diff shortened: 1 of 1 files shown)


feiskyer:
CI failed because critest not found, seems data not shared in different stages. 

@asifdxtreme Adds another make in the second build?

asifdxtreme:
@feiskyer make was not required in Static check phase so I have move it to test phase, now the build has passed

PR #127, opened 2017-08-30
Add supplemental groups test.

Random-Liu:
@feiskyer 


Random-Liu:
@feiskyer Thanks!

PR #129, opened 2017-09-04
automate load config file from default path

### LINKED ISSUE #128: crictl should automate load config file from default path.
crictl should automate load config file from default path  no matter wether it is exist in command line args.
Now even though I has writed config file in /etc/crictl.yaml , I must enter "crictl  --config /etc/crictl.yaml".

@Random-Liu

yanxuean:
fix #128 
Signed-off-by: yanxuean <yan.xuean@zte.com.cn>

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -134,8 +134,19 @@ func main() {
 	}
 
 	app.Before = func(context *cli.Context) error {
+		isUseConfig := false
 		configFile := context.GlobalString("config")
-		if configFile == "" {
+		if _, err := os.Stat(configFile); err == nil {
+			isUseConfig = true
+		} else {
+			if context.IsSet("config") || !os.IsNotExist(err) {
+				// note: the absence of default config file is normal case
+				// when user have not setted it in cli
+				logrus.Fatalf("Falied to load config file: %v", err)
+			}
+		}
+
+		if !isUseConfig {
 			RuntimeEndpoint = context.GlobalString("runtime-endpoint")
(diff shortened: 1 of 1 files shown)


yanxuean:
@feiskyer    @Random-Liu   PTAL

feiskyer:
@yanxuean Thanks for the PR. crictl will always fatal if `/etc/crictl.yaml` doesn't exist, while this is not expected. 

Could you update the PR and fallback to flags and env if `/etc/crictl.yaml` doesn't exist?

yanxuean:
yes，I will solve  it

yanxuean:
@feiskyer  @Random-Liu  PTAL, Thanks

cmd/crictl/main.go
```diff
@@ -133,8 +134,26 @@ func main() {
 	}
 
 	app.Before = func(context *cli.Context) error {
+		isUseConfig := false
 		configFile := context.GlobalString("config")
-		if configFile == "" {
+		if !context.IsSet("config") {
```
feiskyer: This could be simplified as

```go
if _, err := os.Stat(configFile); err == nil {
    isUseConfig = true
}
```

yanxuean: done

yanxuean: We can not omit the error handling. 
We always show result to user when user self indicate the config file in command line.  
But we should not show result to user when the default config file is absence. This is a normal case.

feiskyer:
@yanxuean  LGTM. thanks a lot

Random-Liu:
@yanxuean Thanks!

Issue #24, opened 2017-03-17
critest: experimental features


feiskyer:
critest for experimental features, including

- [x] apparmor #140 
- [x] sysctls #139 
- [x] seccomp #137

We should test sysctl, seccomp and apparmor. These tests may not be conformance test, but are required in CRI validation test suite.

From @Random-Liu: Probably just migrate test from node e2e https://github.com/kubernetes/kubernetes/blob/master/test/e2e/node/security_context.go.

note: sysctl/apparmor tests are also not included in node e2e tests.

feiskyer:
@mikebrow  Are you working on other tests listing here?

mikebrow:
Not yet. I'll open a WIP PR if I start any new tests and flag you.

feiskyer:
@mikebrow OK, I will work on sysctls tests first.

feiskyer:
testings added, closing this one.

PR #135, opened 2017-09-15
Add seccomp validation tests

feiskyer:
Part of #24: Add seccomp validation tests.

### DIFF
--- a/hack/install-docker.sh
+++ b/hack/install-docker.sh
@@ -18,10 +18,20 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# Install docker
+
+# Install docker.
 sudo sh -c 'echo "deb https://apt.dockerproject.org/repo ubuntu-$(lsb_release -cs) main" > /etc/apt/sources.list.d/docker.list'
 curl -fsSL https://apt.dockerproject.org/gpg | sudo apt-key add -
 sudo apt-key fingerprint 58118E89F3A912897C070ADBF76221572C52609D
 sudo apt-get update
-sudo apt-get -y install "docker-engine=1.13.1-0~ubuntu-$(lsb_release -cs)"
+sudo apt-get -y install "docker-engine=17.03.1~ce-0~ubuntu-$(lsb_release -cs)"
+
+# docker debs don't support seccomp, so we install a static version instead.
+curl -sSL -o docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-17.03.1-ce.tgz
+tar xzvf docker.tgz
+sudo cp -f docker/* /usr/bin
+rm -f docker.tgz
+
+# Restart docker daemon.
+sudo service docker restart
 
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -18,9 +18,11 @@ package validate
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


feiskyer:
CI failed because dockerd installed doesn't support seccomp:

seccomp profiles are not supported on this daemon, you cannot specify a custom seccomp profile.

hack/install-docker.sh
```diff
@@ -18,10 +18,20 @@ set -o errexit
...
+# Install docker.
 sudo sh -c 'echo "deb https://apt.dockerproject.org/repo ubuntu-$(lsb_release -cs) main" > /etc/apt/sources.list.d/docker.list'
 curl -fsSL https://apt.dockerproject.org/gpg | sudo apt-key add -
 sudo apt-key fingerprint 58118E89F3A912897C070ADBF76221572C52609D
 sudo apt-get update
-sudo apt-get -y install "docker-engine=1.13.1-0~ubuntu-$(lsb_release -cs)"
+sudo apt-get -y install "docker-engine=17.03.1~ce-0~ubuntu-$(lsb_release -cs)"
+
+# docker debs don't support seccomp, so we install a static version instead.
+curl -sSL -o docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-17.03.1-ce.tgz
```
Random-Liu: Cool! :) Ref https://github.com/moby/moby/issues/22870

hack/install-docker.sh
```diff
@@ -18,10 +18,20 @@ set -o errexit
...
 
-# Install docker
+
+# Install docker.
 sudo sh -c 'echo "deb https://apt.dockerproject.org/repo ubuntu-$(lsb_release -cs) main" > /etc/apt/sources.list.d/docker.list'
 curl -fsSL https://apt.dockerproject.org/gpg | sudo apt-key add -
 sudo apt-key fingerprint 58118E89F3A912897C070ADBF76221572C52609D
 sudo apt-get update
-sudo apt-get -y install "docker-engine=1.13.1-0~ubuntu-$(lsb_release -cs)"
+sudo apt-get -y install "docker-engine=17.03.1~ce-0~ubuntu-$(lsb_release -cs)"
```
Random-Liu: Do we still need this given that we are using static version now.

Random-Liu:
@feiskyer LGTM with one question.

hack/install-docker.sh
feiskyer: this is for manage docker service, e.g. we could restart docker with `service docker restart` command. Or else, we need to manage service by ourself, e.g. if docker is started in first stage, we need to kill dockerd and start again in second stage.

Random-Liu: Another question is that why do we use docker 17.03? There is no available 1.13 static binary?

feiskyer: yep, didn't found 1.13 static binary. 17.03 is the oldest one and also will be supported in kuberentes 1.8.

Random-Liu: @feiskyer I don't think `17.03` is supported in Kubernetes 1.8, but yeah I also checked, 17.03 is the oldest one.

I'm fine with testing it for now. We'll be setting up CRI validation in Kubernetes test infra soon, we could test against docker 1.13 then.

feiskyer: >  I don't think 17.03 is supported in Kubernetes 1.8

@Random-Liu See https://github.com/kubernetes/kubernetes/issues/42926.

Random-Liu: If so, it means that https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/system/types.go#L162 needs to be updated.

@yguo0905 told me that docker 1.13 is the newest version.

yguo0905: Sorry, I should have mentioned that they are the same minor release. Created https://github.com/kubernetes/kubernetes/pull/52666 to add 17.03 support to the spec.

PR #137, opened 2017-09-15
Additional Seccomp tests

mikebrow:
Adding support for seccomp testing. 

Will add pod testing after validating these container tests.

@Random-Liu

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/cri-tools/pkg/framework"
 	internalapi "k8s.io/kubernetes/pkg/kubelet/apis/cri"
 	runtimeapi "k8s.io/kubernetes/pkg/kubelet/apis/cri/v1alpha1/runtime"
@@ -37,6 +38,27 @@ import (
 
 const (
 	nginxContainerImage string = "nginx"
+	localhost           string = "localhost/"
+	// profile which denies sethostname syscall
+	seccompBlockHostNameProfile = `{
+     "defaultAction": "SCMP_ACT_ALLOW",
+     "syscalls": [
+         {
+             "name": "sethostname",
+             "action": "SCMP_ACT_ERRNO"
+         }
+     ]
+}`
+	// profile which denies chmod syscall
+	seccompBlockChmodProfile = `{
+     "defaultAction": "SCMP_ACT_ALLOW",
+     "syscalls": [
+         {
+             "name": "chmod",
+             "action": "SCMP_ACT_ERRNO"
+         }
+     ]
+}`
 )
 
 var _ = framework.KubeDescribe("Security Context", func() {
@@ -50,7 +72,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 		ic = f.CRIClient.CRIImageClient
 	})
 
-	Context("runtime should support NamespaceOption in the Security Context", func() {
+	Context("NamespaceOption", func() {
 		var podID string
 		var podConfig *runtimeapi.PodSandboxConfig
 		podSandboxName := "NamespaceOption-PodSandbox-" + framework.NewUUID()
@@ -62,7 +84,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			rc.RemovePodSandbox(podID)
 		})
 
-		It("runtime should support HostPID [security context]", func() {
+		It("runtime should support HostPID", func() {
 			By("create podSandbox for security context HostPID")
 			namespaceOption := &runtimeapi.NamespaceOption{
 				HostPid:     true,
@@ -116,7 +138,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 		})
 
-		It("runtime should support HostIpc is true [security context]", func() {
+		It("runtime should support HostIpc is true", func() {
 			By("create shared memory segment on the host")
 			out, err := exec.Command("ipcmk", "-M", "1M").Output()
 			framework.ExpectNoError(err, "failed to execute ipcmk -M 1M")
@@ -149,7 +171,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			Expect(string(out)).To(ContainSubstring(segmentID), "The shared memory segment should be included in the container")
 		})
 
-		It("runtime should support HostIpc is false [security context]", func() {
+		It("runtime should support HostIpc is false", func() {
 			By("create shared memory segment on the host")
 			out, err := exec.Command("ipcmk", "-M", "1M").Output()
 			framework.ExpectNoError(err, "failed to execute ipcmk -M 1M")
@@ -182,7 +204,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			Expect(string(out)).NotTo(ContainSubstring(segmentID), "The shared memory segment should be included in the container")
 		})
 
-		It("runtime should support HostNetwork is true [security context]", func() {
+		It("runtime should support HostNetwork is true", func() {
 			srv, err := net.Listen("tcp", ":0")
 			if err != nil {
 				framework.Failf("Failed to listen a tcp port: %v", err)
@@ -203,7 +225,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			podID = createAndCheckHostNetwork(rc, ic, podSandboxName, ports[len(ports)-1], true)
 		})
 
-		It("runtime should support HostNetwork is false [security context]", func() {
+		It("runtime should support HostNetwork is false", func() {
 			srv, err := net.Listen("tcp", ":0")
 			if err != nil {
 				framework.Failf("Failed to listen a tcp port: %v", err)
@@ -225,7 +247,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 		})
 	})
 
-	Context("runtime should support container with security context", func() {
+	Context("bucket", func() {
 		var podID, hostPath string
 		var podConfig *runtimeapi.PodSandboxConfig
 
@@ -236,7 +258,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			rc.RemovePodSandbox(podID)
 		})
 
-		It("runtime should support SupplementalGroups [security context]", func() {
+		It("runtime should support SupplementalGroups", func() {
 			By("create pod")
 			podID, podConfig = framework.CreatePodSandboxForContainer(rc)
 
@@ -269,7 +291,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			Expect(groups).To(ContainElement("5678"))
 		})
 
-		It("runtime should support RunAsUser [security context]", func() {
+		It("runtime should support RunAsUser", func() {
 			By("create pod")
 			podID, podConfig = framework.CreatePodSandboxForContainer(rc)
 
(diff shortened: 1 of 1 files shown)


Random-Liu:
@mikebrow According to https://github.com/kubernetes-incubator/cri-tools/pull/135, seccomp is not supported in the current version of Docker we are using.

mikebrow:
Wish I'd known someone was drafting seccomp tests :-)

mikebrow:
@Random-Liu @feiskyer all good, I don't see any overlap in our test cases. I'll normalize and combine them on Monday. For example, he's using chmod and grep which should work with docker/default and they do not require caps to be set before generating the default seccomp profile (which from my investigation is different with and without certain caps and on different platforms. I'm using hostname which requires caps to make docker/default allow sethostname, so I'm testing with and without caps on to cover both fail and pass cases.  I'm also testing the privileged case to make sure you can't block a command when privileged, and i've got a test for runtime/default as that may also be used instead of docker/default. And since I did the push I was adding a test for "foo" as a profile to make sure there is a test forcing the profile name to be of the format "localhost/foo"

feiskyer:
@mikebrow Sorry for didn't let you know this. But cool with more complete test cases.

feiskyer:
@mikebrow needs a rebase now

mikebrow:
Rebased. 

@feiskyer  I see there was a change to the CRI api documentation

https://github.com/kubernetes/kubernetes/commit/c242432a3b566ac82d3613689d8af92228cb6f58#diff-df996ad3ce05a000c2009d6e8cfba507L486

Talked to @Random-Liu about this. Will just remove the runtime/default tests for now.

mikebrow:
@Random-Liu @feiskyer rebased. Tests passing ready for review.

pkg/validate/security_context.go
```diff
@@ -402,6 +424,41 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			checkNetworkManagement(rc, containerID, false)
 		})
+	})
+
+	Context("runtime should support seccomp security context", func() {
```
feiskyer: nits: `[security context]` same with other places.

mikebrow: the top level bucket is "[k8s.io] security context" 
This context is for seccomp tests within [k8s.io] security context.

Not sure what the point is of having brackets to both the left and right most extent of the individual tests.

Let me refactor all of the titles making it easier to regexp and only having brackets on the left for ease of reading.

mikebrow:
Refactored the context container descriptions to fix regexp options and make it easier to read. Removed repetitive text from the suite descriptions again to make it easier to read and easier to identify regexp options.

pkg/validate/security_context.go
Random-Liu: @feiskyer On second thought, I also think it is fine to remove the tag given how cri validation test is organized today. :)

pkg/validate/security_context.go
```diff
@@ -225,7 +247,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 		})
 	})
 
-	Context("runtime should support container with security context", func() {
+	Context("bucket", func() {
```
Random-Liu: What do you mean by "bucket"?

pkg/validate/security_context.go
```diff
@@ -402,8 +424,43 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			checkNetworkManagement(rc, containerID, false)
 		})
+	})
+
+	Context("SeccompProfilePath", func() {
```
Random-Liu: Probably just `Seccomp`?

pkg/validate/security_context.go
```diff
@@ -37,6 +38,27 @@ import (
 
 const (
 	nginxContainerImage string = "nginx"
+	localhost           string = "localhost/"
+	// profile which denies sethostname syscall
+	seccompBlockHostNameProfile = `{
```
Random-Liu: Why not moving those 2 consts into the Seccomp context. No one else is using it.

pkg/validate/security_context.go
```diff
@@ -402,8 +424,43 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+
+	Context("SeccompProfilePath", func() {
+		var podID, profileDir, blockHostNameProfilePath, blockchmodProfilePath string
+		var podConfig *runtimeapi.PodSandboxConfig
+		var err error
+		sysCaps := []string{"ALL"}
+
+		BeforeEach(func() {
+			profileDir, err = createSeccompProfileDir()
+			if err != nil {
```
Random-Liu: Don't think we could just log and return, it will still run test then.
You should do assertion here `Expect(err).NotTo(HaveOccurred())`.

feiskyer:
@mikebrow thanks 👍

Random-Liu:
@feiskyer I'm still reviewing. There are issues. :p

feiskyer:
hmm, didn't notice issues. we could open another fix pr for the issues.

pkg/validate/security_context.go
```diff
@@ -450,8 +502,123 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+				"container-with-block-hostname-seccomp-profile-test-",
+				localhost+blockHostNameProfilePath, nil, privileged, expectContainerCreateToPass)
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+			checkSetHostname(rc, containerID, true)
+		})
+
+		It("runtime should block setting hostname with no seccomp profile specified", func() {
```
Random-Liu: no seccomp profile means "unconfined", why set hostname will be blocked? By whom?
Default capabilities? If so, I feel like this is more a capability test instead of seccomp test.

mikebrow:
There was an offline discussion about adding another context to separate docker/default tests. I'll push a pr for that in the morn...

pkg/validate/security_context.go
```diff
@@ -450,8 +502,123 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+			containerID := framework.CreateDefaultContainer(rc, ic, podID, podConfig,
+				"container-with-no-seccomp-profile-test-")
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+			checkSetHostname(rc, containerID, false)
+		})
+
+		It("runtime should support setting hostname with docker/default seccomp profile and sysCaps", func() {
```
Random-Liu: Move this to a separate `docker/default` context.

pkg/validate/security_context.go
```diff
@@ -450,8 +502,123 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+			containerID := createSeccompContainer(rc, ic, podID, podConfig,
+				"container-with-dockerdefault-seccomp-profile-test-", "docker/default", sysCaps, privileged, expectContainerCreateToPass)
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+			checkSetHostname(rc, containerID, true)
+		})
+
+		It("runtime should block sethostname with docker/default seccomp profile and no sysCaps", func() {
```
Random-Liu: ditto.

pkg/validate/security_context.go
```diff
@@ -450,8 +502,123 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+			containerID := createSeccompContainer(rc, ic, podID, podConfig,
+				"container-with-runtimedefault-seccomp-profile-test-", "unconfined", nil, privileged, expectContainerCreateToPass)
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+			checkSetHostname(rc, containerID, false)
+		})
+
+		It("runtime should block sethostname with nil seccomp profile and no sysCaps", func() {
```
Random-Liu: Same. This is a default capability test, not seccomp test.

pkg/validate/security_context.go
```diff
@@ -450,8 +502,123 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+			containerID := createSeccompContainer(rc, ic, podID, podConfig,
+				"container-with-dockerdefault-seccomp-profile-test-", "docker/default", nil, privileged, expectContainerCreateToPass)
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+			checkSetHostname(rc, containerID, false)
+		})
+
+		It("runtime should block sethostname with unconfined seccomp profile and no sysCaps", func() {
```
Random-Liu: Same. This is a default capability test, not seccomp test.

pkg/validate/security_context.go
Random-Liu: WE can add a test to create a default container, and check whether seccomp is enabled. That is useful.

pkg/validate/security_context.go
Random-Liu: @mikebrow This is the issue we need to fix.

Random-Liu:
@mikebrow Please address the comments.

pkg/validate/security_context.go
mikebrow: sure I'll move..

pkg/validate/security_context.go
mikebrow: a bucket of tests...  This is the second "bucket" of bucket-2, would be another possible name.  This context is already within the "Security Context" .. the other buckets at this level are NamespaceOption and SeccompProfilePath.  

This bucket comprises tests for: SupplementalGroups, RunAsUser, RunAsUserName, ReadOnlyRootfs, Privileged, and Capability. I couldn't come up with a better name for these than the "Security Context bucket"

pkg/validate/security_context.go
mikebrow: I like using the litteral CRI spec name here as it doesn't test all of Seccomp here. It is literally testing the CRI spec defined options for the SeccompProfilePath:		
```&runtimeapi.LinuxPodSandboxConfig{
			SecurityContext: &runtimeapi.LinuxSandboxSecurityContext{
				SeccompProfilePath: string,
			},
		},```

File: pkg/validate/security_context.go (hunk 5 above)
Comment (mikebrow) [reply pkg/validate/security_context.go]: agreed will update ..

File: pkg/validate/security_context.go (hunk 6 above)
Comment (mikebrow) [reply pkg/validate/security_context.go]: This is a duplicate test, I'll delete it.

File: pkg/validate/security_context.go (hunk 7 above)
Comment (mikebrow) [reply pkg/validate/security_context.go]: Yes, we both had that idea at the same time :-)

File: pkg/validate/security_context.go (hunk 10 above)
Comment (mikebrow) [reply pkg/validate/security_context.go]: it's both a default caps test and a seccomp test. See unconfined and default rules below:
```
	// Seccomp profile for the sandbox, candidate values are:
	// * docker/default: the default profile for the docker container runtime
	// * unconfined: unconfined profile, ie, no seccomp sandboxing
	// * localhost/<full-path-to-profile>: the profile installed on the node.
	//   <full-path-to-profile> is the full path of the profile.
	// Default: "", which is identical with unconfined.
	SeccompProfilePath string `protobuf:"bytes,7,opt,name=seccomp_profile_path,json=seccompProfilePath,proto3" json:"seccomp_profile_path,omitempty"`
```
If you have an idea for a better way to test the unconfined and nil SeccompProfilePath options.. I'm all ears :)

pkg/validate/security_context.go
mikebrow: see above response to your comments.. maybe we should do a quick call on this.

PR #140, opened 2017-09-19
Add validation tests for Apparmor

feiskyer:
Part of #24. 

Seems [apiextensions-apiserver](https://github.com/kubernetes/apiextensions-apiserver) hasn't synced recently, so `k8s.io/apiextensions-apiserver/pkg/features` will not be found in any packages and updating vendors for `k8s.io/apiextensions-apiserver` will report `WARNING: package k8s.io/apiextensions-apiserver is unused, consider removing it from vendor.conf`.

### DIFF
--- a/pkg/validate/apparmor.go
+++ b/pkg/validate/apparmor.go
@@ -1,0 +1,194 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validate
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/kubernetes-incubator/cri-tools/pkg/framework"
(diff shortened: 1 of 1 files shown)


feiskyer:
Update: actually only one function is used from `k8s.io/kubernetes/pkg/security/apparmor` package, replaced with a local function instead.

pkg/validate/apparmor.go
```diff
@@ -0,0 +1,193 @@
...
+profile cri-validate-apparmor-test-audit-write flags=(attach_disconnected) {
+  #include <abstractions/base>
+  file,
+  # Only audit file writes.
+  audit /** w,
+}
+`
+)
+
+var _ = framework.KubeDescribe("Security Context", func() {
```
Random-Liu: So we could have different describe with the same name? Didn't know that

pkg/validate/apparmor.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	glog.V(2).Infof("Loaded profiles: %v", out)
+	return nil
+}
+
+// isAppArmorEnabled returns true if apparmor is enabled for the host.
+// This function is forked from
+// https://github.com/opencontainers/runc/blob/1a81e9ab1f138c091fe5c86d0883f87716088527/libcontainer/apparmor/apparmor.go
+// to avoid the libapparmor dependency.
```
Random-Liu: Add TODO to reuse the function after we could properly sync api?

pkg/validate/apparmor.go
feiskyer: Hmm, should be Apparmor here.

pkg/validate/apparmor.go
feiskyer: ack

feiskyer:
@Random-Liu Addressed comments, PTAL

Random-Liu:
LGTM

Issue #131, opened 2017-09-06
Test for new CRI features


Random-Liu:
We need CRI validation test for new CRI features:

- [x] NoNewPrivilege (May be hard to test) #141
- [x] Mount propagation (easy to test) #142 

We should at least add test for 2).

miaoyq:
@Random-Liu OK, I will do my best to finish this. : )

Random-Liu:
@miaoyq I think only do 2) is good enough for now. :)

feiskyer:
@Random-Liu @miaoyq updated descriptions to include a list of todos.

feiskyer:
@miaoyq Are you working on Mount propagation tests?

miaoyq:
@feiskyer  Yes, have finished the coding, testing now.

feiskyer:
@miaoyq cool. I will work on NoNewPrivilege.

feiskyer:
all done.

PR #141, opened 2017-09-20
Add validation for NoNewPrivs

feiskyer:
Part of #131.

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -38,6 +38,7 @@ import (
 const (
 	nginxContainerImage string = "nginx"
 	localhost           string = "localhost/"
+	noNewPrivsImage     string = "gcr.io/google_containers/nonewprivs:1.2"
 )
 
 var _ = framework.KubeDescribe("Security Context", func() {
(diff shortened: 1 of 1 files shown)


Random-Liu:
LGTM

pkg/validate/security_context.go
```diff
@@ -619,6 +621,80 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+				Linux: &runtimeapi.LinuxContainerConfig{
+					SecurityContext: &runtimeapi.LinuxContainerSecurityContext{
+						RunAsUser: &runtimeapi.Int64Value{
+							Value: uid,
+						},
+					},
+				},
+				LogPath: fmt.Sprintf("%s.log", name),
+			}
+			if noNewPrivs != nil {
```
mikebrow: I don't understand the nil test. The spec doesn't use *bool.. so nil isn't an option. Will be false or true.

pkg/validate/security_context.go
```diff
@@ -619,6 +621,80 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+		matchOutput := func(name, output string) {
+			By("check container's output")
+			expectedLog := &logMessage{
+				log:    []byte(output + "\n"),
+				stream: stdoutType,
+			}
+			verifyLogContents(podConfig, fmt.Sprintf("%s.log", name), expectedLog)
+		}
+
+		It("should allow privilege escalation when not explicitly set and uid != 0", func() {
```
mikebrow: same test as below below since unset isn't an option.  If the CRI guys want unset let's see if it's not to late to change it to *bool.

feiskyer: @mikebrow yep, default is equal to false.

feiskyer: This is for testing the default behavior. Do you think we should remove this case?

pkg/validate/security_context.go
feiskyer: This is for testing the default behavior. Do you think we should remove this case?

pkg/validate/security_context.go
mikebrow: So yeah, since it's not a *bool, nil isn't an option. Go is one of those languages where a bool starts out life as a false, an int starts with the value 0, a float 0.0 etc.. I'd just change the parameter to a bool instead of *bool.. and remove the nil test.. 

Cheers!

feiskyer:
@mikebrow Addressed comments. PTAL.

pkg/validate/security_context.go
```diff
@@ -579,6 +579,70 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+			rc.RemovePodSandbox(podID)
+			By("clean up the log dir")
+			os.RemoveAll(logPath)
+		})
+
+		createContainerWithNoNewPrivs := func(name string, noNewPrivs bool, uid int64) string {
+			By(fmt.Sprintf("create container %s", name))
+			containerConfig := &runtimeapi.ContainerConfig{
+				Metadata: framework.BuildContainerMetadata(name, framework.DefaultAttempt),
+				Image:    &runtimeapi.ImageSpec{Image: noNewPrivsImage},
```
mikebrow: I think you either lost your noNewPrivsImage in a rebase or maybe while cut and pasting the constant to move it to this context?

But this looks good to me. Just fix the build break :-)

feiskyer: hmm, lost in the rebase. Fixed.

Random-Liu:
Seems good to go? @feiskyer Feel free to merge it.

PR #143, opened 2017-09-20
Responding to comments on #137

mikebrow:
Addresses comments from @Random-Liu on #137.

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -27,7 +27,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/cri-tools/pkg/framework"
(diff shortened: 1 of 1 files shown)


mikebrow:
@Random-Liu per discussion I also removed the two unconfined cases with no additional caps set that were relying on the default caps currently set by docker and containerd.  As discussed we can test out default caps over in a cri-containerd bucket and leave this un-tested for now since the default caps are not specified (currently) in CRI.

Random-Liu:
> per discussion I also removed the two unconfined cases with no additional caps set that were relying on the default caps currently set by docker and containerd. As discussed we can test out default caps over in a cri-containerd bucket and leave this un-tested for now since the default caps are not specified (currently) in CRI.

Thanks! :)

Random-Liu:
LGTM

feiskyer:
LGTM

PR #144, opened 2017-09-21
Minor improvement on seccomp test.

Random-Liu:
In previous test, we don't know whether it's the default capability list blocks `sethostname` or seccomp does.

With this PR, we'll explicitly set `SYS_ADMIN`, which should allow `sethostname` when seccomp is unconfined; and seccomp should be able to block `sethostname` even when `SYS_ADMIN` exists.


feiskyer:
LGTM

Random-Liu:
@mikebrow @feiskyer Thanks for reviewing!

PR #146, opened 2017-09-27
Bump cri-tools version to 0.2

feiskyer:
For #145 .

PR #147, opened 2017-09-28
Fix streaming validation tests

feiskyer:
Fix the streaming validation error in #146.


feiskyer:
self merging since this is only a vendor updates.

PR #148, opened 2017-09-28
Add validation for volume when host path doesn't exist

miaoyq:
Runtime should support starting container with volume when host path doesn't exist

/cc @feiskyer @Random-Liu 

Signed-off-by: Yanqiang Miao <miao.yanqiang@zte.com.cn>


feiskyer:
LGTM

Random-Liu:
LGTM

Issue #151, opened 2017-10-04
Should SecurityContext information be in the sandbox config


baude:
In trying to get criod (https://github.com/kubernetes-incubator/cri-o) to pass the cri-tests, I am seeing an issue where criod expects there to be SecurityContext information in the config.

@Random-Liu Is SecurityContext required to be in the config or is it optional? If the latter, i can fix our code.  If the former, I'm happy to provide a PR if you have some default options you want defined.

@runcom fyi

mrunalp:
@feiskyer ping

feiskyer:
@runcom @mrunalp Sorry for the late response (long holiday here). LinuxContainerSecurityContext is optional (it may be nil), but LinuxPodSandboxSecurityContext is always not nil, see [here](https://github.com/kubernetes/kubernetes/blob/6e2249b78406d66e039d5285a12782119f7cf04c/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L135-L141) and [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/security_context.go#L29).

baude:
@feiskyer thanks for the response! closing as such

PR #152, opened 2017-10-11
Set LinuxSandboxSecurityContext as required in config

baude:
Kube expects LinuxSandboxSecurityContext to be part of the container
config (i.e. not nil).

This was discussed in issue #151:

https://github.com/kubernetes-incubator/cri-tools/issues/151

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/pkg/benchmark/pod.go
+++ b/pkg/benchmark/pod.go
@@ -49,7 +49,9 @@ var _ = framework.KubeDescribe("PodSandbox", func() {
 
 			config := &runtimeapi.PodSandboxConfig{
 				Metadata: framework.BuildPodSandboxMetadata(podSandboxName, uid, namespace, framework.DefaultAttempt),
-				Linux:    &runtimeapi.LinuxPodSandboxConfig{},
+				Linux: &runtimeapi.LinuxPodSandboxConfig{
+					SecurityContext: &runtimeapi.LinuxSandboxSecurityContext{},
+				},
 			}
 
 			operation := b.Time("create PodSandbox", func() {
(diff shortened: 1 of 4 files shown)


runcom:
LGTM

Random-Liu:
Sorry, I don't quite get it. Even if the field is nil, container runtime should tolerant it, right? Or else it's a bit fragile to me.

runcom:
The thing is, what should the runtime do if it's nil? Should we error out?

yujuhong:
Let's try to define what fields are optional in CRI. If a required field is not present in the configuration, runtime should error out IMO. 

As for this particular case, it should be required for Linux containers, since we don't have specific configurations for Windows yet. But in the future, this may not be true.

feiskyer:
> Based on what @feiskyer said #151 (comment), LinuxContainerSecurityContext is not required, LinuxSandboxSecurityContext is required, is that also expected?

A container may be not privileged while the sandbox is privileged (because other container is privileged). That's why we always set LinuxSandboxSecurityContext.Privileged in current implementation. 

And actually I think both of them are optional, and default is nil (means false for bool types). We may change kuberuntime to set LinuxSandboxSecurityContext.Privileged only when it's true.

> Let's try to define what fields are optional in CRI.

+1 for this. We should add docs for each field in CRI.

yujuhong:
> And actually I think both of them are optional, and default is nil (means false for bool types). We may change kuberuntime to set LinuxSandboxSecurityContext.Privileged only when it's true. WDYT?

If "optional" means we can assume the defaults are valid, then what you said makes sense.

In that case, runtime should still use the getters to get the default values.

baude:
The only problem with this approach is the getters dont return empty structs when the gettee is a struct as of today.  I think that is the main point in this discussion.  Where should that come from.  From the config or if the daemons realize the config is missing optional pieces, can they inject an empty (default) struct in its place?

Random-Liu:
@baude You don't need an empty struct. The getters generated by grpc can handle `nil` receiver.

baude:
Closing.

PR #164, opened 2017-10-19
security_context: do not rely on /etc/hosts

runcom:
It depends on runtime and CNI plugin used when testing. This patch uses
/etc/passwd which should be always available instead.

@feiskyer @Random-Liu @mrunalp @sameo  PTAL


mrunalp:
:+1:

rhatdan:
You really should rely on /, which is the only thing guaranteed to be in the container.  :^)

runcom:
> You really should rely on /, which is the only thing guaranteed to be in the container. :^)

changed to `/` - thanks Dan

sameo:
lgtm

feiskyer:
@runcom LGTM. Thanks for the fix.

PR #165, opened 2017-10-19
Port map test - IPv4 only please

rajatchopra:
When checking from loopback addresses, only IPv4 addresses work because no route_localnet flag is available for IPv6. 
This is for cases that do not use a user land proxy for port mapping but use SNAT/DNAT rules e.g. [cni portmap](https://github.com/containernetworking/plugins/tree/master/plugins/meta/portmap)

/cc @mrunalp


mrunalp:
Tested that this works fine! :+1:

baude:
+1

feiskyer:
@rajatchopra LGTM. Thanks.

Issue #166, opened 2017-10-26
crictl should consume extra debug information from CRI


Random-Liu:
We added the verbose option in CRI to make it possible for container runtime to provide more information for debug. kubernetes/kubernetes#53965

`crictl` should properly consume these information in `inspect`, `inspecti`, `inspects` and `info`.

miaoyq:
I will work on this.

miaoyq:
@Random-Liu @feiskyer  Could we output information for "inspect", "inspecti" and "inspects" only in json or yaml format?

PR #178, opened 2017-11-04
Use `protobuf/jsonpb` instead of `golang json` in some cases

### LINKED ISSUE #154: We should use github.com/golang/protobuf/jsonpb instead of golang json package.
Currently if a container exit with exit code 0, when we do `crictl inspect`, there won't be exit code field, because that is omitted https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/cri/v1alpha1/runtime/api.pb.go#L2119.

We should use `github.com/golang/protobuf/jsonpb`, so that we could configure `EmitDefaults` to fix this.

miaoyq:
For #154 

/cc @Random-Liu 

Signed-off-by: Yanqiang Miao <miao.yanqiang@zte.com.cn>

### DIFF
--- a/cmd/crictl/util.go
+++ b/cmd/crictl/util.go
@@ -24,6 +24,8 @@ import (
 	"sort"
 
 	"github.com/ghodss/yaml"
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
 	"github.com/urfave/cli"
 	"google.golang.org/grpc"
 	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
@@ -178,8 +180,17 @@ func outputYAML(v interface{}) error {
 	return nil
 }
 
-func outputStatusInfo(status interface{}, info map[string]string, format string) error {
-	statusByte, err := json.Marshal(status)
+func protobufObjectToJSON(obj proto.Message) (string, error) {
+	jsonpbMarshaler := jsonpb.Marshaler{EmitDefaults: true, Indent: "  "}
+	marshaledJSON, err := jsonpbMarshaler.MarshalToString(obj)
+	if err != nil {
+		return "", err
+	}
+	return marshaledJSON, nil
+}
+
+func outputStatusInfo(status proto.Message, info map[string]string, format string) error {
+	statusByte, err := protobufObjectToJSON(status)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


vendor/github.com/golang/protobuf/jsonpb/jsonpb.go
```diff
@@ -0,0 +1,829 @@
+// Go support for Protocol Buffers - Google's data interchange format
```
feiskyer: This is a vendor update, does vendor.conf need update?

Random-Liu:
@miaoyq How do we handle extra information then?

vendor/github.com/golang/protobuf/jsonpb/jsonpb.go
miaoyq: Only update jsonpb here, I think we dont need update vendor.conf unless other model depend on the new version of protobuf.  :)

miaoyq:
@Random-Liu We could handle extra info via "outputStatusInfo"

miaoyq:
@Random-Liu "protobufObjectToJSON(status)" handles resp.Status, and "json.Marshal" handles resp.Info

feiskyer:
> "protobufObjectToJSON(status)" handles resp.Status, and "json.Marshal" handles resp.Info

LGTM. @Random-Liu do you have any other concerns?

miaoyq:
>>"protobufObjectToJSON(status)" handles resp.Status, and "json.Marshal" handles resp.Info
>
>LGTM. @Random-Liu do you have any other concerns?

@Random-Liu @feiskyer  Updated, also related to #166.

cmd/crictl/util.go
```diff
@@ -178,8 +180,18 @@ func outputYAML(v interface{}) error {
...
 
-func outputStatusInfo(status interface{}, info map[string]string, format string) error {
-	statusByte, err := json.Marshal(status)
+func protobufObjectToJSON(obj proto.Message) (string, error) {
+	jsonpbMarshaler := jsonpb.Marshaler{EmitDefaults: true, Indent: "  "}
+	marshaledJSON, err := jsonpbMarshaler.MarshalToString(obj)
+	if err != nil {
+		return "", err
+	}
+
```
Random-Liu: nit: unnecessary empty line.

Random-Liu:
@miaoyq For some other commands, we also support yaml/json output. We should also use jsonpb for those cases. I'm fine with doing that in another PR, but please make sure the issue is open.

cmd/crictl/util.go
miaoyq: Done.

miaoyq:
>@miaoyq For some other commands, we also support yaml/json output. We should also use jsonpb for those cases. I'm fine with doing that in another PR, but please make sure the issue is open.

Will do.

PR #191, opened 2017-11-09
stop sandbox return with enter.

sak0:
cmdline stop sandbox return with enter. 

Signed-off-by: CuiHaozhi <cuihaozhi@chinacloud.com.cn>


sak0:
looks like ci not pass cause "go version go1.8.5"

sak0:
https://github.com/kubernetes/kubernetes/pull/55301
kubernetes Bump minimum supported go version to 1.9.1

PR #194, opened 2017-11-15
cmd: crictl: add a config command

runcom:
This is a fairly simple patch that just adds a `config` command to get and set configurations similar to `git config`.
This would be very needed for higher level commands like `kubeadm` in order to gather stuff like the `runtime-endpoint` to query.

@feiskyer @Random-Liu PTAL

```
$ crictl config --get runtime-endpoint
/var/run/crio.sock

$ crictl config runtime-endpoint /var/run/dockershim.sock

$ crictl config --get runtime-endpoint
/var/run/dockershim.sock

$ cat /etc/crictl.yaml
runtime-endpoint: /var/run/dockershim.sock
```

### DIFF
--- a/cmd/crictl/config.go
+++ b/cmd/crictl/config.go
@@ -17,8 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"io/ioutil"
+	"os"
+	"strconv"
 
+	"github.com/Sirupsen/logrus"
+	"github.com/urfave/cli"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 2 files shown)


runcom:
Bump

Random-Liu:
@runcom Thanks! This is really useful!

cmd/crictl/config.go
```diff
@@ -45,3 +50,80 @@ func ReadConfig(filepath string) (*Config, error) {
...
+		cli.StringFlag{
+			Name:  "get",
+			Usage: "get value: name",
+		},
+	},
+	Action: func(context *cli.Context) error {
+		configFile := context.GlobalString("config")
+		if _, err := os.Stat(configFile); err != nil {
+			// can't continue w/o config file
+			logrus.Fatalf("Falied to load config file: %v", err)
```
Random-Liu: Can we create an empty config here instead?

cmd/crictl/config.go
```diff
@@ -45,3 +50,80 @@ func ReadConfig(filepath string) (*Config, error) {
...
+		}
+		// Get config from file.
+		config, err := ReadConfig(configFile)
+		if err != nil {
+			logrus.Fatalf("Falied to load config file: %v", err)
```
Random-Liu: Can we return error instead? And same for the other log.

Random-Liu:
LGTM other than the comments.

cmd/crictl/config.go
feiskyer: Agreed. We can create the default config file `/etc/crictl.yaml` if it doesn't exist.

feiskyer:
@runcom Thanks for adding this.

cmd/crictl/config.go
runcom: fixed

cmd/crictl/config.go
runcom: fixed

runcom:
Comments addressed guys

PR #195, opened 2017-11-18
Update kubernetes to use new cri log library.

Random-Liu:
Depends on https://github.com/kubernetes/kubernetes/pull/55922.


yujuhong:
/lgtm
Is there a bot?

Random-Liu:
@yujuhong Yeah, there is. :)
But there is no submit queue. I'll manually merge it after test passes. Thanks!

PR #204, opened 2017-11-29
Include all states for `latest` and `last`.

Random-Liu:
Keep docker CLI behavior.
/cc @abhi 


abhi:
LGTM

PR #206, opened 2017-11-29
adding verbose debug output for inspect & inspects

mikebrow:
See https://github.com/kubernetes/kubernetes/pull/53965

Changes inspect (container) and inspects (sandbox) to be verbose by default, and using json format by default per discussion below.

Verbose is a new option for the subject status requests afforded by the CRI api.

I've also added "quiet" options to the these two inspect(s) in case someone does not want verbose, and to otherwise exercise the non verbose status requests.  

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -221,6 +221,10 @@ var containerStatusCommand = cli.Command{
 			Name:  "output, o",
 			Usage: "Output format, One of: json|yaml|table",
 		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "Do not show verbose information",
+		},
 	},
 	Action: func(context *cli.Context) error {
 		containerID := context.Args().First()
@@ -232,7 +236,7 @@ var containerStatusCommand = cli.Command{
 			return err
 		}
 
-		err := ContainerStatus(runtimeClient, containerID, context.String("output"))
+		err := ContainerStatus(runtimeClient, containerID, context.String("output"), context.Bool("quiet"))
 		if err != nil {
 			return fmt.Errorf("Getting the status of the container failed: %v", err)
 		}
@@ -462,12 +466,17 @@ func RemoveContainer(client pb.RuntimeServiceClient, ID string) error {
 
 // ContainerStatus sends a ContainerStatusRequest to the server, and parses
 // the returned ContainerStatusResponse.
-func ContainerStatus(client pb.RuntimeServiceClient, ID, output string) error {
+func ContainerStatus(client pb.RuntimeServiceClient, ID, output string, quiet bool) error {
+	verbose := !(quiet)
+	if output == "" { // default to json output
+		output = "json"
+	}
 	if ID == "" {
 		return fmt.Errorf("ID cannot be empty")
 	}
 	request := &pb.ContainerStatusRequest{
 		ContainerId: ID,
+		Verbose:     verbose,
 	}
 	logrus.Debugf("ContainerStatusRequest: %v", request)
 	r, err := client.ContainerStatus(context.Background(), request)
@@ -476,11 +485,15 @@ func ContainerStatus(client pb.RuntimeServiceClient, ID, output string) error {
 		return err
 	}
 
-	if output == "json" || output == "yaml" {
+	switch output {
+	case "json", "yaml":
 		return outputStatusInfo(r.Status, r.Info, output)
+	case "table": // table output is after this switch block
+	default:
+		return fmt.Errorf("output option cannot be %s", output)
 	}
 
-	// output in table format by default.
+	// output in table format
 	fmt.Printf("ID: %s\n", r.Status.Id)
 	if r.Status.Metadata != nil {
 		if r.Status.Metadata.Name != "" {
@@ -503,6 +516,9 @@ func ContainerStatus(client pb.RuntimeServiceClient, ID, output string) error {
 			fmt.Printf("Finished: %v\n", units.HumanDuration(time.Now().UTC().Sub(ftm))+" ago")
 		}
 		fmt.Printf("Exit Code: %v\n", r.Status.ExitCode)
+	}
+	if verbose {
+		fmt.Printf("Info: %v\n", r.GetInfo())
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


Random-Liu:
Look at https://github.com/kubernetes-incubator/cri-tools/blob/master/cmd/crictl/info.go
We should generate similar output.

mikebrow:
> Look at https://github.com/kubernetes-incubator/cri-tools/blob/master/cmd/crictl/info.go
We should generate similar output.

It does..

cmd/crictl/container.go
```diff
@@ -221,6 +221,10 @@ var containerStatusCommand = cli.Command{
 			Name:  "output, o",
 			Usage: "Output format, One of: json|yaml|table",
 		},
+		cli.BoolFlag{
```
Random-Liu: I think for inspect, we should always set `verbose=true` in crictl.
`verbose=false` is for kubelet using CRI.

cmd/crictl/container.go
```diff
@@ -221,6 +221,10 @@ var containerStatusCommand = cli.Command{
 			Name:  "output, o",
 			Usage: "Output format, One of: json|yaml|table",
```
Random-Liu: @feiskyer Do you still think we should have the `table` format?
I think for inspect, we may just need `json|yaml`. Just like `crictl info`. See an example [here](https://github.com/kubernetes-incubator/cri-containerd/pull/445#issuecomment-348025737)
The problem is that timestamp in CRI is unix timestamp, we may want to do a conversion in crictl.

runcom: I'm not a big fan of table, let's drop it in favor of machine readable output

cmd/crictl/container.go
mikebrow: @Random-Liu so you are saying you want a --quiet option for less verbose? or take out the option all-together? There were already other verbose flags so I figured that's what we would want with this verbose option.

Random-Liu: I feel like by default we should enable verbose, or else it's really troublesome to do -v every time. :)

As for whether we need a `quiet` option, I'm not sure.

cmd/crictl/container.go
mikebrow: I'm not a big fan of table formats either, but as @Random-Liu says the unix timestamp is not human readable output:

```
crictl --runtime-endpoint /var/run/cri-containerd.sock inspect -v -o json 44025055b48da66cb67bfc221f800fe50f1256d3e22f1cf714294344b29d582e 
{
  "status": {
    "id": "44025055b48da66cb67bfc221f800fe50f1256d3e22f1cf714294344b29d582e",
    "metadata": {
      "name": "busybox",
      "attempt": 0
    },
    "state": "CONTAINER_CREATED",
    "createdAt": "1511994785429097447",
    "startedAt": "0",
    "finishedAt": "0",
    "exitCode": 0,
    "image": {
      "image": "docker.io/library/busybox:latest"
    },
    "imageRef": "docker.io/library/busybox@sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0",
    "reason": "",
    "message": "",
    "logPath": ""
  },
```
vs:
```
crictl --runtime-endpoint /var/run/cri-containerd.sock inspect -v 44025055b48da66cb67bfc221f800fe50f1256d3e22f1cf714294344b29d582e 
ID: 44025055b48da66cb67bfc221f800fe50f1256d3e22f1cf714294344b29d582e
Name: busybox
State: CONTAINER_CREATED
Created: 28 minutes ago
```

cmd/crictl/container.go
mikebrow: @Random-Liu I hear ya.. think it over no rush, easy enough to switch it to verbose by default, or quiet mode option. But I think when you see the verbose mode from the CRI-O team you'll be begging to turn down the volume :-) @runcom

cmd/crictl/container.go
runcom: We can do some conversion there, not a big deal

cmd/crictl/container.go
runcom: ? Not following, what about CRI-O? We aren't even using that verbose stuff and we have no plan to output anything.
In any case, I'm in favor of going verbose by default like Lantao

cmd/crictl/container.go
Random-Liu: Yeah, we should do some conversion. :)

cmd/crictl/container.go
feiskyer: +1 of enabling verbose.

cmd/crictl/container.go
feiskyer: Agreed, table is not useful for inspect.

cmd/crictl/container.go
mikebrow: I merely meant that CRI-O keeps a lot of data about the container to possibly use in a verbose output of status about the container.

runcom: Yeah but that's exactly the verbose intention, despite what runtimes want to output

mikebrow: I vote inspect to be verbose by default for crictl (and provide a silent flag to turn verbose off).
I vote to make the json format the default (and I don't care if you want to kill the table format or keep it as an option).

mikebrow:
Ok I've changed inspect to use json format by default, made the default verbose, and introduced a quiet flag just because.

cmd/crictl/container.go
mikebrow: default changed to json..

cmd/crictl/container.go
mikebrow: done..

cmd/crictl/container.go
mikebrow: so.. @Random-Liu @runcom @feiskyer wrt. "some conversion"  
I suggest converting the timestamps to RFC3339Nano 
"2006-01-02T15:04:05.999999999Z07:00"
We can do it here but I don't like changing the output values provided from the runtime when representing them in json/yaml form. Suggest we specify in CRI that time values in status response info map should be in said RFC3339Nano format. Thoughts?

mikebrow:
I've added a new commit for the inspects (sandbox inspect) changes.

Random-Liu:
/lgtm

I think we don't need the table output, but let's keep it for now in case that someone may like it.

PR #208, opened 2017-11-30
Cleanup the label filter and fix a container list issue.

Random-Liu:
This PR:
1) Add `parseLabelStringSlice` util function.
2) Fix an issue that filter is performed after json/yaml output in `crictl ps`. /cc @abhi 


abhi:
LGTM

PR #217, opened 2017-12-13
Fix container/sandbox list.

Random-Liu:
For https://github.com/kubernetes-incubator/cri-tools/issues/179.

This PR:
1) Added `-l` and `-n` in `crictl sandboxes`.
2) Fix container list order, latest created container should be on top.
3) Change sandbox to sort by created timestamp, too. @yanxuean Are you ok with that? If user want to know sandboxes in the same namespace, they could always use `--namespace` to filter.

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -38,7 +38,7 @@ type containerByCreated []*pb.Container
 func (a containerByCreated) Len() int      { return len(a) }
 func (a containerByCreated) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
 func (a containerByCreated) Less(i, j int) bool {
-	return a[i].CreatedAt < a[j].CreatedAt
+	return a[i].CreatedAt > a[j].CreatedAt
 }
 
 type createOptions struct {
@@ -591,7 +591,6 @@ func ListContainers(client pb.RuntimeServiceClient, opts listOptions) error {
 		return err
 	}
 	r.Containers = getContainersList(r.GetContainers(), opts)
-	sort.Sort(containerByCreated(r.Containers))
 
 	switch opts.output {
 	case "json":
@@ -655,6 +654,7 @@ func ListContainers(client pb.RuntimeServiceClient, opts listOptions) error {
 }
 
 func getContainersList(containersList []*pb.Container, opts listOptions) []*pb.Container {
+	sort.Sort(containerByCreated(containersList))
 	n := len(containersList)
 	if opts.latest {
 		n = 1
@@ -669,5 +669,5 @@ func getContainersList(containersList []*pb.Container, opts listOptions) []*pb.C
 		return b
 	}(n, len(containersList))
 
-	return containersList[len(containersList)-n:]
+	return containersList[:n]
 }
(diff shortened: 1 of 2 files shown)


cmd/crictl/sandbox.go
```diff
@@ -192,6 +186,14 @@ var listPodSandboxCommand = cli.Command{
...
 			Usage: "Output format, One of: json|yaml|table",
 		},
+		cli.BoolFlag{
+			Name:  "latest, l",
+			Usage: "Show recently created sandboxes",
+		},
+		cli.IntFlag{
+			Name:  "last, n",
+			Usage: "Show last n recently created sandboxes",
+		},
```
feiskyer: Seems two options are same. Merge last to latest and defaulting latest to one?

Random-Liu: Initially just want to keep same UX with docker, but of course we could make change.

Random-Liu: @feiskyer I tried to merge them into one flag. But I find that it's hard to set default for `-l`, because once you define it as an int flag, you have to specify a value when using it.

Given that I think it's fine to keep 2 flags as what docker does.

cmd/crictl/container.go
```diff
@@ -590,8 +590,8 @@ func ListContainers(client pb.RuntimeServiceClient, opts listOptions) error {
 	if err != nil {
 		return err
 	}
+	sort.Sort(containerByCreated(r.GetContainers()))
```
yanxuean: I think getContainersList should include sort.Sort inside.

yanxuean:
>Change sandbox to sort by created timestamp, too. @yanxuean Are you ok with that? If user want to know sandboxes in the same namespace, they could always use --namespace to filter.

I am OK.

cmd/crictl/container.go
yanxuean: It is more completer in logic.

Random-Liu: Agree. Will do.

cmd/crictl/sandbox.go
```diff
@@ -172,12 +166,12 @@ var listPodSandboxCommand = cli.Command{
 			Usage: "filter by pod sandbox namespace",
```
yanxuean: Maybe we can add shortname for namespace?

cmd/crictl/sandbox.go
```diff
@@ -396,7 +400,8 @@ func ListPodSandboxes(client pb.RuntimeServiceClient, opts listOptions) error {
 	if err != nil {
 		return err
 	}
-	sort.Sort(sandboxBySort(r.Items))
+	sort.Sort(sandboxByCreated(r.GetItems()))
+	r.Items = getSandboxesList(r.GetItems(), opts)
```
yanxuean: ditto

cmd/crictl/sandbox.go
feiskyer: OK, didn't notice this problem. let's keep both

cmd/crictl/sandbox.go
Random-Liu: `last` is using the shortname `n`.

cmd/crictl/sandbox.go
Random-Liu: Done.

cmd/crictl/container.go
Random-Liu: Done.

PR #218, opened 2017-12-14
 Bump version to 1.0.0-alpha.0

feiskyer:


mrunalp:
I think we should match the version of this with kube so it is easy to tell what version works with which version of kube.

feiskyer:
@mrunalp Yep, that's another choice of versioning. But I'm thinking we should do that after CRI and cri-tools are GA. And branches will be required for each releases so as to match k8s versions.  

@Random-Liu WDYT?

PR #222, opened 2018-01-11
Rename `sandbox` to `podsandbox`.

Random-Liu:
This is discussed in sig-node. I sent the PR here to carry on the discussion.

The rule I used in this PR:
1) In command `s` -> `p`, and `sandbox` -> `pod`;
2) In usage `SANDBOX` -> `PODSANDBOX`;
3) `sandbox-config.yaml` -> `podsandbox-config.yaml`;
4) In flag `--sandbox` -> `--podsandbox`
5) In command description `sandbox` -> `pod sandbox`.

@yujuhong @feiskyer @runcom @mrunalp @mikebrow @abhi @yanxuean @miaoyq 


cmd/crictl/sandbox.go
```diff
@@ -147,8 +147,8 @@ var podSandboxStatusCommand = cli.Command{
 }
 
 var listPodSandboxCommand = cli.Command{
-	Name:  "sandboxes",
-	Usage: "List sandboxes",
+	Name:  "pods",
```
miaoyq: `pods` -> `podsandboxes`  ?

Random-Liu: This command will be frequently used, I feel like `podsandboxes` is too long, which will be very annoying in the future. :P

feiskyer: `pods` looks more simple

miaoyq: Make sense. :-)

mikebrow:
@Random-Liu I was expecting full replacement of sandbox for pod, not sandbox to pod sandbox in some cases? Might be wrong but I thought the point was only internal developers of kublet know what sandbox refers to.

Random-Liu:
@mikebrow I think at CRI level pod = `pod sandbox + containers`, so we may not want to call it a pod.

However, `sandbox` is new to user, so we call it `pod sandbox` to make it a bit more straight forward.

Issue #223, opened 2018-01-17
Add validation support for windows containers


feiskyer:
We should add validation support for windows containers, includes

- Fix building for windows
- Validation adjusting for windows containers

cc @JiangtianLi

JiangtianLi:
Thanks, @feiskyer. Let me know if I can help.

yujuhong:
> We should add validation support for windows containers, includes
> 
> * Fix building for windows
> * Validation adjusting for windows containers

@feiskyer has anything been done in these areas?

feiskyer:
The travis CI doesn't support Windows Server 2019, so we still need to find other infra for running those tests.

PR #225, opened 2018-01-17
 Fix building on Windows

feiskyer:
This is the first step of #223, it fixes building on Windows and add support of tcp endpoints.


Makefile
```diff
@@ -55,6 +55,12 @@ clean:
 	find . -name \*~ -delete
 	find . -name \#\* -delete
 
+cross: check-gopath
+		GOOS=windows $(GO) build -o $(CURDIR)/_output/critest.exe \
```
Random-Liu: too many indents?

feiskyer: Fixed.

PR #226, opened 2018-01-17
cmd: crictl: image: send FQN image names and fix listing for images with domain:port

runcom:
This patch does two things:

- send normalized image names over the CRI like Kube does today, otherwise, we'll end up sending strings like `busybox` which the runtime should not interpret (we send `docker.io/library/busybox:latest` instead, again, like Kube does)
- fix images list when the image domain contains `domain:port` (right now it's broken and everything after the domain, including the port and path is printed under TAG)

@mrunalp @feiskyer @Random-Liu  PTAL


mrunalp:
LGTM

PR #227, opened 2018-01-25
Set StdinOnce to true for attach test

mrunalp:


mrunalp:
We don't need to keep the stdin open for the test that we have.

runcom:
LGTM

Issue #228, opened 2018-01-26
CRI log parser improvements


jreyeshdez:
The function [`parseCRILog`](https://github.com/kubernetes-incubator/cri-tools/blob/bb2c5293bfa6dd5207e1ce3f2d5a32b09f1cc549/pkg/validate/container.go#L431) might have some room to introduce some changes to make it more performance as calling `strings.Fields` only once at the beginning of the method should be more efficient. I think it could even be more efficient if we use `strings.SplitN` as I think we do not need to split the whole message.

Likewise, it would be good to check the resulting array to make sure it has the correct or the expected size and avoid errors when accessing different fields, for example `2016-10-06T00:17:10.113242941Z stdout` as an input would cause panic.

If we use `strings.SplitN` then we wouldn't need `strings.Join`.

feiskyer:
@Random-Liu What do you think about this?

Random-Liu:
> The function parseCRILog might have some room to introduce some changes to make it more performance as calling strings.Fields only once at the beginning of the method should be more efficient. I think it could even be more efficient if we use strings.SplitN as I think we do not need to split the whole message.

That is util function used in the validation test, and performance shouldn't be a problem there. But feel free to send a PR to improve it. :) Thanks!

Random-Liu:
I think we can close this one.

PR #232, opened 2018-01-31
 CRI log parser improvements

jreyeshdez:
→ Possible fix for #228 :
- Modified [`parseCRILog`](https://github.com/kubernetes-incubator/cri-tools/blob/bb2c5293bfa6dd5207e1ce3f2d5a32b09f1cc549/pkg/validate/container.go#L431) to use `strings.SplitN` instead of multiple calls to `strings.Fields`.
- Check the resulting array to make sure it has the correct or the expected size and avoid errors when accessing different fields.

### DIFF
--- a/pkg/validate/container.go
+++ b/pkg/validate/container.go
@@ -19,6 +19,7 @@ package validate
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


feiskyer:
@Random-Liu Would you like to take another look?

Random-Liu:
@feiskyer No. I trust you. :D

Issue #233, opened 2018-02-06
Split Install / Uninstall into separate commands


mikebrow:
Ran into a desire, over on cri-containerd, to only install crictl in some circumstances vs both crictl and critest. Would rather stay away from roll your own custom builds of the tools. Build is already split up.

https://github.com/containerd/cri-containerd/pull/595

feiskyer:
Fixed by #234.

PR #234, opened 2018-02-06
splits install and uninstall actions for those who need only one of..

mikebrow:
This PR addresses issue #233 to split up install and uninstall into separate installs and uninstalls for crictl and critest. Build is already split up..

@Random-Li


Random-Liu:
Nice! Thanks! @mikebrow

PR #238, opened 2018-02-08
CRI v1alpha2

Random-Liu:
For https://github.com/kubernetes-incubator/cri-tools/issues/236.

Update CRI to v1alpha2.


Random-Liu:
Will merge after test passes.

PR #252, opened 2018-02-17
Improve image test and make test run in parallel.

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-tools/issues/241.

This PR:
1) Added `images/` directory which contains test images used in CRI validation test. @runcom We can also move your test image to `images/`. I'll push it to `gcr.io/cri-tools` for you.
2) Improve image test to make it be able to run in parallel with other tests.

By running test in parallel, it only takes <20s now.
```
•
Ran 56 of 57 Specs in 18.565 seconds
SUCCESS! -- 56 Passed | 0 Failed | 0 Pending | 1 Skipped 

```

### DIFF
--- a/hack/run-critest.sh
+++ b/hack/run-critest.sh
@@ -31,7 +31,7 @@ sleep 10
 
 # Run e2e test cases
 # Skip reopen container log test because docker doesn't support it.
-critest --skip="runtime should support reopening container log" v
+critest --skip="runtime should support reopening container log" --ginkgo-flags=--nodes=8 v
 
 # Run benchmark test cases
 critest b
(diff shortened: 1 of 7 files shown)


runcom:
LGTM but Travis doesn't like the changes

runcom:
@Random-Liu Travis' failing

runcom:
ping @Random-Liu @feiskyer

Random-Liu:
@runcom @feiskyer Fixed the test.

The previous test image I was using is a scratch image with an ENV set.
However, docker returns 0 size for that image, and it fails the size check https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/remote/remote_image.go#L88.

Random-Liu:
I think @feiskyer is on vacation. Apply LGTM based on https://github.com/kubernetes-incubator/cri-tools/pull/252#issuecomment-366434388

PR #267, opened 2018-03-15
critest: Add test for pid namespace and exec

umohnani8:
Add test for pid namespace = container
Add test for pid namespace = pod
Add test for exec where tty and stdin set to true

Addresses issue https://github.com/kubernetes-incubator/cri-tools/issues/237

Signed-off-by: umohnani8 <umohnani@redhat.com>

### DIFF
--- a/pkg/validate/security_context.go
+++ b/pkg/validate/security_context.go
@@ -184,6 +184,56 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			Expect(string(out)).NotTo(ContainSubstring(segmentID), "The shared memory segment should be included in the container")
 		})
 
+		It("runtime should support PodPID", func() {
+			By("create podSandbox for sharing process namespace")
+			namespaceOption := &runtimeapi.NamespaceOption{
+				Pid: runtimeapi.NamespaceMode_POD,
+			}
+			framework.Logf("Pid namespace is %q", namespaceOption.Pid)
+			podID, podConfig = createNamespacePodSandbox(rc, namespaceOption, podSandboxName, "")
+
+			By("create nginx container")
+			prefix := "nginx-container-"
+			containerName := prefix + framework.NewUUID()
+			containerID, _, _ := createNamespaceContainer(rc, ic, podID, podConfig, containerName, nginxContainerImage, namespaceOption, nil, "")
+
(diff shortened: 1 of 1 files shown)


umohnani8:
@mrunalp PTAL

pkg/validate/security_context.go
```diff
@@ -184,6 +184,56 @@ var _ = framework.KubeDescribe("Security Context", func() {
...
+			By("start container")
+			startContainer(rc, containerID)
+			Eventually(func() runtimeapi.ContainerState {
+				return getContainerStatus(rc, containerID).State
+			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
+
+			By("get nginx container pid")
+			command := []string{"cat", "/proc/1/cmdline"}
+			output := execSyncContainer(rc, containerID, command)
+			Expect(string(output)).To(ContainSubstring("pause"))
```
mrunalp: pause depends upon what container is being used by the runtime. I guess doing a not of "master process" will be more portable across runtimes and different pod infra containers.

pkg/validate/streaming.go
```diff
@@ -74,10 +74,38 @@ var _ = framework.KubeDescribe("Streaming", func() {
...
+				Stderr:      true,
+			}
+			req := createExec(rc, execReq)
 
 			By("check the output of exec")
-			checkExec(rc, req)
+			checkExec(rc, req, "hello\n", false)
+		})
+
+		It("runtime should support exec with -it flags", func() {
```
mrunalp: You should make a different PR for exec tests. Also we should add a test for exec without -it in that PR.

pkg/validate/security_context.go
umohnani8: okay, fixed.

pkg/validate/streaming.go
umohnani8: Removed the exec test, will open another PR with those tests.

Random-Liu:
@feiskyer What is the difference between these 2 test cases? Shouldn't we test one container can/can't not access pid in another container?

Random-Liu:
I see. One is contain, another is not contain. :)

PR #269, opened 2018-03-15
critest: Add test for exec

umohnani8:
Add test for exec where tty and stdin set to true

Signed-off-by: umohnani8 <umohnani@redhat.com>

### DIFF
--- a/pkg/validate/streaming.go
+++ b/pkg/validate/streaming.go
@@ -158,7 +158,21 @@ func createExec(c internalapi.RuntimeService, execReq *runtimeapi.ExecRequest) s
 func checkExec(c internalapi.RuntimeService, execServerURL, stdout string, isTty bool) {
 	localOut := &bytes.Buffer{}
 	localErr := &bytes.Buffer{}
-	localIn := &bytes.Buffer{}
+	localInRead, localInWrite := io.Pipe()
+
+	// Wait until output read and then shutdown localIn pipe.
+	go func() {
+		ticker := time.NewTicker(5 * time.Second)
+		for {
+			switch {
+			case len(localOut.String()) >= len(stdout):
+				fallthrough
+			case <-ticker.C != time.Time{}:
+				localInWrite.Close()
+				break
+			}
+		}
+	}()
 
 	// Only http is supported now.
 	// TODO: support streaming APIs via tls.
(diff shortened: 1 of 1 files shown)


umohnani8:
@mrunalp PTAL

umohnani8:
The exec test with stdin=true and tty=true is passing on cri-o, but fails on docker. Could be a bug in docker?

feiskyer:
For stdin=true and tty=true, docker exec will stop as soon as stdin is closed (and so with containerd @Random-Liu). Since the input is an empty buffer, it is closed immediately, so the output streams are closed immediately, too.

@umohnani8 Could you keep the stdin open until getting the output in `checkExec()`?

umohnani8:
@feiskyer how do I do that in the test? I thought the StreamOptions was accounting for that.

feiskyer:
@umohnani8  Let me add a patch to your PR

pkg/validate/streaming.go
```diff
@@ -119,39 +147,52 @@ var _ = framework.KubeDescribe("Streaming", func() {
...
+	By("exec given command in container: " + execReq.ContainerId)
+	resp, err := c.Exec(execReq)
+	framework.ExpectNoError(err, "failed to exec in container %q", execReq.ContainerId)
 	framework.Logf("Get exec url: " + resp.Url)
 	return resp.Url
 }
 
-func checkExec(c internalapi.RuntimeService, execServerURL string) {
+func checkExec(c internalapi.RuntimeService, execServerURL, stdout string, isTty bool) {
 	localOut := &bytes.Buffer{}
```
Random-Liu: Use safe buffer. https://github.com/umohnani8/cri-tools/blob/094e42ff5cde024a4e6bae544c1993e0f88d83a6/pkg/validate/streaming.go#L236

`bytes.Buffer` is not thread safe.

Actually use similar code with `checkAttach`. https://github.com/umohnani8/cri-tools/blob/094e42ff5cde024a4e6bae544c1993e0f88d83a6/pkg/validate/streaming.go#L257

feiskyer: Good catch 👍

umohnani8:
@feiskyer I see now, thanks for the patch :)

Random-Liu:
Actually we should test different behaviors when `stdin` and `tty` are true/false. Currently, the 2 test cases are exactly the same.

/lgtm for now.

PR #272, opened 2018-03-19
Revert "Allow combining of short boolean options (e.g. crictl ps -aq)."

runcom:
This reverts commit 804971138bc9f9f7f51311ab38837e01db7c6aa9.

This has broken "crictl exec" arguments in master (shown here in CRI-O as well https://github.com/kubernetes-incubator/cri-o/pull/1457):

Mar 19 17:28:52 runcom.internal crio[25912]: time="2018-03-19
17:28:52.150599893+01:00" level=debug msg="ExecSyncRequest
&ExecSyncRequest{ContainerId:7520b,Cmd:[-c echo hello world
sh],Timeout:0,}"

You can see above the args are totally wrong. Not sure what caused this
but given the change wasn't _needed_ let's revert and make exec work
again. Tested successfully with CRI-O.

@feiskyer @mrunalp @umohnani8 PTAL


mrunalp:
LGTM

umohnani8:
LGTM

mrunalp:
@Random-Liu PTAL

adelton:
So what was the `exec` command exactly? As part of https://github.com/kubernetes-incubator/cri-tools/pull/264 I've rebased `vendor/github.com/urfave` so it's possible that it has changed the behaviour and the breakage is not due to the `UseShortOptionHandling: true` change per se.

But to investigate I'd need to know how to reproduce.

runcom:
It's not related to UseShortOptionHandling indeed. Something else. You can reproduce by running cri-o and issuing the following

crictl exec --sync CNTID sh -c "echo hello world" 

You'll get the same log as above, which is clearly messing up with arguments order.

PR #274, opened 2018-03-20
until there is an auth plan tls testing should support insecure

mikebrow:
Over on https://github.com/containerd/cri/pull/681 we added support to containerd/cri and expected that critools would support tls without auth.  However, the following error occurs when attempt to run validation:

Mar 19 20:49:18.310: INFO: Unexpected error occurred: error sending request: Post https://10.20.0.36:10010/attach/Vy2abuwM: x509: certificate signed by unknown authority

This commit relaxes the certificate requirement on the server, before an encrypted stream can be started from crictl and critest validate.  

We should probably discuss, and/or open an issue to discuss, security requirements for streaming for the CRI API.

@Random-Liu 

### DIFF
--- a/cmd/crictl/attach.go
+++ b/cmd/crictl/attach.go
@@ -99,7 +99,7 @@ func Attach(client pb.RuntimeServiceClient, opts attachOptions) error {
 		return err
 	}
 	logrus.Debugf("Attach URL: %v", URL)
-	attach, err := remoteclient.NewSPDYExecutor(&restclient.Config{}, "POST", URL)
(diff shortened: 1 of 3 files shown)


Random-Liu:
Discussed with @tallclair, today the behavior of streaming authentication is not defined in CRI yet. And at least apiserver on GCE doesn't actually validate the certificate. I think it makes sense to skip the certificate validation in CRI validation test for now, until this is well-defined in CRI.

/lgtm

feiskyer:
LGTM

PR #282, opened 2018-03-31
Add run as group test

Random-Liu:
Fixes https://github.com/kubernetes-incubator/cri-tools/issues/280.

This PR:
1) Update Kubernetes to v1.10.0 version.
2) Add CRI test for `RunAsGroup`.
3) Cleanup the code a little bit:
  a) Replace several `[]byte` with `string`, because we are using `string` in most of the time.
  b) Unify the `AfterEach` in security_context.go.
  c) Simplify `verifyLogContents`.
    1) Accept `output` and `stream` directly.
    2) Find output in container log instead of matching each line.


Random-Liu:
The test failure is caused by a bug in dockershim https://github.com/kubernetes/kubernetes/pull/61965.

PR #284, opened 2018-04-06
Report unknown JSON fields, to catch errors early.

adelton:
Unfortunately, for YAML, we do not have the `UnmarshalStrict` available.


adelton:
Unfortunately, travis suggests that this seems to require golang 1.10. Does it make sense to put this on hold until 1.10 is the lowest version used for cri-tools?

feiskyer:
@adelton The changed files are in vendors. We don't change vendor files this way. Please file a PR to kubernetes repo (the vendored file is here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/yaml/decoder.go), and then update vendors here after it merged.

adelton:
Thank you. I'll file against Kubernetes, probably when it moves to golang 1.10.

PR #289, opened 2018-04-17
Add host network portforward test.

Random-Liu:
We recently found a bug that after getting rid of `nsenter`, containerd doesn't support portforward for hostnetwork pod. https://github.com/containerd/cri/issues/738

This should be tested in CRI validation test. But this test should not be `Conformance`, because it requires host network.

### DIFF
--- a/pkg/validate/networking.go
+++ b/pkg/validate/networking.go
@@ -30,14 +30,19 @@ import (
 )
 
 const (
-	defaultDNSServer            string = "10.10.10.10"
-	defaultDNSSearch            string = "google.com"
-	defaultDNSOption            string = "ndots:8"
-	resolvConfigPath            string = "/etc/resolv.conf"
-	nginxImage                  string = "nginx"
-	nginxContainerPort          int32  = 80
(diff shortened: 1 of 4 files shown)


pkg/validate/networking.go
```diff
@@ -30,14 +30,19 @@ import (
...
-	nginxHostPortForPortForward int32  = 8001
+	defaultDNSServer   string = "10.10.10.10"
+	defaultDNSSearch   string = "google.com"
+	defaultDNSOption   string = "ndots:8"
+	resolvConfigPath   string = "/etc/resolv.conf"
+	nginxImage         string = "nginx"
+	hostNetNginxImage  string = "gcr.io/cri-tools/hostnet-nginx"
+	nginxContainerPort int32  = 80
+	// The following host ports must not be in-use when running the test.
+	nginxHostPortForPortMapping        int32 = 8000
```
feiskyer: 8000 is usually used for local services. Replace those ports above 10000?

Random-Liu:
@feiskyer Are you OK with this? :)

PR #293, opened 2018-04-26
Update kubernetes to fix `crictl logs`.

Random-Liu:
Update kubernetes version to include fix https://github.com/kubernetes/kubernetes/pull/63186. Because CRI is not changed in K8s, it should be safe to update K8s to HEAD.

I've verified that this change will get rid of the log line `I0426 08:34:13.679995    6949 remote_runtime.go:43] Connecting to runtime service unix:///run/containerd/containerd.sock`.

This PR also update the logrus dependency to `v1.0.0`, because the k8s packages vendored in don't have dependency on logrus anymore. We need to specify logrus dependency in our own vendor.conf now.


Random-Liu:
This PR also change travis to only use golang 1.10.

Kubernetes now requires 1.10 now. https://github.com/kubernetes/kubernetes/pull/63046

feiskyer:
Gofmt check failed. @Random-Liu Need to run gofmt

Random-Liu:
@feiskyer Fixed.

PR #300, opened 2018-05-05
Make image-user test images multi-arch

### LINKED ISSUE #287: Make test-images multiarch
The "make test-cri" command fails on non-x86 architecture because of the images created from image-test and image-user directory are not multi-arch. If these images are made multi-arch when built and pushed, then it would be easy to test on other non-x86 architectures as well

nitkon:
Fixes #287 
This fixes Issue 287 partially. Here "docker push" is used instead of
"gcloud docker -- push" as its depreciated now. [Reference](https://github.com/kubernetes/kubernetes/commit/c0b73645637c248091358a37c7e99bfcfc983f71)

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>

### DIFF
--- a/images/image-user/Makefile
+++ b/images/image-user/Makefile
@@ -12,22 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+include ../../hack/make-rules/Makefile.manifest
+include ../../hack/make-rules/BASEIMAGES
+
 .PHONY: all test-image-user-uid test-image-user-username test-image-user-uid-group test-image-user-username-group
+
+REGISTRY = gcr.io/cri-tools
+ALL_ARCH = amd64 arm arm64 ppc64le s390x
+TAG = latest
+IMAGES_LIST = test-image-user-uid test-image-user-username test-image-user-uid-group test-image-user-username-group
 
 all: test-image-user-uid test-image-user-username test-image-user-uid-group test-image-user-username-group
 
 test-image-user-uid:
-	docker build . -t gcr.io/cri-tools/$@ --build-arg USER=1002
-	gcloud docker -- push gcr.io/cri-tools/$@
+	$(foreach arch,$(ALL_ARCH),docker build . -t $(REGISTRY)/$@-$(arch) --build-arg ARCH=$($(arch)) --build-arg USER=1002;)
+	$(foreach arch,$(ALL_ARCH),docker push $(REGISTRY)/$@-$(arch);)
 
 test-image-user-username:
-	docker build . -t gcr.io/cri-tools/$@ --build-arg USER=www-data
-	gcloud docker -- push gcr.io/cri-tools/$@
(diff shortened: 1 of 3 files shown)


images/image-user/Makefile
```diff
@@ -12,22 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all test-image-user-uid test-image-user-username test-image-user-uid-group test-image-user-username-group
+include ../../hack/make-rules/Makefile.manifest
+
+.PHONY: all test-image-user-uid test-image-user-username test-image-user-uid-group test-image-user-username-group push-manifest
+
+REGISTRY = gcr.io/cri-tools
+ALL_ARCH = amd64 ppc64le
```
feiskyer: Add all arches that kubernetes has already supported? e.g. `amd64 arm arm64 ppc64le s390x`

images/image-user/Dockerfile
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM busybox
+ARG ARCH
+FROM $ARCH/busybox
```
feiskyer: seems there are no `arm/busybox` and `arm64/busybox` images?

images/image-user/Makefile
nitkon: @feiskyer : Should we use arm32v7 and arm64v8 ? [Reference](https://github.com/docker-library/official-images#architectures-other-than-amd64)

images/image-user/Dockerfile
nitkon: @feiskyer : Should we use arm32v7 and arm64v8 ? [Reference](https://github.com/docker-library/official-images#architectures-other-than-amd64)

mkumatag: yes we have to use `arm32v7` and `arm64v8`

nitkon:
@feiskyer : @mkumatag :  Addressed your comments.

images/image-user/Makefile
```diff
@@ -12,22 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+include ../../hack/make-rules/Makefile.manifest
+include ../../hack/make-rules/Baseimage
```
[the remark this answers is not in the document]
nitkon: @feiskyer :  Thnx! Fixed.

feiskyer:
thanks

/lgtm

feiskyer:
@Random-Liu Could you build and push those new images?

mkumatag:
It depends on the `.docker/config` file entry with proper credentials to login `gcr.io`. So you need to check whether docker command is able to push the image to gcr or not.

nitkon:
@Random-Liu : I have addressed your comment in the following [PR](https://github.com/kubernetes-incubator/cri-tools/pull/311)

Random-Liu:
> It depends on the .docker/config file entry with proper credentials to login gcr.io. So you need to check whether docker command is able to push the image to gcr or not.

I did `gcloud credential` thing, let me try again tonight.

PR #302, opened 2018-05-07
Fix Go versin

feiskyer:
1.10.2 is required now to build kubernetes master.


nitkon:
continuous-integration/travis-ci was failing because of the following. 
_Kubernetes requires go1.10.2 or greater.
Please install go1.10.2 or later._

LGTM

PR #320, opened 2018-06-04
Add new container host path test.

Random-Liu:
Change CRI validation test to conform to https://github.com/kubernetes/kubernetes/pull/61460.


Random-Liu:
It seems that standalone dockershim doesn't work now. Will investigate later.

feiskyer:
Yep, kubelet --experimental-dockershim will exit silently. Let me file a PR to fix that

Random-Liu:
Removed the non-existent host path test. Let's finalize https://github.com/kubernetes/kubernetes/pull/61460#discussion_r192981712 first.

Random-Liu:
@feiskyer Please take a look. This PR only adds symlink host path test now.

PR #327, opened 2018-06-18
Bump cri-tools to v1.11.0

### LINKED ISSUE #297: Release and versioning
The current version schema is pretty confusing.

We have branch which follows Kubernetes release `release-1.10`, `release-1.11` etc.

We also have a version `v1.0.0-alpha.0`, `v1.0.0-beta.0` etc.

I think `critest` and `crictl` have different CRI version compatibility requirement:
* `critest` is used to test all CRI features. A new field or slightly behavior change in CRI usually means a new corresponding CRI validation test, but doesn't necessarily mean a CRI version update, e.g. `RunAsGroup`. So I think `critest` version has to follow Kubernetes version.
* `crictl` should be able to work with any CRI container runtime, as long as the CRI version matches. There might be field ignored or defaulted because of version skew, but things should not break. So I think `crictl` version can potentially follow CRI version in the future.

For now, given that CRI version schema is still not clear yet, let's graduate cri-tools to GA, and start following Kubernetes version for both `critest` and `crictl`.

@yujuhong @feiskyer @mrunalp @runcom @mikebrow

feiskyer:
Bumps cri-tools to v1.11.0 and updates changelog for the new release.

Closes #297.

/assign @Random-Liu 

cc @luxas @jberkus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,20 @@
 <!-- TOC -->
-- [v1.0.0-beta.1](#v100-beta1)
-- [v1.0.0-beta.0](#v100-beta0)
+
+- [v1.11.0](#v1110)
     - [CRI validation testing (critest)](#cri-validation-testing-critest)
     - [CRI CLI (crictl)](#cri-cli-crictl)
-- [v1.0.0-alpha.0](#v100-alpha0)
+- [v1.0.0-beta.1](#v100-beta1)
     - [CRI validation testing (critest)](#cri-validation-testing-critest-1)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,20 @@
 <!-- TOC -->
+
+- [v1.11.0](#v1110)
+        - [CRI validation testing (critest)](#cri-validation-testing-critest)
```
Random-Liu: Match indent

feiskyer: Fixed

luxas:
Thanks!

PR #332, opened 2018-06-23
Allow filtering the pods with prefix matching of name and namespace

yujuhong:


Random-Liu:
Can we change to regular expression instead? Seems more useful. :)

yujuhong:
> Can we change to regular expression instead? Seems more useful. :)

Done. PTAL

PR #333, opened 2018-06-26
adds id text to each command

mikebrow:
While using crictl @estesp reported a nit with the UI.. particularly that it would be nice if we told users POD-ID was needed. As is it says POD, which is confusing. 

I checked and we're just not consistent across the UI so here's a PR to fix that.

@Random-Liu 

address issue #334 


feiskyer:
lgtm

PR #339, opened 2018-07-11
Set version based on git tags.

Random-Liu:
I want to release a 1.11.1 from `release/1.11`. However, I don't know whether I should only update the version number in `release/1.11` or update both `release/1.11` and master.

If we only update `release/1.11`, master will still stay at `1.11.0`.
But if we update both, master isn't actually at `1.11.1`.

This PR changes to use `git tags` instead, which contains commit number for non-tag commits.

@yujuhong @feiskyer 

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,20 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+// Version holds the complete version number. Filled in at linking time.
+var Version = "unknown"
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+// Version holds the complete version number. Filled in at linking time.
+var Version = "1.11.0+unknown"
```
yujuhong: Do we need to keep updating this `Version`?

Random-Liu: That's how containerd does it. We can keep this as a simple `unknown` string. :)

yujuhong: This looks like a placeholder so I just wondered what we're going to do with it

Random-Liu: I'll change this to `unknown`.

Random-Liu:
Apply LGTM based on https://github.com/kubernetes-incubator/cri-tools/pull/339#issuecomment-404327854

feiskyer:
LGTM.

ncdc:
@Random-Liu does `hack/release.sh` need to be updated to set the ldflags with the version info?

feiskyer:
yep, @ncdc would you like make a PR for that?

ncdc:
Sure, I can do it today

ncdc:
PR is #367

PR #349, opened 2018-07-13
Cherrypick #339 to release-1.11.

Random-Liu:
I missed cherrypick of #339.

I already manually updated the version for v1.11.1. Cherrypick this to make future release correct.


feiskyer:
release-1.11 is still using master kubelet for CI. Could you change that to kubernetes release-1.11 in file `hack/install-kubelet.sh`?

Random-Liu:
Sure. Will do in this PR.

Random-Liu:
@feiskyer I decided to do in HEAD first.

We can do better than hard code in the script. :) https://github.com/kubernetes-incubator/cri-tools/blob/release-1.10/hack/install-kubelet.sh#L21

feiskyer:
Yep, +1 for the new method.

PR #354, opened 2018-07-13
Properly returns errors when the output format is not supported

yujuhong:


Random-Liu:
NVM, there is no panic... I misremembered. Then we may not need a cherry-pick.

PR #363, opened 2018-08-03
Fix crictl ps and crictl pods.

Random-Liu:
@yanxuean This should fix the issue for you.


yanxuean:
I am late.   It will be better if we can remove "table" and "default" case.  
But these code was added in https://github.com/kubernetes-incubator/cri-tools/pull/206. 
In fact, I don't know what it mean:)   Maybe @mikebrow know it.

PR #370, opened 2018-08-24
crictl: Fallback if info key is not JSON

mrunalp:
This allows info map to support non JSON keys


mrunalp:
@Random-Liu This PR implements what we discussed.
cc: @runcom

feiskyer:
@mrunalp Any use cases of this?

PR #374, opened 2018-09-04
A few fixes to defaults on Windows.

jterry75:
Makes the default config location GOOS dependent to support Windows
paths if the tool is running on Windows.

Sets the default connection endpoint for Windows to be
\\.\pipe\dockershim.


jterry75:
@Random-Liu - Working on getting the CLA signed.

Random-Liu:
It seems that the library we are using already supported windows. https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/util

jterry75:
@Random-Liu - Only supports TCP see [util_windows.go#L48](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/util/util_windows.go#L48). Given that CRI is local only it should use npipe on Windows.

Random-Liu:
@jterry75 I see. If containerd is going to provide npipe on Windows, we should eventually update the kubelet util library to support it. But I'm fine with starting from crictl/critest for development first. :)

jterry75:
@Random-Liu - I pushed and update that keep the old behavior if indeed people need tcp support for crictl on Windows. This will just inspect the EP and return an dialer for NamedPipes if the EP is a pipe path otherwise uses the util package as it used to.

feiskyer:
dockershim on Windows listens to `"tcp://localhost:3735"`, so tcp should also be supported on Windows.

jterry75:
@feiskyer - I can update it there that's fine. This change would support both tcp and npipe on Windows now though. Do you see any issues with that?

jterry75:
@feiskyer @Random-Liu Here is a change in my private fork of Kubernetes as well: [jterry75/kubernetes:a2f144](https://github.com/jterry75/kubernetes/commit/a2f1f4489dc69a9143e5cbe5b78515037134b0de). I assume this is the right long term solution that you were looking for?

jterry75:
@feiskyer - Changes the default to tcp as you requested but leaves support for npipe

feiskyer:
@jterry75 Thanks. LGTM, but let's wait a while for #376 because the repo has been moved to new org.

cmd/crictl/main_unix.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"k8s.io/kubernetes/pkg/kubelet/util"
+)
+
+const (
+	defaultConfigPath      = "/etc/crictl.yaml"
+	defaultRuntimeEndpoint = "unix:///var/run/dockershim.sock"
+)
+
+func GetAddressAndDialer(endpoint string) (string, func(addr string, timeout time.Duration) (net.Conn, error), error) {
```
feiskyer: Could you also add comments for this function?

jterry75: Absolutely!

jterry75:
@feiskyer - PTAL. Rebased on latest master and added comments as requested.

feiskyer:
@jterry75 Thanks.

/lgtm

PR #377, opened 2018-09-12
Add new test to critest for privileged container

umohnani8:
Test ensures that when a privileged container is run,
the mount label persists and is in the right format.
Also checks the process label.

Adds test for https://github.com/kubernetes-sigs/cri-o/pull/1781

### DIFF
--- a/pkg/validate/selinux_linux.go
+++ b/pkg/validate/selinux_linux.go
@@ -17,6 +17,7 @@ limitations under the License.
 package validate
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/kubernetes-sigs/cri-tools/pkg/framework"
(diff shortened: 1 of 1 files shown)


umohnani8:
@mrunalp PTAL

pkg/validate/security_context.go
```diff
@@ -1133,3 +1151,24 @@ func checkSetHostname(rc internalapi.RuntimeService, containerID string, setable
...
+	Expect(string(stdout)).To(ContainSubstring("object_r:container_file_t"))
+
+	// Check that the correct process label is set for privileged and unprivileged containers
+	cmd = []string{"cat", "/proc/self/attr/current"}
+	stdout, stderr, err = rc.ExecSync(containerID, cmd, time.Duration(defaultExecSyncTimeout)*time.Second)
+	msg = fmt.Sprintf("cmd %v, stdout %q, stderr %q", cmd, stdout, stderr)
+	Expect(err).NotTo(HaveOccurred(), msg)
+
+	if privileged {
+		Expect(string(stdout)).To(ContainSubstring("system_r:spc_t"))
```
runcom: not sure all this is going to work on non rhel systems (like fedora).

btw, I was thinking about having 2 separate tests:

- mount label: just have a cmd for the pod that does `ls -ldZ /` and make sure level is applied even if privileged
- process label: just have a cmd that `cat /proc/self/attr/current`

all of that to avoid having pods running `top`

mrunalp: For mount_label we can grep for it in /proc/1/mountinfo.

feiskyer: The tests failed on Travis (based on Ubuntu):

```
Expected
        <string>: lrwxrwxrwx 1 root root ? 7 Feb  7  2018 bin -> usr/bin
    to contain substring
        <string>: object_r:container_file_t
    /home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/security_context.go:1161
```

pkg/validate/security_context.go
```diff
@@ -451,6 +451,24 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			checkNetworkManagement(rc, containerID, notPrivileged)
 		})
 
+		It("selinux mount label should persist when container is privileged", func() {
```
feiskyer: could you move this case to `selinux_linux.go`?

umohnani8: Moved.

pkg/validate/security_context.go
umohnani8: The labels work fine on my fedora 28. Also split the it into 2 separate tests.

runcom: ubuntu doesn't have selinux enabled/installed by default for sure

umohnani8: @runcom I moved the test to selinux_linux.go, which checks if selinux is enabled before running the tests. That should be good right?

runcom: yup yup! sorry overlooked that :+1:

PR #384, opened 2018-09-19
Fix grpc dial timeout

mfpierre:
The `WithTimeout` option in `grpc.Dial` is only valid if `WithBlock()` is present see https://godoc.org/google.golang.org/grpc#WithTimeout

As a result using `crictl` on a non existing socket with this fix has a different behavior because the timeout is honored:

before: `FATA[0000] getting status of runtime failed: rpc error: code = Unavailable desc = grpc: the connection is unavailable`

after: `FATA[0010] failed to connect: failed to connect: context deadline exceeded`

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -57,7 +57,7 @@ func getRuntimeClientConnection(context *cli.Context) (*grpc.ClientConn, error) 
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


feiskyer:
@mfpierre Thanks for the fix.

/lgtm
/approve

feiskyer:
There is also such issues in kubelet CRI clients. Let me file a PR for that in k/k.

PR #427, opened 2019-01-21
crictl stats: complain if user is not root

lsm5:
$ ./bin/crictl stats
FATA[0000] You're probably not root, your effective user id is 1000

 $ sudo ./bin/crictl stats
CONTAINER           CPU %               MEM                 DISK                INODES
07ac205b107e5       0.00                2.273MB             0B                  0

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>


lsm5:
@mrunalp PTAL

mrunalp:
We can close this in favor of #428

PR #428, opened 2019-01-23
Reduce the default connection timeout value

mrunalp:
In interactive use, if user doesn't sudo or running as root
then has to wait too long before getting a connection failure.

@runcom @Random-Liu ptal
cc: @lsm5


runcom:
:+1: 

LGTM

PR #431, opened 2019-01-30
Change exec-sync on exit code != 0

saschagrunert:
Hey, feel free to shut me down if the current implementation is the totally intended behavior. As a user I would expect that crictl fails if the return value of exec-sync is also not 0.


runcom:
I'm :+1: on this overall actually

feiskyer:
+1 with this

/lgtm
/approve

PR #434, opened 2019-02-11
update link to point to correct godep instructions

yangkev:
This fixes the broken link to point to the new docs for dependency management.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,4 +23,6 @@ Follow either of the two links above to access the appropriate CLA and instructi
 
 ### Adding dependencies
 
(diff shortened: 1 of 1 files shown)


yangkev:
Seems like this might not be pertinent since `vndr` is being used instead of `godep`.

feiskyer:
Yep, could you update the guides to `vndr` instead?

yangkev:
Sounds good.

I updated the doc to point to [`vndr`](https://github.com/LK4D4/vndr), as well as the general k8s dependency management (with `godep`) docs for best practices. I wonder if it is too confusing that there are two tools mentioned (`vndr` and `godep`) now?

feiskyer:
> I wonder if it is too confusing that there are two tools mentioned (`vndr` and `godep`) now?

What do you think of spliting the docs into two parts: (1) restore the versions via godep; (2) update vendors for cri-tools with `vndr` (`hack/update-vendor.sh` will do this).

feiskyer:
close-reopen for re-trigger Travis CI

PR #442, opened 2019-03-11
add simple completion support for zsh

### LINKED ISSUE #435: crictl completion for zsh
#213 states that installing the completion using `source <(crictl completion)` works for both bash and zsh. Unfortunately this is only true if `bashcompinit` has been loaded in your zshrc. Depending on your zsh setup this might or might not be the case when you source the output of the completion sub-command.

As far as i undertstand this could be fixed by adding the contents of this [file](https://github.com/urfave/cli/blob/master/autocomplete/zsh_autocomplete) to [cmd/crictl/completion.go](../blob/master/cmd/crictl/completion.go) and add a parameter to the completion sub-command that selects between bash or zsh.

cpointner:
fixes #435.

### DIFF
--- a/cmd/crictl/completion.go
+++ b/cmd/crictl/completion.go
@@ -34,33 +34,95 @@ var bashCompletionTemplate = `_cli_bash_autocomplete() {
 
 complete -F _cli_bash_autocomplete crictl`
 
+func bashCompletion(c *cli.Context) error {
+	subcommands := []string{}
+	for _, command := range c.App.Commands {
+		if command.Hidden {
+			continue
+		}
+		for _, name := range command.Names() {
+			subcommands = append(subcommands, name)
+		}
+	}
+
+	for _, flag := range c.App.Flags {
+		// only includes full flag name.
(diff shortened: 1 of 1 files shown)


cmd/crictl/completion.go
```diff
@@ -34,33 +34,95 @@ var bashCompletionTemplate = `_cli_bash_autocomplete() {
...
-		fmt.Fprintln(c.App.Writer, fmt.Sprintf(bashCompletionTemplate, strings.Join(subcommands, "\n")))
-		return nil
+		shell := c.Args().First()
+		switch shell {
+		case "bash":
+			return bashCompletion(c)
+		case "zsh":
+			return zshCompletion(c)
+		default:
+			return fmt.Errorf("unknown shell: %q", shell)
```
feiskyer: could we log only bash and zsh supported?

cpointner: yeah you are right. Just pushed the change, do you want me to squash it together?

feiskyer: thanks

feiskyer:
close-reopen for re-trigger Travis CI

PR #453, opened 2019-03-25
Add `curl` to download example

bennythejudge:
Add an example to download `crictl` using `curl` instead of `wget`.

### DIFF
--- a/docs/crictl.md
+++ b/docs/crictl.md
@@ -20,7 +20,7 @@ rm -f crictl-$VERSION-linux-amd64.tar.gz
 - using `curl`:
 
 ```sh
-VERSION="v1.11.1"
+VERSION="v1.13.0"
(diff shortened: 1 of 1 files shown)


docs/crictl.md
```diff
@@ -15,6 +17,15 @@ sudo tar zxvf crictl-$VERSION-linux-amd64.tar.gz -C /usr/local/bin
 rm -f crictl-$VERSION-linux-amd64.tar.gz
 ```
 
+- using `curl`:
+
+```sh
+VERSION="v1.11.1"
```
feiskyer: could we use the same version v1.13.0 as above

bennythejudge: Thanks for the comment, I have made the change. 
I will resolve this convo.

PR #458, opened 2019-04-04
Set SelinuxRelabel to true for bind mounts

umohnani8:
The volume mount tests were not setting selinuxRelabel to
true, which was causing the paths mounted in to not have the
correct label. We caught this with the SELinux rework we are
doing in cri-o.

The Selinux rework PR: https://github.com/cri-o/cri-o/pull/2096


umohnani8:
@rhatdan @mrunalp PTAL

rhatdan:
This should be a noop if SELinux is disabled.  Currently if SELinux is enabled and the containers are running with a locked down container type like container_t they would get permission denied when trying to read/write these volumes.  With the Relabel, the volumes will have a label that the container_t processes are allowed to use.

PR #459, opened 2019-04-11
Improve error message for failed to connect

umohnani8:
Add more info to the error message when crictl fails to connect
to the runtime either due to permissions or the runtime not being
started.


umohnani8:
@mrunalp PTAL

PR #462, opened 2019-05-10
Fix log directory typo in pod config examples.

congliu01:
"log_directory" is the right key instead of "logDirectory".


congliu01:
It looks I already have CCLA through my employer.
> Contribute to CNCF as an employee
You are already authorized to contribute code to this project through your membership with Google.

PR #465, opened 2019-05-29
Update `rmp` to test for running sandbox

### LINKED ISSUE #464: Stop sandbox implicitly on `rmp`
A running sandbox should be stopped before removal so allow proper unmounting of all devices. I would like to see this feature in crictl, WDYT?

saschagrunert:
Running pod sandboxes have to be stopped before the actual removal. To
achieve a better user-experience, we now error in that case and give the
user the option to ignore it with the added `-f` flag for `rmp`.

Closes #464

### DIFF
--- a/cmd/crictl/sandbox.go
+++ b/cmd/crictl/sandbox.go
@@ -103,16 +103,38 @@ var removePodCommand = cli.Command{
 	Name:      "rmp",
 	Usage:     "Remove one or more pods",
 	ArgsUsage: "POD-ID [POD-ID...]",
-	Action: func(context *cli.Context) error {
-		if context.NArg() == 0 {
-			return cli.ShowSubcommandHelp(context)
+	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name:  "force, f",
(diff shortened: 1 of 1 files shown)


cmd/crictl/sandbox.go
```diff
@@ -103,16 +103,33 @@ var removePodCommand = cli.Command{
...
-			err := RemovePodSandbox(runtimeClient, id)
+		for i := 0; i < ctx.NArg(); i++ {
+			id := ctx.Args().Get(i)
+
+			resp, err := runtimeClient.PodSandboxStatus(context.Background(),
+				&pb.PodSandboxStatusRequest{PodSandboxId: id})
+			if err != nil {
+				return err
+			}
+			if !ctx.Bool("force") &&
```
feiskyer: if force is true and sandbox state is READY, then invoke `StopPodSandbox()` before `RemovePodSandbox`?

saschagrunert: Yes that sounds good, I will change it as adapted.

feiskyer:
The CI has been passed from travis page, but didn't reflect the status from github hook. Merged this manually.

PR #467, opened 2019-05-31
support live monitoring critool stats

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

Add support for live critool stats;
Also refactor the table display

### DIFF
--- a/cmd/crictl/display.go
+++ b/cmd/crictl/display.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+)
+
+const (
+	columnContainer = "CONTAINER"
+	columnImage     = "IMAGE"
+	columnImageID   = "IMAGE ID"
+	columnCreated   = "CREATED"
+	columnState     = "STATE"
+	columnName      = "NAME"
+	columnAttempt   = "ATTEMPT"
+	columnPodID     = "POD ID"
+	columnNamespace = "NAMESPACE"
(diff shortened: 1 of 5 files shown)


yeya24:
ping @feiskyer

cmd/crictl/display.go
```diff
@@ -0,0 +1,50 @@
...
+	imageID   = "IMAGE ID"
+	created   = "CREATED"
+	state     = "STATE"
+	name      = "NAME"
+	attempt   = "ATTEMPT"
+	podID     = "POD ID"
+	namespace = "NAMESPACE"
+	size      = "SIZE"
+	tag       = "TAG"
+	digest    = "DIGEST"
```
feiskyer: Could you add a prefix for those const names?

cmd/crictl/image.go
```diff
@@ -19,16 +19,13 @@ package main
...
-
-	units "github.com/docker/go-units"
+	"github.com/docker/go-units"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"golang.org/x/net/context"
 	pb "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
+	"sort"
+	"strings"
 )
```
feiskyer: nit: move those two to the first group

cmd/crictl/stats.go
```diff
@@ -158,36 +156,40 @@ func ContainerStats(client pb.RuntimeServiceClient, opts statsOptions) error {
...
-			continue
-		}
-		var cpuPerc float64
-		if cpu != 0 {
-			// Only generate cpuPerc for running container
-			duration := s.GetCpu().GetTimestamp() - old.GetCpu().GetTimestamp()
-			if duration == 0 {
-				return fmt.Errorf("cpu stat is not updated during sample")
+	display := newTableDisplay(20, 1, 3, ' ', 0)
+	for range time.Tick(500 * time.Millisecond) {
```
feiskyer: Could we add a new option for this, e.g. `--watch` or `-w`?

yeya24: I think live monitoring is enabled by default, like `docker stats`. So do we still need this watch option?

yeya24: WDYT? @feiskyer Do we really need this option? What is the use case of non live monitoring?

feiskyer: just like kubectl top, it only show the output for one time. We can enable it by default, but it could be disabled by an option

yeya24: Thanks for your reply, I have added this -w option to enable live monitoring, PTAL.

feiskyer:
thanks. 
/lgtm
/approve

feiskyer:
Close and re-open to trigger the failed travis CI.

PR #471, opened 2019-06-12
Fallback to repo digest name if repoTag is <none>:<none>

saschagrunert:
It could be possible that the repo tags are defined with "<none>:<none>". In that case, we should still try to get the image name from the repo digest field to provide the highest level of information to the user.


saschagrunert:
It looks like that the test fails on other branches as well. This does not seem to be a valid ID:
```
 Id: "sha256:1d42c7758256ea94f23224403e71455e512de249b689426a5516c56e9446207f",
```
:innocent:

feiskyer:
Close and re-open to trigger the failed travis CI.

PR #477, opened 2019-06-26
Add registry prefix option

saschagrunert:
This allows to pre-pull the images and running benchmarks against a
local registry. 

Beside this, the succinct output for benchmarks seems more reasonable
to me, so we put it on per default.


pkg/framework/util.go
```diff
@@ -297,6 +297,7 @@ func ListImage(c internalapi.ImageManagerService, filter *runtimeapi.ImageFilter
 
 // PullPublicImage pulls the public image named imageName.
 func PullPublicImage(c internalapi.ImageManagerService, imageName string, podConfig *runtimeapi.PodSandboxConfig) string {
+	imageName = TestContext.RegistryPrefix + imageName
```
feiskyer: the imageName may already contains the image registry, e.g. `"gcr.io/cri-tools/test-image-1:latest"`. So adding another prefix may not work. I think we should parse the image first and replace the registry with `TestContext.RegistryPrefix`.

saschagrunert: I agree, I changed the implementation to parse the image name before. Since the parsing expects a fully qualified image name, we now fallback to the intial `imageName` if the parsing fails. If not, we split up the domain and replace it.

saschagrunert:
Please take a look again, I had to change the benchmark images to be fully qualified to make the registry prefix replacement work.

pkg/framework/util.go
```diff
@@ -297,9 +298,21 @@ func ListImage(c internalapi.ImageManagerService, filter *runtimeapi.ImageFilter
 
 // PullPublicImage pulls the public image named imageName.
 func PullPublicImage(c internalapi.ImageManagerService, imageName string, podConfig *runtimeapi.PodSandboxConfig) string {
-	if !strings.Contains(imageName, ":") {
-		imageName = imageName + ":latest"
-		Logf("Use latest as default image tag.")
+
+	ref, err := reference.ParseNamed(imageName)
```
feiskyer: report an error when `err != nil`? In what cases, the error would happen?

saschagrunert: Hm, yeah I guess we have to handle three possible paths here:

- The image name is fully qualified, then we parse it and modify the domain
- The image name is not fully qualified, then we can prepend the registry-prefix
- The image name is wrong, then we fail.

I adapted the implementation to work like this.

feiskyer: thanks

feiskyer:
Tests failed:

```
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/pkg/framework/framework.go:72
  SeccompProfilePath
  /home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/security_context_linux.go:483
    should support seccomp unconfined on the container [AfterEach]
    /home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/security_context_linux.go:529
    failed to pull image: rpc error: code = Unknown desc = invalid reference format
    Unexpected error:
        <*status.statusError | 0xc000401fb0>: {
            Code: 2,
            Message: "invalid reference format",
            Details: nil,
        }
        rpc error: code = Unknown desc = invalid reference format
    occurred

```

saschagrunert:
I fixed the failing CI, in the end we would now need fully qualified images for every test. WDYT?

feiskyer:
> I fixed the failing CI, in the end we would now need fully qualified images for every test. WDYT?

Could we default to `docker.io` if the registry is not included in the image?

saschagrunert:
> Could we default to `docker.io` if the registry is not included in the image?

Yeah I think so, I will make `docker.io/library` the default prefix and if that not changes, then the tests should need no further adaption.

It's a bit tricky, sorry for all the different kind of approaches :innocent:

PR #488, opened 2019-07-28
Switch to multi-arch nonewprivs image.

nitkon:
The current image in use "gcr.io/google_containers/nonewprivs:1.2"
"exec format error" on non amd64 archs. Thus running tests on such
arch fails. Lets move to a multi-arch image then.

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>


nitkon:
/assign @yujuhong
cc @feiskyer

PR #490, opened 2019-07-31
Switch to go-modules

saschagrunert:
This removes the old vendoring for go module support. We also add an
additional verification step to travis. Documentation has been updated
as well.

---
After that, I would like to update to the latest Kubernetes alpha to propose the new CRI API changes which are introduced with dual stack IPv6 support, WDYT?


feiskyer:
The idea looks good, but I wonder why so many vendor files are removed and updated? Generally, if we only switch go modules, the vendor files should still be same, right?

saschagrunert:
> The idea looks good, but I wonder why so many vendor files are removed and updated? Generally, if we only switch go modules, the vendor files should still be same, right?

Ah I'm sorry, I also updated the dependencies as well. The thing is that dependent projects switched to go modules too, which will lead into ma much cleaner go.mod file.

feiskyer:
@saschagrunert Thanks, then it makes sense.

hack/tree_status.sh
```diff
@@ -14,17 +14,14 @@
...
-cd ${ROOT}
-
-echo "Run vndr to update vendor"
-vndr -whitelist github.com/onsi/ginkgo/*
-
-echo "Sort vendor.conf..."
-sort vendor.conf -o vendor.conf
-
-echo "Please commit the change made by this file..."
+STATUS=$(git status --porcelain)
```
feiskyer: I think we should also run go mod command before checking git status, e.g. 
```
$(GO) mod tidy && \
		$(GO) mod vendor && \
		$(GO) mod verify
```

hack/tree_status.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
feiskyer: could we rename this file to hack/verify-go-modules.sh?

go.mod
```diff
@@ -0,0 +1,53 @@
...
+	github.com/urfave/cli v0.0.0-20180821064027-934abfb2f102
+	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80
+	golang.org/x/sys v0.0.0-20190730183949-1393eb018365
+	google.golang.org/grpc v1.22.1
+	gopkg.in/yaml.v2 v2.2.2
+	k8s.io/api v0.0.0
+	k8s.io/apimachinery v0.0.0
+	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/cri-api v0.0.0
+	k8s.io/kubernetes v0.0.0
```
feiskyer: v0.0.0 means latest. For k8s.io dependencies, do we need to change this to a specific version, e.g. v1.15.0?

hack/tree_status.sh
saschagrunert: Okay, added this as well.

hack/tree_status.sh
saschagrunert: Yes, changed it.

go.mod
saschagrunert: Not necessarily since we override it in the `replace` section: `k8s.io/kubernetes => k8s.io/kubernetes v1.15.1`

PR #494, opened 2019-08-02
Move from gometalinter to golangci-lint

saschagrunert:
gometalinter has been deprecated so we can move to golangci-lint as
successor project. The installation of the tool is not necessary any
more, since we vendor the dependency directly into the repository. This
allows more reproducible lint results. All linters which currently pass
have been enabled. The gofmt and govet lint scripts have been removed
because they are now included in golangci-lint as well.


feiskyer:
CI is still failing. Not sure what is introduced, @saschagrunert could you check why the CI is failing?

saschagrunert:
> CI is still failing. @saschagrunert could you check why the CI is failing?

Yes, I'm just wondering if the issues are the same like on the master branch. I'll do some investigation now.

saschagrunert:
The windows issue is fixed, but the ppc critest seems flaky. I'm checking further test cases...

saschagrunert:
It works now, if the flakes coming again I think we have to re-evaluate the critest related integration tests...

feiskyer:
> It works now, if the flakes coming again I think we have to re-evaluate the critest related integration tests...

Looks good to me.

PR #496, opened 2019-08-02
Pull image on container creation

saschagrunert:
I'm not sure if there were already some discussions if we want this or not, so I'd like to propose this one: We now try to pull the image from the container manifest before actually creating the container. This makes the overall usability a bit nicer.

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -51,15 +51,49 @@ type createOptions struct {
 type runOptions struct {
 	// configPath is path to the config for container
 	configPath string
+
 	// podConfig is path to the config for sandbox
 	podConfig string
+
+	// the image pull options
+	*pullOptions
+}
+
+type pullOptions struct {
+	// do not pull the image on container creation
+	dontPull bool
+
+	// creds is string in the format `USERNAME[:PASSWORD]` for accessing the
+	// registry during image pull
+	creds string
+
+	// auth is a base64 encoded 'USERNAME[:PASSWORD]' string used for
+	// authentication with a registry when pulling an image
+	auth string
+}
+
+var pullFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "no-pull",
+		Usage: "Do not pull the image on container creation",
+	},
+	cli.StringFlag{
+		Name:  "creds",
+		Value: "",
+		Usage: "Use `USERNAME[:PASSWORD]` for accessing the registry",
+	},
+	cli.StringFlag{
+		Name:  "auth",
+		Value: "",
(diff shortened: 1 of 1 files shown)


cmd/crictl/container.go
```diff
@@ -432,13 +446,19 @@ func CreateContainer(client pb.RuntimeServiceClient, opts createOptions) (string
 		}
 	}
 
+	// Try to pull the image before container creation
+	image := config.GetImage().GetImage()
+	if _, err := PullImage(iClient, image, nil); err != nil {
```
feiskyer: I like the idea, but if local image is used, the pull operation would fail depending on the runtime implementations.

So, I think we'd better add an option here, so that the pull can be skipped in some cases?

saschagrunert: Yes, I agree. I added the option to `no-pull` as well as the pull authentication related options (this should work too I guess).

feiskyer: thanks

mikebrow:
But the command is create container not pull image. Oh man.. this is a huge performance hit and a huge behavior change, by default. Kublet has image pull policies..

saschagrunert:
> But the command is create container not pull image. Oh man.. this is a huge performance hit and a huge behavior change, by default. Kublet has image pull policies.. did they add --no-pull?
> 
> quick check.. nope..

Hm okay to which "they" are you referring?

mikebrow:
kublet

mikebrow:
https://kubernetes.io/docs/concepts/containers/images/#pre-pulled-images

mikebrow:
sry just trying to figure out what this means...

saschagrunert:
Okay, the main intention of such PRs, like this one or the one added `crictl run`, is to mimic parts of the kubelet to be able to use `crictl` in a more convenient fashion. If any of those behavioral changes introduces a regression then we can find a way to turn them off by default.

Sometimes avoiding breaking changes is not that easily achievable in conjunction with semver because we're just following k8s versions. :-/

mikebrow:
perhaps a new flag to use the old default

mikebrow:
crictl run.. yeah makes sense there as trying to mimic docker run.. but most of the time I think kubelet is trying to follow pull policy in this regard. Some users want to run disconnected for example, others don't want to update their images regardless of if connected.

saschagrunert:
> crictl run.. yeah makes sense there as trying to mimic docker run.. but most of the time I think kubelet is trying to follow pull policy in this regard. Some users want to run disconnected for example, others don't want to update their images regardless of if connected.

Yeah I see the point. We probably should turn it off by default. :+1:

mikebrow:
suggest adding the new default config flag... and pushing that backwards via service. Sorry didn't see this when it was pushed.  So busy so many projects :-)

mikebrow:
crictl isn't being used by kublet for create so this isn't as important as I thought it might be

PR #509, opened 2019-08-21
Use IsAppArmorEnabled of Kubernetes instead of local one

saschagrunert:
The function is now available so we can use this one instead of our own
implementation.


feiskyer:
thanks for the cleanup

PR #510, opened 2019-08-21
Enable deadcode, goimports, misspell and prealloc linters

saschagrunert:
This enables the linters and fixes the corresponding reports.

### DIFF
--- a/pkg/validate/image.go
+++ b/pkg/validate/image.go
@@ -198,7 +198,7 @@ func testPullPublicImage(c internalapi.ImageManagerService, imageName string, po
 
 // pullImageList pulls the images listed in the imageList.
 func pullImageList(c internalapi.ImageManagerService, imageList []string, podConfig *runtimeapi.PodSandboxConfig) []string {
-	var ids []string
+	ids := []string{}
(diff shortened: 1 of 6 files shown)


feiskyer:
Thanks to enable more linters. Could you check why the CIs are failing?

saschagrunert:
> Thanks to enable more linters. Could you check why the CIs are failing?

I'm not sure and it looks like flakes. I did a rebase and will investigate further.

pkg/validate/image.go
```diff
@@ -198,7 +198,7 @@ func testPullPublicImage(c internalapi.ImageManagerService, imageName string, po
 
 // pullImageList pulls the images listed in the imageList.
 func pullImageList(c internalapi.ImageManagerService, imageList []string, podConfig *runtimeapi.PodSandboxConfig) []string {
-	var ids []string
+	ids := []string{}
```
saschagrunert: The issue was that the previous pre-allocation to `len(imageList)` compromised the test result. So this is fixed now. :)

feiskyer: thanks for the fix

PR #512, opened 2019-08-29
Treat first `crictl stats` argument as filter ID

saschagrunert:
The first argument supplied to `crictl stats` can be taken as the filter
ID of the container, whereas `--id` gets precedence.


saschagrunert:
Rebased on top of the latest master branch to get the travis changes in.

Issue #514, opened 2019-08-29
ppc64le tests are failing


Random-Liu:
It blocks all PRs right now. I'll send a PR to disable the test for now.

PR #515, opened 2019-08-29
Disable ppc64le test.

Random-Liu:
For https://github.com/kubernetes-sigs/cri-tools/issues/514.

Let's disable the test for now. @nitkon Can you fix the test? So that we can enable the test again.


nitkon:
@Random-Liu: SInce when are the ppc64le Travis failing? Any specific reason you noticed?

Random-Liu:
@nitkon It is hard to tell. Based on travis cron job, it seems to be 3 days ago.

PR #517, opened 2019-08-30
The check should not have 1s timeout.

Random-Liu:
We should not set the 1 second timeout on the check exec, it may timeout.

This is especially important on windows, because the exec is slower. :)

/cc @yujuhong @feiskyer 


Random-Liu:
Apply lgtm based on https://github.com/kubernetes-sigs/cri-tools/pull/517#pullrequestreview-281839448

Issue #364, opened 2018-08-08
Add testing for crictl


yujuhong:
The crictl command-line tool is used by components such as kubeadm, and is also shipped by default on the Kubernetes nodes in GCE.

We should have testing for it to catch bugs. One option is to add a fake runtime implementation. Any thoughts/suggestions?

(We can also include a crictl test in kubernetes, but the turn-around time for bug fix would be long...)

/cc @Random-Liu

feiskyer:
+1, I think we can use dockershim to test crictl, which is also used for critest

feiskyer:
nop, test functionality of crictl itself, e.g. the expected output of each commands.

xlgao-zju:
> e.g. the expected output of each commands.

I mean maybe we can use the `ginkgo` framework in the test of crictl too?

Random-Liu:
Actually we can use crictl to write CRI validation test, but that requires a lot of code refactoring. :p

yujuhong:
kubectl tests are essentially a pile of shell scripts. I wonder if there are better ways to do this.

yujuhong:
> Actually we can use crictl to write CRI validation test, but that requires a lot of code refactoring. :p

We should still keep the existing CRI validation test, since kubelet doesn't use crictl to talk to CRI shims ;)

feiskyer:
> FWIW, We use crictl directly in CRI-O integration tests.

👍 [bats](https://github.com/bats-core/bats-core) is used in cri-o. Seems it's much better than direct shell scripts.

feiskyer:
@saschagrunert there's no tests for crictl yet, would you like to work on it?

saschagrunert:
> @saschagrunert there's no tests for crictl yet, would you like to work on it?

Sure, do we want to use bats (like in CRI-O), or ginkgo (like in Kubernetes). I tend to go with ginkgo, since this dependency is already available.

feiskyer:
is ginkgo working well for CLI tests?

saschagrunert:
> is ginkgo working well for CLI tests?

Yes, with gomega’s [gexec](https://onsi.github.io/gomega/#gexec-testing-external-processes), I’ll prepare a PR.

hickeyma:
@saschagrunert @feiskyer @yujuhong Just wondering if this is covered by the work in #613?

saschagrunert:
[...]

PR #529, opened 2019-09-05
Add first crictl e2e test framework and suite

saschagrunert:
This adds a new Makefile target `test-e2e`, which uses the existing
ginkgo framework to run end-to-end tests for `crictl`.

The test framework and suite setups CRI-O in a tmp sandbox, whereas
every test can specify if a runtime is needed or not. This way it is
possible to create a new isolated environment where `crictl` can operate
on. Please be aware that external dependencies like `runc` and the CNI
plugins have to be existend on the system to acually work. The tests
have to be executed as root for now.

An example test suite for the `help` as well as the `info` command has
been added as well.

Relates to: https://github.com/kubernetes-sigs/cri-tools/issues/364


saschagrunert:
Now runs seamlessly in travis :)

PR #530, opened 2019-09-10
Add `crictl info` e2e tests

saschagrunert:
I also fixed the cni-config directory to make the network actually work. The `info` tests now validates this.

### DIFF
--- a/test/e2e/info_test.go
+++ b/test/e2e/info_test.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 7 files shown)


test/e2e/info_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+)
+
+// The actual test suite
+var _ = t.Describe("info", func() {
+	It("should succeed", func() {
+		// Given
+		endpoint, testDir, crio := t.StartCrio()
```
feiskyer: Use `BeforeEach` and `AfterEach` for setup/teardown?

saschagrunert: Sure, I changed it as suggested.

feiskyer: thanks

PR #531, opened 2019-09-10
Add GOPROXY and GOSUMDB to speedup modules

saschagrunert:
This should speed up the CI a bit.


Makefile
```diff
@@ -14,6 +14,9 @@
 
 GO ?= go
 
+export GOPROXY=https://proxy.golang.org
+export GOSUMDB=https://sum.golang.org
```
feiskyer: Could we only set this in travis config file?

saschagrunert: Hm, I think we could also speed up local development with this. Adding it to the Makefile and script would keep things consistent between travis and my local working env. WDYT?

saschagrunert: Okay, I add them the travis only.

PR #536, opened 2019-09-19
Avoid empty fields in crictl inspect(p/i) result.

Random-Liu:
We should cherrypick this into 1.16.

Result without the fix:
```console
$ crictl inspectp 6
{
  "status": {
    "id": "6f96eea53f1ead33307865932470eab19b2b12f4f9151b333524321a802b7daa",
    "metadata": {
      "attempt": 1,
      "name": "nginx-sandbox1",
      "namespace": "default1",
      "uid": "hdishd83djaidwnduwk28bcsb11"
    },
    "state": "SANDBOX_READY",
    "createdAt": "2019-09-18T22:15:27.128377789-07:00",
    "network": {
      "additionalIps": [],
      "ip": "10.88.0.22"
    },
    "linux": {
      "namespaces": {
        "options": {
          "ipc": "POD",
          "network": "POD",
          "pid": "POD"
        }
      }
    },
    "labels": {},
    "annotations": {
      "k1": "v1",
      "k2": "v2",
      "test.io.abc": "test-test"
    },
    "runtimeHandler": "",
    "-": null,                 <------------------------ This
    "-": null                  <------------------------ This
  },
  "": "",                       <------------------------ This
  "info": {
    "pid": 37905,
    "processStatus": "running",
    "netNamespaceClosed": false,
    "image": "k8s.gcr.io/pause:3.1",
    "snapshotKey": "6f96eea53f1ead33307865932470eab19b2b12f4f9151b333524321a802b7daa",
    "snapshotter": "overlayfs",
    "runtimeHandler": "",
    "runtimeType": "io.containerd
...
```

### DIFF
--- a/cmd/crictl/util.go
+++ b/cmd/crictl/util.go
@@ -210,7 +210,7 @@ func outputProtobufObjAsYAML(obj proto.Message) error {
 
 func outputStatusInfo(status string, info map[string]string, format string) error {
 	// Sort all keys
-	keys := make([]string, len(info))
+	keys := []string{}
 	for k := range info {
 		keys = append(keys, k)
 	}
@@ -267,7 +267,7 @@ func marshalMapInOrder(m map[string]interface{}, t interface{}) (string, error) 
 	v := reflect.ValueOf(t)
 	for i := 0; i < v.Type().NumField(); i++ {
 		field := jsonFieldFromTag(v.Type().Field(i).Tag)
-		if field == "" {
+		if field == "" || field == "-" {
 			continue
 		}
 		value, err := json.Marshal(m[field])
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,8 @@ jobs:
       name: crictl e2e
       os: linux
       script:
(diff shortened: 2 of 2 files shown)


cmd/crictl/util.go
```diff
@@ -210,7 +210,7 @@ func outputProtobufObjAsYAML(obj proto.Message) error {
 
 func outputStatusInfo(status string, info map[string]string, format string) error {
 	// Sort all keys
-	keys := make([]string, len(info))
+	var keys []string
```
saschagrunert: I think something like this should fix the linter warning:
```suggestion
keys := []string{}
```

feiskyer: yep 

```
cmd/crictl/util.go:213:2: Consider preallocating `keys` (prealloc)
	var keys []string
	^
Makefile:92: recipe for target 'lint' failed
make: *** [lint] Error 1
The command "make lint" exited with 2.
```

Random-Liu: Done

PR #538, opened 2019-09-19
Skip running portforward on host network test for non-linux

liyanhui1228:


Random-Liu:
/lgtm

Windows container doesn't support host network, we should not run host network test on windows.

liyanhui1228:
Looks like the test failures are not related to this change, can we skip it?

Random-Liu:
/cc @saschagrunert crictl test is failing.

@liyanhui1228 Let's comment out the test from .travis.

Random-Liu:
@liyanhui1228 My change is merged. Rebase your change, and the test should be skipped.

liyanhui1228:
Needs lgtm label reapplied, thanks!

Issue #539, opened 2019-09-19
crictl e2e test is failing.


Random-Liu:
It is currently blocking PRs.
```
Summarizing 8 Failures:
[Fail] crictl: pull [It] should succeed without tag or digest 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:104
[Fail] crictl: pull [It] should succeed with tag 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:104
[Fail] crictl: pull [It] should fail on not existing image 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:126
[Fail] crictl: info [It] should succeed 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:104
[Fail] crictl: info [AfterEach] should fail with additional argument 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:205
[Fail] crictl: pull [It] should succeed with digest 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:104
[Fail] crictl: pull [AfterEach] should succeed to show the help 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:205
[Fail] crictl: version [It] should succeed 
/home/travis/gopath/src/github.com/kubernetes-sigs/cri-tools/test/framework/framework.go:104
Ran 12 of 12 Specs in 246.295 seconds
FAIL! -- 4 Passed | 8 Failed | 0 Pending | 0 Skipped
Ginkgo ran 1 suite in 4m18.491441555s
Test Suite Failed
Makefile:108: recipe for target 'test-e2e' failed
make: *** [test-e2e] Error 1
```

saschagrunert:
The issue is that we moved conmon out of the repository. I will fix this and propose a PR.

PR #540, opened 2019-09-20
Fix e2e tests by cloning conmon

saschagrunert:
Closes https://github.com/kubernetes-sigs/cri-tools/issues/539 by adding an additional conmon clone step.

### DIFF
--- a/test/framework/framework.go
+++ b/test/framework/framework.go
@@ -128,18 +128,25 @@ func (t *TestFramework) CrictlExpectFailureWithEndpoint(
 
 func SetupCrio() string {
 	const (
-		crioURL = "https://github.com/cri-o/cri-o"
-		timeout = 10 * time.Minute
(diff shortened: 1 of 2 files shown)


test/framework/framework.go
```diff
@@ -128,18 +128,25 @@ func (t *TestFramework) CrictlExpectFailureWithEndpoint(
 
 func SetupCrio() string {
```
Random-Liu: Instead of building cri-o, can we use released binaries? The test should not be testing newest cri-o because it is not cri-o CI.

feiskyer: agreed. @saschagrunert could you download a stable release of cri-o in the test?

saschagrunert: Yes, I’ll do when I’m back in my office in the middle of the next week. :)

saschagrunert:
Closing in favor of https://github.com/kubernetes-sigs/cri-tools/pull/545

PR #541, opened 2019-09-24
Pass pod config when pulling image during create

kevpar:
Signed-off-by: Kevin Parsons <kevpar@microsoft.com>

The change to pull image as part of container create broke LCOW (Linux Containers on Windows), as it depends on knowing the pod config at pull time to select the correct snapshotter. We already have the pull config available in this function, so this change just passes it through to the pull operation.

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -574,7 +574,7 @@ func CreateContainer(
 
 		// Try to pull the image before container creation
 		image := config.GetImage().GetImage()
-		if _, err := PullImage(iClient, image, auth); err != nil {
+		if _, err := PullImageWithSandbox(iClient, image, auth, podConfig); err != nil {
 			return "", err
 		}
 	}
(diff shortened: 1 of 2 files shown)


jterry75:
@Random-Liu - Why is this code ever here? We do not allow CRI to pull the image at `ContainerCreate` time in the daemon why do we do it client side. I don't see really why this expectation is an improvement over just issuing the set of `crictl pull` requests that you need and then separately the `crictl create` requests. Seems like business logic in `crictl` vs just a cmdline invocation around CRI API. Thoughts?

feiskyer:
> @Random-Liu - Why is this code ever here? We do not allow CRI to pull the image at ContainerCreate time in the daemon why do we do it client side. I don't see really why this expectation is an improvement over just issuing the set of crictl pull requests that you need and then separately the crictl create requests. Seems like business logic in crictl vs just a cmdline invocation around CRI API. Thoughts?

This is client side change only for crictl. In Kubelet, the image would be pulled before creating the container; while in crictl, image is also pulled to make crictl easier to use.

jterry75:
@feiskyer - Thanks!

cgwalters:
Could this have caused https://github.com/kubernetes-sigs/cri-tools/issues/548 ?

jterry75:
@cgwalters - Seems very likely yes.

PR #544, opened 2019-10-01
fixed Tim St. Clair's username

geekygirldawn:
His username was incorrect in this OWNERS file.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -9,7 +9,7 @@ approvers:
   - heartlock
   - mrunalp
   - saschagrunert
-  - timstclair
+  - tallclair
   - xlgao-zju
   - yujuhong
   - runcom


geekygirldawn:
/cc @mrbobbytables 
Since I know you're also doing some OWNERS file cleanup :)

feiskyer:
seems these are two different people from different companies. @timstclair @timothysc WDYT?

OWNERS
```diff
@@ -9,7 +9,7 @@ approvers:
   - heartlock
   - mrunalp
   - saschagrunert
-  - timstclair
```
mrbobbytables: I think this is meant to be @tallclair - https://github.com/timstclair has a pointer to the newer GH handle.

geekygirldawn: Ah, good point ... the other Time St. Clair :) I've updated the commit with the correct GitHub handle.

geekygirldawn:
OK, I've updated this with the username for the correct Tim St. Clair :)

feiskyer:
@mrbobbytables @geekygirldawn thanks for the update
/lgtm
/approve

tallclair:
Haha, thanks for this.

PR #552, opened 2019-11-07
  Download golangci-lint via wget 

haircommander:
Kubernetes now requires golang 1.13.4. When using golang 1.13.4 to build critest, we run into golangci-lint#659. Fix this with a couple of changes:

- install the golangci-lint binary for 1.17.1 with curl (similar to how CRI-O currently does), instead of using go mods
    - this includes dropping golangci-lint on ppc (which shouldn't matter)
    - which also includes breaking `make install.tools` into two different targets: `lint` and `ginkgo`
- bump the go version used by travis to 1.13 (because kubelet now requires 1.13.4 to build)
- revendor cri-tools with golang 1.13 (using the command: `podman run --privileged --rm --env HOME=/root -v $PWD:/src -w /src golang:1.13 make vendor`)


haircommander:
@saschagrunert @mrunalp @Random-Liu PTAL

saschagrunert:
Ah, interesting. Can we remove it from vendoring at all and just download the binary via the curl script, like we do in CRI-O? I recently hit compilation issues with golangci-lint and they recommend to use their prebuilt ones.

haircommander:
@saschagrunert good idea, fixed!

Makefile
```diff
@@ -95,7 +95,12 @@ lint: $(GOLANGCI_LINT)
 install.tools: $(GOLANGCI_LINT) $(GINKGO)
 
 $(GOLANGCI_LINT):
-	$(call go-build,./vendor/github.com/golangci/golangci-lint/cmd/golangci-lint)
+	export \
+		VERSION=v1.17.1 \
```
saschagrunert: Let’s use 1.21.0 here. :)

haircommander: fixed!

saschagrunert:
@haircommander we have to exclude the golangci lint check for travis on ppc, because it is not supported right now. :-/

saschagrunert:
@mrunalp @feiskyer @Random-Liu PTAL

.golangci.yml
```diff
@@ -2,6 +2,7 @@
 run:
   concurrency: 6
   deadline: 5m
+  modules-download-mode: vendor
```
saschagrunert: This should not be needed, do we have to re-vendor with go 1.13 to fix this up?

haircommander: I am not sure the issue. I can try that.

haircommander: ah nice that worked

haircommander:
huzzah, (finally) green tests. @mrunalp @feiskyer @Random-Liu PTAL

PR #580, opened 2020-03-30
Update CI go version to v1.14.x and update travis deps

saschagrunert:
We now update runc and CNI plugins in travis as well as switching to the
latest go 1.14.x.


feiskyer:
Kubernetes  is still using go1.13.9, let's align Go version with it?

saschagrunert:
Yeah sounds good let's align with them.

PR #583, opened 2020-04-01
Remove pod sandoxes in parallel

saschagrunert:
We now use a classic work-stealing approach to remove the sandboxes in
parallel if multiple are defined.

/cc @ereslibre


saschagrunert:
> what about using [AggregateGoroutines](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go#L234)?

Thanks for the hint, used the suggested API in the implementation :+1:

PR #590, opened 2020-04-21
Update SELinux testing for MCS

erikwilson:
Check for MCS categories, different labels for different sandboxes,
and cleanup all sandboxes.

### DIFF
--- a/pkg/validate/selinux_linux.go
+++ b/pkg/validate/selinux_linux.go
@@ -18,10 +18,11 @@ package validate
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/kubernetes-sigs/cri-tools/pkg/framework"
-	"github.com/opencontainers/selinux/go-selinux"
+	selinux "github.com/opencontainers/selinux/go-selinux"
 	internalapi "k8s.io/cri-api/pkg/apis"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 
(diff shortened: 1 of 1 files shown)


erikwilson:
/assign @saschagrunert

Apologies for not filing an issue first, this PR will check that MCS labels are created, that container labels match in the same pod, and are different in multiple/unique unprivileged pods.

pkg/validate/selinux_linux.go
```diff
@@ -35,122 +36,144 @@ var _ = framework.KubeDescribe("SELinux", func() {
...
+				label2 := checkProcessLabelMCS(rc, containerID2, false)
+				Expect(label1).NotTo(Equal(label2))
 			})
 		})
-	}
+	})
 })
 
-func createContainerWithSelinux(rc internalapi.RuntimeService, ic internalapi.ImageManagerService, sandboxID string, sandboxConfig *runtimeapi.PodSandboxConfig, options *runtimeapi.SELinuxOption, shouldStart, shouldCreate bool) string {
+func SELIt(description string, doTest func()) {
```
feiskyer, line 157: What's the reason of moving this check required for every test case?

pkg/validate/selinux_linux.go
```diff
@@ -187,28 +210,65 @@ func checkContainerSelinux(rc internalapi.RuntimeService, containerID string, sh
...
+func checkProcessLabelMCS(rc internalapi.RuntimeService, containerID string, privileged bool) string {
+	// Check that the correct process label mcs is set
+	cmd := []string{"cat", "/proc/self/attr/current"}
+	stdout, stderr, err := rc.ExecSync(containerID, cmd, time.Duration(defaultExecSyncTimeout)*time.Second)
+	label := strings.Trim(string(stdout), "\x00")
+	msg := fmt.Sprintf("cmd %v, stdout %q, stderr %q", cmd, stdout, stderr)
+	Expect(err).NotTo(HaveOccurred(), msg)
 	if privileged {
-		Expect(string(stdout)).To(ContainSubstring("system_r:spc_t"))
+		Expect(label).To(MatchRegexp(`:s0(-s0)?(:c0\.c1023)?$`))
```
feiskyer, line 269: would you add some explanation for these two regexp?

pkg/validate/selinux_linux.go
```diff
@@ -187,28 +210,65 @@ func checkContainerSelinux(rc internalapi.RuntimeService, containerID string, sh
...
+	}
+}
 
+func checkMountLabelMCS(rc internalapi.RuntimeService, containerID string) {
+	// Check that the mount label mcs is set
+	cmd := []string{"cat", "/proc/1/mountinfo"}
+	stdout, stderr, err := rc.ExecSync(containerID, cmd, time.Duration(defaultExecSyncTimeout)*time.Second)
+	msg := fmt.Sprintf("cmd %v, stdout %q, stderr %q", cmd, stdout, stderr)
+	Expect(err).NotTo(HaveOccurred(), msg)
+	Expect(string(stdout)).To(MatchRegexp(`,context="[^"]*:s0(-s0)?:c[0-9]+,c[0-9]+",`))
```
feiskyer, line 258: same here

feiskyer:
@erikwilson any updates on the PR?

pkg/validate/selinux_linux.go
erikwilson: The main reason is because if selinux is not available the tests will not be registered, even as skipped. 

I do not think the current solution is ideal, I can revert it if desired, but it would be nice to see the tests are skipped rather than ignored and have a way to run them on non-selinux systems.

erikwilson:
Sorry for the slow response, thank you for the review @feiskyer. Updated with comments. I think we found another test which should be performed, which should check mount labels for image volumes, related to this change: https://github.com/rancher/cri/commit/8000d0528e25f12d6ab7d722560cb72b2f9a2198

Please let me know if I can clarify the comments or behavior regarding MCS testing.

pkg/validate/selinux_linux.go
erikwilson: Reverted this change to the old behavior to help move the PR through.

erikwilson:
Thanks for the feedback @feiskyer, updated the PR per comments, I think this should be good to go as is and we can add more tests in future PRs if helpful.

PR #594, opened 2020-04-27
Enable critest to use CRI server configuration file

### LINKED ISSUE #592: Enable critest to use CRI server configuration file similar to crictl
`crictl` tool allows CRI server configuration to be set in the following ways:
- CLI flags like `--runtime-endpoint` and `image-endpoint`
- Config file as specified by CLI flag (`--config`), in default system location, or in the binary path
- default values

`critest` uses:
- CLI flags like `--runtime-endpoint` and `image-endpoint`
- default values

I think `critest` CRI server configuration should work similar to `crictl`. For example, if user configures `crictl` to use `containerd` in a config file, then `critest` should work the same without having to pass the `containerd` runtime flag to it.

hickeyma:
This commit refactors `crictl` config logic so that `critest` can also use config file for CRI server configuration.

Closes #592

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -19,17 +19,20 @@ package main
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"sort"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
+
 	"google.golang.org/grpc"
+
 	internalapi "k8s.io/cri-api/pkg/apis"
 	"k8s.io/kubernetes/pkg/kubelet/remote"
 	"k8s.io/kubernetes/pkg/kubelet/util"
 
+	"github.com/kubernetes-sigs/cri-tools/pkg/common"
 	"github.com/kubernetes-sigs/cri-tools/pkg/version"
 )
 
@@ -60,7 +63,7 @@ func getRuntimeClientConnection(context *cli.Context) (*grpc.ClientConn, error) 
 
 	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithBlock(), grpc.WithTimeout(Timeout), grpc.WithContextDialer(dialer))
 	if err != nil {
-		return nil, fmt.Errorf("failed to connect, make sure you are running as root and the runtime has been started: %v", err)
+		return nil, errors.Wrap(err, "connect, make sure you are running as root and the runtime has been started")
 	}
 	return conn, nil
 }
@@ -80,7 +83,7 @@ func getImageClientConnection(context *cli.Context) (*grpc.ClientConn, error) {
 
 	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithBlock(), grpc.WithTimeout(Timeout), grpc.WithContextDialer(dialer))
 	if err != nil {
-		return nil, fmt.Errorf("failed to connect, make sure you are running as root and the runtime has been started: %v", err)
+		return nil, errors.Wrap(err, "connect, make sure you are running as root and the runtime has been started")
 	}
 	return conn, nil
 }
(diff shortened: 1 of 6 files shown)


cmd/crictl/main.go
```diff
@@ -161,42 +160,21 @@ func main() {
...
-					isUseConfig = true
-				} else if !os.IsNotExist(err) {
-					logrus.Fatalf("Failed to load config file: %v", err)
-				}
+		var config *common.ServerConfiguration
+		var err error
+
+		if config, err = common.GetServerConfigFromFile(context.String("config"), os.Args[0]); err != nil {
+			if context.IsSet("config") {
+				logrus.Fatalf(err.Error())
```
saschagrunert, line 168:
```suggestion
				logrus.Fatal(err)
```

cmd/crictl/main.go
```diff
@@ -161,42 +160,21 @@ func main() {
 	}
 
 	app.Before = func(context *cli.Context) error {
```
saschagrunert, line 162:
```suggestion
	app.Before = func(context *cli.Context) (err error) {
```

cmd/crictl/main.go
```diff
@@ -161,42 +160,21 @@ func main() {
...
-				configFile = filepath.Join(filepath.Dir(os.Args[0]), "crictl.yaml")
-				if _, err := os.Stat(configFile); err == nil {
-					isUseConfig = true
-				} else if !os.IsNotExist(err) {
-					logrus.Fatalf("Failed to load config file: %v", err)
-				}
+		var config *common.ServerConfiguration
+		var err error
+
+		if config, err = common.GetServerConfigFromFile(context.String("config"), os.Args[0]); err != nil {
```
saschagrunert, line 166:
```suggestion
		if config, err = common.GetServerConfigFromFile(context.String("config"), os.Executable()); err != nil {
```

pkg/common/config.go
```diff
@@ -0,0 +1,69 @@
...
+	// Debug enable debug output
+	Debug bool
+}
+
+// GetServerConfigFromFile returns the CRI server configuration from file
+func GetServerConfigFromFile(configFileName, currentDir string) (*ServerConfiguration, error) {
+	serverConfig := ServerConfiguration{}
+	if _, err := os.Stat(configFileName); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, fmt.Errorf("Failed to load config file: %v", err)
```
saschagrunert, line 43: Let's use the pkg/errors package here and in other locations as well:
```suggestion
			return nil, errors.Wrap(err, "load config file")
```

pkg/common/file.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// WriteConfig writes config to file
+// an error if the file was unable to be written to.
+func WriteConfig(c *Config, filepath string) error {
+	data, err := yaml.Marshal(c)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(gofilepath.Dir(filepath), 0755); err != nil {
```
saschagrunert, line 58: I think we already build with go 1.13 only, so:
```suggestion
	if err := os.MkdirAll(gofilepath.Dir(filepath), 0o755); err != nil {
```

cmd/crictl/main.go
hickeyma: Done

cmd/crictl/main.go
hickeyma: Done

cmd/crictl/main.go
hickeyma: Updated

pkg/common/config.go
hickeyma: Sure. I updated it for the files changed in this PR. I will open a new PR to cover the rest of the files in the project if thats ok? I want to avoid changing files not directly related to the fix.

pkg/common/file.go
hickeyma: Done

hickeyma:
Thanks for the review @saschagrunert. Updated and ready for review again.

hickeyma:
Thanks @feiskyer @saschagrunert for the reviews and merge.

PR #595, opened 2020-04-29
Use errors.wrap instead of fmt.Errorf

hickeyma:
When displaying an error returned from a function call, it is better to use the errors.wrap function to display it with customized message. Its error handling has more advantages.

This PR is as a result of a suggestion from @saschagrunert in https://github.com/kubernetes-sigs/cri-tools/pull/594#discussion_r416364459

### DIFF
--- a/cmd/crictl/attach.go
+++ b/cmd/crictl/attach.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/url"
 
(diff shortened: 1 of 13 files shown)


hickeyma:
Thanks for the reviews @saschagrunert @feiskyer and merge @feiskyer

PR #598, opened 2020-04-30
Update the default CRI server defaults

### LINKED ISSUE #597: Should there be a default for the CRI server endpoint?
Currently, the CRI server endpoint default is `dockershim.sock`. I am wondering if this makes sense as it could be `dockershim.sock` or `containerd.sock` or `cri-o.sock`?

Would it be better that the user specifies explicitly the endpoint? This way it avoids connecting unexpectedly to a running server.

hickeyma:
Currently, the default CRI server is dockershim. This is a deprecated server with cri-tools. Also, does it make sense to even have a default as the user should explicitly define it.

This PR adds `containerd` and `cri-o` to the default list and it checks each in order till one works. It also adds some looging that the defaults are being used. This at least provdes better usability to the user if they don't configure the server.

Closes #597

### DIFF
--- a/cmd/crictl/attach.go
+++ b/cmd/crictl/attach.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/url"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/net/context"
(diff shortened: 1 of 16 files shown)


hickeyma:
Made a mistake with my branches. Closing in favour of #599

PR #599, opened 2020-04-30
Update the default CRI server

### LINKED ISSUE #597: Should there be a default for the CRI server endpoint?

hickeyma:
Currently, the default CRI server is dockershim. This is a deprecated server with cri-tools. Also, does it make sense to even have a default as the user should explicitly define it.

This PR adds `containerd` and `cri-o` to the default list and it checks each in order till one works. It also adds some logging that the defaults are being used. This at least provdes better usability to the user if they don't configure the server.

Closes #597

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"os"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -60,9 +59,15 @@ func getRuntimeClientConnection(context *cli.Context) (*grpc.ClientConn, error) 
 	if RuntimeEndpointIsSet && RuntimeEndpoint == "" {
 		return nil, fmt.Errorf("--runtime-endpoint is not set")
 	}
-	logrus.Debugf("get runtime connection")
+	logrus.Debug("get runtime connection")
 	// If no EP set then use the default endpoint types
 	if !RuntimeEndpointIsSet {
+		logrus.Warningf("runtime connect using default endpoints: %v. "+
+			"As the default settings are now deprecated, you should set the "+
+			"endpoint instead.", defaultRuntimeEndpoints)
+		logrus.Debug("Note that performance maybe affected as each default " +
+			"connection attempt takes n-seconds to complete before timing out " +
+			"and going to the next in sequence.")
 		return getConnection(defaultRuntimeEndpoints)
 	}
 	return getConnection([]string{RuntimeEndpoint})
@@ -79,6 +84,12 @@ func getImageClientConnection(context *cli.Context) (*grpc.ClientConn, error) {
 	logrus.Debugf("get image connection")
 	// If no EP set then use the default endpoint types
 	if !ImageEndpointIsSet {
+		logrus.Warningf("image connect using default endpoints: %v. "+
+			"As the default settings are now deprecated, you should set the "+
+			"endpoint instead.", defaultRuntimeEndpoints)
+		logrus.Debug("Note that performance maybe affected as each default " +
+			"connection attempt takes n-seconds to complete before timing out " +
+			"and going to the next in sequence.")
 		return getConnection(defaultRuntimeEndpoints)
 	}
 	return getConnection([]string{ImageEndpoint})
@@ -89,21 +100,9 @@ func getConnection(endPoints []string) (*grpc.ClientConn, error) {
 		return nil, fmt.Errorf("endpoint is not set")
 	}
 	endPointsLen := len(endPoints)
-	if endPointsLen > 1 {
-		logrus.Warningf("using default endpoints: %s", endPoints)
-	}
-
 	var conn *grpc.ClientConn
 	for indx, endPoint := range endPoints {
-		if endPointsLen > 1 {
-			if strings.Contains(endPoint, "dockershim") {
-				logrus.Warningf("connect using default endpoint: %s. Note: Dockershim endpoint is now deprecated.", endPoint)
-			} else {
-				logrus.Warningf("connect using default endpoint: %s", endPoint)
-			}
-		} else {
-			logrus.Debugf("connect using endpoint: %s", endPoint)
-		}
+		logrus.Debugf("connect using endpoint: %s", endPoint)
 		addr, dialer, err := util.GetAddressAndDialer(endPoint)
 		if err != nil {
 			if indx == endPointsLen-1 {
@@ -112,7 +111,6 @@ func getConnection(endPoints []string) (*grpc.ClientConn, error) {
 			logrus.Error(err)
 			continue
 		}
-
 		conn, err = grpc.Dial(addr, grpc.WithInsecure(), grpc.WithBlock(), grpc.WithTimeout(Timeout), grpc.WithContextDialer(dialer))
 		if err != nil {
 			errMsg := errors.Wrapf(err, "connect endpoint '%s', make sure you are running as root and the endpoint has been started", endPoint)
@@ -121,6 +119,7 @@ func getConnection(endPoints []string) (*grpc.ClientConn, error) {
 			}
 			logrus.Error(errMsg)
 		} else {
+			logrus.Debugf("connected successfully using endpoint: %s", endPoint)
 			break
 		}
 	}
@@ -167,6 +166,11 @@ func main() {
 		completionCommand,
 	}
 
+	runtimeEndpointUsage := fmt.Sprintf("Endpoint of CRI container runtime "+
+		"service (default: uses in order the first successful one of %v). "+
+		"Default is now deprecated and the endpoint should be set instead.",
+		defaultRuntimeEndpoints)
+
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{
 			Name:    "config",
@@ -179,13 +183,14 @@ func main() {
 			Name:    "runtime-endpoint",
 			Aliases: []string{"r"},
 			EnvVars: []string{"CONTAINER_RUNTIME_ENDPOINT"},
-			Usage:   "Endpoint of CRI container runtime service",
+			Usage:   runtimeEndpointUsage,
 		},
 		&cli.StringFlag{
 			Name:    "image-endpoint",
 			Aliases: []string{"i"},
 			EnvVars: []string{"IMAGE_SERVICE_ENDPOINT"},
-			Usage:   "Endpoint of CRI image manager service",
+			Usage: "Endpoint of CRI image manager service (default: uses " +
+				"'runtime-endpoint' setting)",
 		},
 		&cli.DurationFlag{
 			Name:    "timeout",
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


cmd/crictl/main_unix.go
```diff
@@ -20,5 +20,5 @@ package main
 
 const (
 	defaultConfigPath      = "/etc/crictl.yaml"
-	defaultRuntimeEndpoint = "unix:///var/run/dockershim.sock"
+	defaultRuntimeEndpoint = "unix:///var/run/dockershim.sock unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock"
```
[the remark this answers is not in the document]
hickeyma: Yeah, was thinking the same thing but decided to go with the string instead.

I agree that it would be safer and more robust code to have a separate array with the different defaults.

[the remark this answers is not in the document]
hickeyma: Updated

cmd/crictl/main_windows.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	defaultRuntimeEndpoint = "npipe:////./pipe/dockershim"
+	defaultRuntimeEndpoint = "npipe:////./pipe/dockershim npipe:////./pipe/containerd npipe:////./pipe/crio"
```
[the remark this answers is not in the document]
hickeyma: Updated

hickeyma:
Thanks for the review and feedback @mikebrow. Updated and ready for review again.

cmd/crictl/main.go
```diff
@@ -55,17 +56,14 @@ func getRuntimeClientConnection(context *cli.Context) (*grpc.ClientConn, error)
...
-	}
-
-	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithBlock(), grpc.WithTimeout(Timeout), grpc.WithContextDialer(dialer))
-	if err != nil {
-		return nil, errors.Wrap(err, "connect, make sure you are running as root and the runtime has been started")
+	logrus.Debugf("get runtime connection")
+	// As dockershim is deprecated as a default CRI server, it still needs to
+	// be supported. This allows other default endpoint types to be
+	// checked if cannot connect to dockershim
+	if strings.Contains(RuntimeEndpoint, "dockershim") {
```
mikebrow, line 63: They might not use the exact same path to `[0]` "unix:///var/run/dockershim.sock".. I think you need to try to connect to exactly what they set at the cli or config, if so configured. Suggestion: set a flag to skip trying the defaults if configured..  then here if !configured you can go with the defaults?

feiskyer:
> Currently, the default CRI server is dockershim. This is a deprecated server with cri-tools. Also, does it make sense to even have a default as the user should explicitly define it.

I prefer we still use dockershim as the default CRI server because most Kubernetes users are still using dockershim. To switch to a different runtime, users could provide a config file for crictl.

> This PR adds containerd and cri-o to the default list and it checks each in order till one works.

This would introduce a few seconds latency for containerd and crio.  @hickeyma @mikebrow do you think this is ok for users?

I have another idea. @hickeyma @mikebrow what do you think if we add a sub-command to generate the config file based on runtime type?

mikebrow:
> > Currently, the default CRI server is dockershim. This is a deprecated server with cri-tools. Also, does it make sense to even have a default as the user should explicitly define it.
> 
> I prefer we still use dockershim as the default CRI server because most Kubernetes users are still using dockershim. To switch to a different runtime, users could provide a config file for crictl.
> 
> > This PR adds containerd and cri-o to the default list and it checks each in order till one works.
> 
> This would introduce a few seconds latency for containerd and crio. @hickeyma @mikebrow do you think this is ok for users?
> 

I don't think the intention is to change the default, or introducing any additional latency for the default or configured options or cli options. He's adding loud warning messages when there is no config or cli and the default fails but there exists a running sock from another container runtime besides the default docker.  

> I have another idea. @hickeyma @mikebrow what do you think if we add a sub-command to generate the config file based on runtime type?

There's a config sub-command that's supposed to have get and set options ala git config pattern... the get works the set isn't there.   ```  crictl config - Get and set crictl options```

cmd/crictl/main.go
feiskyer: +1 as well

feiskyer:
@mikebrow thanks for the clarification.

cmd/crictl/main.go
feiskyer: @hickeyma could you address the comments here?

hickeyma: @mikebrow @feiskyer I have added a bool to identify if endpoints set by cli flag or config file. I used the original check for "dockershim" as I made the assumption that this would not be set and therefore using the default from https://github.com/kubernetes-sigs/cri-tools/blob/master/cmd/crictl/main.go#L190. This may not be the case as the user may set iits as the endpoint also.

hickeyma: Modified.

hickeyma:
Thanks for the reviews @feiskyer @mikebrow. Updated and ready for review again.

cmd/crictl/main_unix.go
```diff
@@ -22,3 +22,5 @@ const (
 	defaultConfigPath      = "/etc/crictl.yaml"
 	defaultRuntimeEndpoint = "unix:///var/run/dockershim.sock"
 )
+
+var defaultRuntimeEndpoints = []string{"unix:///var/run/dockershim.sock", "unix:///run/containerd/containerd.sock", "unix:///run/crio/crio.sock"}
```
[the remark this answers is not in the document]
mikebrow: looks like the string is set by default in the cli? https://github.com/kubernetes-sigs/cri-tools/pull/599/files#diff-20e8c63cd3f9c0ee9151d8e93c4ff9faR186

cmd/crictl/main.go
```diff
@@ -55,17 +60,14 @@ func getRuntimeClientConnection(context *cli.Context) (*grpc.ClientConn, error)
...
-	}
-
-	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithBlock(), grpc.WithTimeout(Timeout), grpc.WithContextDialer(dialer))
-	if err != nil {
-		return nil, errors.Wrap(err, "connect, make sure you are running as root and the runtime has been started")
+	logrus.Debugf("get runtime connection")
+	// As dockershim is deprecated as a default CRI server, it still needs to
+	// be supported. This allows other default endpoint types to be
+	// checked if cannot connect to dockershim
+	if !RuntimeEndpointIsSet {
```
mikebrow, line 67: won't get here see line 60 above.

mikebrow: maybe use =="" as the bool

cmd/crictl/main_unix.go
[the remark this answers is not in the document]
hickeyma: It needs it because it sets `defaultRuntimeEndpoint` if EP is not set by CLI or config file: https://github.com/kubernetes-sigs/cri-tools/blob/master/cmd/crictl/main.go#L190.  Default CLI value: https://github.com/kubernetes-sigs/cri-tools/blob/master/cmd/crictl/main.go#L143

cmd/crictl/main.go
hickeyma: It will get there because it uses a default value if not set:  https://github.com/kubernetes-sigs/cri-tools/blob/master/cmd/crictl/main.go#L190. The check for empty in line 60 is incase empty value set by cli or file.

hickeyma: `if !RuntimeEndpointIsSet` should be a good check for bool false?

cmd/crictl/main_unix.go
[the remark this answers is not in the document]
mikebrow: `defaultRuntimeEndpoints[0]==defaultRuntimeEndpoint`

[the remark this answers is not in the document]
mikebrow: :) if you are going to check `if defaultRuntimeEndpoints[0] || defaultRuntimeEndpoint` then you don't need to set the cli value to a default.. you don't need to do that because your default is the array.

cmd/crictl/main.go
mikebrow: if they set "" are they looking for an error? If it's not "" how is it not set? I'm confused ;-)

hickeyma: yes, logic is a bit funny here now with the changes. I have refactored to hopefully improve it.

hickeyma:
@mikebrow I did a refactor to tidy up the logic around default eps. Let me know what you think.

docs/crictl.md
```diff
@@ -75,6 +75,16 @@ The endpoint can be set in three ways:
 - By setting environment variables `CONTAINER_RUNTIME_ENDPOINT` and `IMAGE_SERVICE_ENDPOINT`
 - By setting the endpoint in the config file `--config=/etc/crictl.yaml`
 
+If the endpoint is not set then it works as follows :
```
mikebrow, line 78: suggest `If the runtime endpoint is not set crictl will, by default, try to connect using`

docs/crictl.md
```diff
@@ -75,6 +75,16 @@ The endpoint can be set in three ways:
...
+
+- Runtime endpoint uses in order the first successful one of the following defaults:
+  - dockershim
+  - containerd
+  - cri-o
+- Image endpoint uses the runtime endpoint setting
+
+> Note: The default endpoints are now deprecated and the runtime endpoint should be set instead.
```
mikebrow, line 86: mention the performance issue with trying and failing

cmd/crictl/main.go
```diff
@@ -140,14 +172,17 @@ func main() {
 			Name:    "runtime-endpoint",
 			Aliases: []string{"r"},
 			EnvVars: []string{"CONTAINER_RUNTIME_ENDPOINT"},
-			Value:   defaultRuntimeEndpoint,
-			Usage:   "Endpoint of CRI container runtime service",
+			Usage: "Endpoint of CRI container runtime service (default: " +
+				"uses in order first successful one of \"dockershim, " +
+				"containerd, cri-o\"). Default is now deprecated and the " +
+				"endpoint should be set instead.",
```
mikebrow, line 178: not sure if I like the logical list here or the actual list...
{"unix:///var/run/dockershim.sock", "unix:///run/containerd/containerd.sock", "unix:///run/crio/crio.sock"}

hickeyma: does this also run on Windows and is this the same CLI output? The defaults were not shown before in the usage.

docs/crictl.md
hickeyma: Sounds better. Updated

docs/crictl.md
hickeyma: Sure, added.

cmd/crictl/main.go
hickeyma: ok, updated now.

hickeyma:
Thanks for review @mikebrow. Updated and ready for review again.

hickeyma:
Thanks for the reviews and feedback @mikebrow

hickeyma:
@feiskyer This is updated and ready for review again.

mikebrow:
seems github's link to travis state got out of whack again..

feiskyer:
yep, let me merge this manually

hickeyma:
Thanks @feiskyer and @mikebrow for sorting this and for the merger @feiskyer

PR #604, opened 2020-05-11
Update set configuration command

### LINKED ISSUE #602: Missing config set option
There's a config sub-command that's supposed to have get and set options ala git config pattern... the get works the set isn't there. 

```crictl config - Get and set crictl options```

### PR DESCRIPTION
**Pull request by hickeyma**
The config command currently implicitly sets a config option when called with `option value`. The get is called with `--get` flag. Also, the command doesn't allow more than one option to be set.
This PR updates it to now include the `--set` flag and enables it to set more than one options as per `opt1=val1,opt2=val2`.

Closes #602 

Note to reviewers:

You can at the moment set a config option as follows:

```console
$ crictl config --help
NAME:
   crictl config - Get and set crictl options

USAGE:
   crictl config [command options] [<options>]

OPTIONS:
   --get value  get value: name
   --help, -h   show help (default: false)

$ cat /etc/crictl.yaml
runtime-endpoint: unix:///run/containerd/containerd.sock
image-endpoint: ""
timeout: 0
debug: false

$ crictl config debug true

$ crictl config image-endpoint unix:///run/containerd/containerd.sock

$ cat /etc/crictl.yaml
runtime-endpoint: unix:///run/containerd/containerd.sock
image-endpoint: unix:///run/containerd/containerd.sock
timeout: 0
debug: true
```

This PR updates the `config` command to add usability improvements including adding the `--set` flag to be more explicit for setting options:

```console
$ crictl config --help
NAME:
   crictl config - Get and set crictl options.

The options are as follows:

- runtime-endpoint: Container runtime endpoint
- image-endpoint: Image endpoint
- timeout: Timeout of connecting to server
- debug: Enable debug output


USAGE:
   crictl config [command options] [<options>]

OPTIONS:
   --get value  show the option value
   --set value  set option (can specify multiple or separate values with commas: opt1=val1,opt2=val2)
   --help, -h   show help (default: false)

$ crictl config --set image-endpoint=unix:///run/containerd/containerd.sock,debug=true

$ crictl config --set image-endpoint=unix:///run/containerd/containerd.sock
```

The PR maintains backwards compatibility

### DIFF
--- a/cmd/crictl/config.go
+++ b/cmd/crictl/config.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -36,7 +37,11 @@ var configCommand = &cli.Command{
 	Flags: []cli.Flag{
 		&cli.StringFlag{
 			Name:  "get",
-			Usage: "get value: name",
+			Usage: "show the option value",
+		},
+		&cli.StringSliceFlag{
+			Name:  "set",
+			Usage: "set option (can specify multiple or separate values with commas: opt1=val1,opt2=val2)",
 		},
 	},
 	Action: func(context *cli.Context) error {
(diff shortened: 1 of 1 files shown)


cmd/crictl/config.go
```diff
@@ -66,37 +71,46 @@ var configCommand = &cli.Command{
...
-		switch key {
-		case "runtime-endpoint":
-			config.RuntimeEndpoint = value
-		case "image-endpoint":
-			config.ImageEndpoint = value
-		case "timeout":
-			n, err := strconv.Atoi(value)
-			if err != nil {
-				logrus.Fatal(err)
+		} else if context.IsSet("set") {
```
mikebrow, line 74: I like the new --set sub command for config and the ability to set multiple end points via separator.

Suggestion:  make the --set sub command also work by default... so command (--set optional)... and add the separator option to the get sub command.

cmd/crictl/config.go
```diff
@@ -66,37 +71,46 @@ var configCommand = &cli.Command{
...
+				options := strings.Split(setting, ",")
+				for _, option := range options {
+					pair := strings.Split(option, "=")
+					if len(pair) != 2 {
+						return fmt.Errorf("incorrectly specified option: %v", setting)
+					}
+					key := pair[0]
+					value := pair[1]
+					switch key {
+					case "runtime-endpoint":
```
mikebrow, line 86: can we make these keys a list and dump them into the usage for config?

cmd/crictl/config.go
hickeyma: Good point. Re-added the default set behaviour.

cmd/crictl/config.go
hickeyma: Added the options to the description.

hickeyma:
Thanks for the review @mikebrow. Updated and ready for review again.

hickeyma:
Ready for review.

hickeyma:
Thanks for reviews @saschagrunert @mikebrow and for merging @saschagrunert

Issue #605, opened 2020-05-11
config command overwrites config


mikebrow:
What I expected to happen was to change only the runtime-endpoint in the config.

But what happened was it generated an entirely new config with defaults overwriting the prior config.
```
mike@mike-VirtualBox:~/go/src/k8s.io/kubernetes$ cat /etc/crictl.yaml
# important note: do not delete me
runtime-endpoint: unix:///run/containerd/containerd.sock

mike@mike-VirtualBox:~/go/src/k8s.io/kubernetes$ sudo crictl config runtime-endpoint unix:///run/crio/crio.sock

mike@mike-VirtualBox:~/go/src/k8s.io/kubernetes$ cat /etc/crictl.yaml
runtime-endpoint: unix:///run/crio/crio.sock
image-endpoint: ""
timeout: 0
debug: false
```

Suggest a --default for generating a new config, and modify current behavior to not delete comment lines blank lines other text in the config.

mikebrow:
also.. zero timeout seems wrong

usage says the default should be 2s

mikebrow:
looks like the func ReadConfig(filepath string) (*Config, error) { code in pkg/common/file.go is missing the default values for the config it creates.. probably need a newConfig func and pass that to ReadConfig... or something to that effect.

mikebrow:
so two issues... when setting a config value the config command 
1) fails to generate a config struct with the correct defaults
2) overwrites the config file vs simply inserting new values, possibly resulting in loss of information

hickeyma:
@mikebrow Pushed #606 as a fix for timeout issues.

hickeyma:
@mikebrow I can work on PR for the writing only changes to file if that's ok?

mikebrow:
Additional issue with config command:
```
mike@mike-VirtualBox:~/go/src/github.com/kubernetes-sigs/cri-tools$ sudo cat /etc/crictl.yaml
cat: /etc/crictl.yaml: No such file or directory
mike@mike-VirtualBox:~/go/src/github.com/kubernetes-sigs/cri-tools$ sudo crictl config
NAME:
   crictl config - Get and set crictl options.

The options are as follows:

- runtime-endpoint: Container runtime endpoint
- image-endpoint: Image endpoint
- timeout: Timeout of connecting to server
- debug: Enable debug output


USAGE:
   crictl config [command options] [<options>]

OPTIONS:
   --get value  show the option value
[...]

PR #606, opened 2020-05-11
Fix server connection timeout

hickeyma:
This PR handles the following issues found with the `timeout` setting in the tools:

- The default server connection timeout is stated in documentation as 10s but in code it is set to 2s. This doc is now changed to 2s.
- It allows a negative timeout value  which is probably not a very good setting as the system will give it an indeterminate value. This means an unknown or indefinite timeout. This now uses the default value when set to negative.
- Finally, there seemed to some bugs the way the timeout value is retrieved and handled from config file and CLI. It doesn't seem to be converted to seconds as expected.

Partial #605 

[EDITED]

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -102,7 +102,7 @@ func getConnection(endPoints []string) (*grpc.ClientConn, error) {
 	endPointsLen := len(endPoints)
 	var conn *grpc.ClientConn
 	for indx, endPoint := range endPoints {
-		logrus.Debugf("connect using endpoint: %s", endPoint)
+		logrus.Debugf("connect using endpoint '%s' with '%s' timeout", endPoint, Timeout)
 		addr, dialer, err := util.GetAddressAndDialer(endPoint)
 		if err != nil {
 			if indx == endPointsLen-1 {
@@ -203,7 +203,7 @@ func main() {
 			Name:    "timeout",
 			Aliases: []string{"t"},
 			Value:   defaultTimeout,
-			Usage: "Timeout of connecting to the server. " +
+			Usage: "Timeout of connecting to the server in seconds (e.g. 2s, 20s.). " +
 				"0 or less is set to default",
 		},
 		&cli.BoolFlag{
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -203,12 +203,13 @@ func main() {
 			Name:    "timeout",
 			Aliases: []string{"t"},
 			Value:   defaultTimeout,
-			Usage:   "Timeout of connecting to the server (default: 2s)",
+			Usage: "Timeout of connecting to the server. " +
+				"0 or less is set to default",
 		},
 		&cli.BoolFlag{
 			Name:    "debug",
 			Aliases: []string{"D"},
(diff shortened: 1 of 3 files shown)


mikebrow:
F0511 21:19:23.550924   21939 server.go:152] unknown flag: --experimental-dockershim

mikebrow:
sorry @hickeyma I missed this the first time I skimmed the logs.. doing to many things at once

so there's more than one issue... the timeout value was one.. but unrelated to the travis fail.

mikebrow:
https://github.com/kubernetes/kubernetes/commit/53adde65ce000c4d90ee8f807e90658426733a52#diff-41db1a9454b65c65b4c44937886b5880

hickeyma:
@mikebrow Thanks for the review. Updated and ready for review again. Description also updated and logic changed from initial commit.

hickeyma:
> see https://github.com/kubernetes-sigs/cri-tools/blob/master/hack/run-critest.sh#L33 looks like k8s finally removed that flag..

Thanks for heads up @mikebrow. PR #607 now opened for that issue.

hickeyma:
CI now passing. PR is good for review again.

cmd/crictl/main.go
```diff
@@ -196,12 +203,12 @@ func main() {
 			Name:    "timeout",
 			Aliases: []string{"t"},
 			Value:   defaultTimeout,
-			Usage:   "Timeout of connecting to the server",
+			Usage:   "Timeout of connecting to the server (default: 2s)",
```
saschagrunert, line 206:
```suggestion
			Usage:   fmt.Sprintf("Timeout of connecting to the server (default: %ds)", defaultTimeout),
```

cmd/crictl/main.go
```diff
@@ -196,12 +203,12 @@ func main() {
...
 		},
 		&cli.BoolFlag{
 			Name:    "debug",
 			Aliases: []string{"D"},
-			Usage:   "Enable debug mode",
+			Usage:   "Enable debug mode (Default: false)",
```
saschagrunert, line 211: Do we have to mention that?

cmd/crictl/main.go
hickeyma: Thanks for the suggestion. Seemed to miss this previously but it is not needed as CLI lib prints the default.

cmd/crictl/main.go
hickeyma: Same comment as above flag

hickeyma:
Thanks for review and comments @saschagrunert and @mikebrow. Updated and ready for review again.

hickeyma:
Thanks for the review comments @mikebrow. I have updated the debug message as suggested.

> also in testing there's an odd issue here where the config timeout format must be int and the --timeout option must be int+s for example 2s not 2.

This is a tricky one because if I change this so that both the CLI flag and the config file property use the same data type then I would break backwards compatibility. I added some extra text to flag and doc to be more explicit on it. Hope this helps.

hickeyma:
Thanks for review @mikebrow @saschagrunert @feiskyer and for merging @feiskyer

PR #621, opened 2020-06-05
Add CRI-O GitHub action artifact

saschagrunert:
We now use the CRI-O logs as artifacts and rename the containerd logs
that they are better distinguishable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,10 +63,10 @@ jobs:
           make test-cri
         working-directory: ${{ github.workspace }}/src/github.com/containerd/cri
 
-      - name: Upload CRI Test log file
+      - name: Upload containerd logs
         uses: actions/upload-artifact@v1
         with:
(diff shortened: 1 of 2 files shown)


saschagrunert:
@feiskyer it seems impossible for me to discard the annotation parsing in GitHub actions, because:

- GitHub actions does not seem to have a feature for disabling the parser
- The error message is directly coming from the runtime client via klog. And the klog `SetOutput(ioutil.discard)` does simply not work: https://github.com/kubernetes/klog/issues/87

I think disabling the logging for the single AppArmor test would be kind of a hack and the most reasonable way to fix that issue seems to be changing the annotation parser in GitHub actions... :shrug:

saschagrunert:
This is now ready for review in any case.

feiskyer:
> it seems impossible for me to discard the annotation parsing in GitHub actions,

OK, then let's keep it as is. The test itself has passed anyway.

PR #624, opened 2020-06-10
Add waits for creations/modifications of timing-sensitive files

ktock:
Following up: #618

In CRI validation tests, there are some subtests which expect some files in testing containers to be created or modified by other processes (e.g. HostPID test expects `/var/run/nginx.pid` to be created by entrypoint process in the nginx container). However, some of these subtests have no waiting logic for creation/modification of these files and race conditions occasionally occur.

This commit mitigates this by adding some waits for creations/modifications of timing-sensitive files.

### DIFF
--- a/pkg/validate/security_context_linux.go
+++ b/pkg/validate/security_context_linux.go
@@ -94,7 +94,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
 
 			By("get nginx container pid")
-			command := []string{"cat", "/var/run/nginx.pid"}
+			command := []string{"sh", "-c", "while [ ! -f /var/run/nginx.pid ]; do sleep 1; done", "&&", "cat", "/var/run/nginx.pid"}
 			output := execSyncContainer(rc, containerID, command)
 			nginxPid := strings.TrimSpace(string(output))
 			framework.Logf("Nginx's pid is %q", nginxPid)
@@ -212,6 +212,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			By("get nginx container pid")
 			command := []string{"cat", "/proc/1/cmdline"}
+			time.Sleep(time.Second) // waits for nginx to be up-and-running
 			o := execSyncContainer(rc, containerID, command)
 			Expect(o).ToNot(ContainSubstring("master process"))
 		})
@@ -236,9 +237,8 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
 
 			By("get nginx container pid")
-			command := []string{"cat", "/proc/1/cmdline"}
-			o := execSyncContainer(rc, containerID, command)
-			Expect(o).To(ContainSubstring("master process"))
+			command := []string{"sh", "-c", `while ! cat /proc/1/cmdline | grep "master process"; do sleep 1; done`}
+			execSyncContainer(rc, containerID, command)
 		})
 
 		It("runtime should support HostNetwork is true", func() {


ktock:
CI fails because of:
```console
# cd .; git clone -- https://github.com/onsi/ginkgo /home/runner/work/cri-tools/cri-tools/src/github.com/onsi/ginkgo
Cloning into '/home/runner/work/cri-tools/cri-tools/src/github.com/onsi/ginkgo'...
fatal: unable to access 'https://github.com/onsi/ginkgo/': Could not resolve host: github.com
package github.com/onsi/ginkgo/ginkgo: exit status 128
##[error]Process completed with exit code 1.
```
Maybe unrelated to this patch :thinking:

pkg/validate/security_context_linux.go
```diff
@@ -95,8 +95,14 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			By("get nginx container pid")
 			command := []string{"cat", "/var/run/nginx.pid"}
-			output := execSyncContainer(rc, containerID, command)
```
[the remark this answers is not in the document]
ktock: Sometimes `cat` is executed before `/var/run/nginx.pid` is created by nginx process. So we get no error from execution itself but `cat` reports "not found". So, to be more accurate, `Eventually` here is for automatic retrying rather than waiting.

pkg/validate/security_context_linux.go
```diff
@@ -237,8 +244,9 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			By("get nginx container pid")
 			command := []string{"cat", "/proc/1/cmdline"}
-			o := execSyncContainer(rc, containerID, command)
-			Expect(o).To(ContainSubstring("master process"))
+			Eventually(func() string {
+				return execSyncContainer(rc, containerID, command)
```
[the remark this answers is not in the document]
ktock: Sometimes, especially on slow filesystems, `cat` is executed before nginx process change the contents of `/proc/1/cmdline` to the string containing `master process`. So we get no error from the execution itself but `cat` reports a string which doesn't meet the condition `Expect(o).To(ContainSubstring("master process"))`. So, same as the above, `Eventually` here is for automatic retrying rather than waiting. Please tell me if I'm missing something about the usage of `Eventually`.

> if there is a valid reason to loop should we not be looping inside execSyncContainer() vs around it?

Should we directly use `internalapi.RuntimeService.ExecSync` here?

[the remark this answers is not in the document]
ktock: And thanks for your comments!

pkg/validate/security_context_linux.go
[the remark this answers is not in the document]
mikebrow: Ah I see now.. thx :-)
maybe try:
`sh -c "while [ ! -f /var/run/nginx.pid ]; do sleep 1; done" && cat /var/run/nginx.pid`

pkg/validate/security_context_linux.go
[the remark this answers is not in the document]
mikebrow: I see now, thx. Did not realize what you meant by timing sensitive file creation. Did not think that the proc / pid info would not be populated at start of the container before the exec sync occurred.

If possible maybe you will want to consider looping in the container waiting on contents up to the 30sec timeout.. rather than exec multiple times. Either way this LGTM.

pkg/validate/security_context_linux.go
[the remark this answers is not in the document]
ktock: Thanks for the comment. I fixed it.

pkg/validate/security_context_linux.go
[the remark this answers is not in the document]
ktock: Thanks for the comment. I fixed it to wait on the contents inside the container up to the timeout.

saschagrunert:
Fix is incoming in #626

pkg/validate/security_context_linux.go
```diff
@@ -236,9 +237,8 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			}, time.Minute, time.Second*4).Should(Equal(runtimeapi.ContainerState_CONTAINER_RUNNING))
 
 			By("get nginx container pid")
-			command := []string{"cat", "/proc/1/cmdline"}
-			o := execSyncContainer(rc, containerID, command)
-			Expect(o).To(ContainSubstring("master process"))
+			command := []string{"sh", "-c", `while ! cat /proc/1/cmdline | grep "master process"; do sleep 1; done`}
```
SergeyKanzhelev, line 240: nit: this command will not produce any output so if test will fail due to timeout, it will be impossible to say what the content of the `/proc/1/cmdline` was. Would the similar command as in the comment above work better?

pkg/validate/security_context_linux.go
```diff
@@ -212,6 +212,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			By("get nginx container pid")
 			command := []string{"cat", "/proc/1/cmdline"}
+			time.Sleep(time.Second) // waits for nginx to be up-and-running
```
SergeyKanzhelev, line 215: I'm curious why this test case fixed with the delay when the next occurrence is  fixed with the re-tryable command?

mikebrow: I think he wants this one to fail :-)

ktock: Yes, this test expects `ToNot(ContainSubstring("master process"))` so I thought retrying this until timeout will result in the very slow test.

ktock:
#626 has gotten merged. Can we retest this PR?

pkg/validate/security_context_linux.go
ktock: One of the possible solutions might be using `Eventually` with `ExecSync` like my initial patch https://github.com/kubernetes-sigs/cri-tools/commit/3ef4d3eabde5a76540f73594d5cd466636d8b009 ?
@mikebrow WDYT?

saschagrunert:
Yes, can you please rebase on the latest master branch?

pkg/validate/security_context_linux.go
mikebrow: I think what he's saying is "while [ ! -f /proc/1/cmdline ]; do sleep 1; done", "&&", "cat", "/proc/1/cmdline"

ktock: I don't think the command solves this problem because the timing-sensitive point here is the *contents* of the file rather than the existence.

After the entrypoint of the nginx image starts, `/proc/1/cmdline` exists but the initial content is `sh /docker-entrypoint.sh nginx -g daemon off;`. After a while, the entrypoint rewrites the content of this file to `nginx: master process nginx -g daemon off;` which the test expects to read. If the test reads this file before the rewrite occurs, the test fails. This is what I saw in our environment and this PR intends to fix it.

So we need to wait for the contents rather than the existence. (Maybe we should move to a more test-friendly image than nginx but let's do it on a following-up PR)

ktock: But thanks for your suggestion!

feiskyer: OK, let's get this merged.

@ktock would you like open an issue to track the following improvements?

PR #626, opened 2020-06-11
Exclude port forward test because of DNS issues

saschagrunert:
We encounter DNS resolving issues when running via a masquerade iptables
rule. We skip the test for now because I found no real solution which
works with GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,17 +94,6 @@ jobs:
           echo "::set-env name=GOPATH::${{ github.workspace }}"
           echo "::add-path::${{ github.workspace }}/bin"
 
-      - name: Setup system
-        run: |
-          # enable necessary kernel modules
-          sudo ip6tables --list >/dev/null
-
-          # enable necessary sysctls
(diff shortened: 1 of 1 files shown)


saschagrunert:
Tried various configurations but updating the DNS config via systemd-resolved breaks the CI completely. The issue with the masquerade seems to be a known problem, I suggest that we skip the test for now. PTAL @hickeyma @mikebrow

hickeyma:
SGTM @saschagrunert. Best option until it gets sorted out. Thanks for looking into this.

hickeyma:
BTW @saschagrunert, do you have the known issue we could document here?

saschagrunert:
> BTW @saschagrunert, do you have the known issue we could document here?

Hm no I just linked the two stackoverflow discussions which seemed related.

hickeyma:
LGTM NB

mikebrow:
@Random-Liu @dchen1107 @feiskyer @heartlock @mrunalp @yujuhong 
Need a review and hopefully a merge to get master going again pls.. only skipping one test on one of 3 container runtimes

Issue #632, opened 2020-06-30
RemoveContainer rule not being tested properly


mikebrow:
The CRI api states:
```
	// RemoveContainer removes the container. If the container is running, the
	// container must be forcibly removed.
	// This call is idempotent, and must not return an error if the container has
	// already been removed.
```
https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1alpha2/api.pb.go#L7309-L7313

cri-tools conformance tests:
```
		It("runtime should support removing container [Conformance]", func() {
			By("create container")
			containerID := framework.CreateDefaultContainer(rc, ic, podID, podConfig, "container-for-remove-test-")

			By("test remove container")
			removeContainer(rc, containerID)
			containers := listContainerForID(rc, containerID)
			Expect(containerFound(containers, containerID)).To(BeFalse(), "Container should be removed")
		})
```
https://github.com/kubernetes-sigs/cri-tools/blob/master/pkg/validate/container.go#L111-L119

As shown above the conformance test does not test for a running container.

It is expected, however, that some if not all of the container runtimes will fail to delete a running container because that may? have been the original dockershim behavior. We should either fix the runtimes and add a test case to test for removing a running container or change the CRI RemoveContainer api to include a force option, fix the container runtimes and then add tests for the new option.  @Random-Liu bringing this back up because I noticed some code in the sidecar kep pod that was relying on this working, and didn't see a tracking issue anywhere in CRI or the runtimes.

mikebrow:
dockershim, containerd, and cri-o all support the new test case now so.. closing

PR #633, opened 2020-06-30
add tests for remove container by force rule

mikebrow:
Testing RemoveContainer API to address issue #632

### DIFF
--- a/pkg/framework/util.go
+++ b/pkg/framework/util.go
@@ -52,11 +52,17 @@ var (
 	// DefaultContainerCommand is the default command used for containers
 	DefaultContainerCommand []string
 
+	// DefaultPauseCommand is the default command used for containers
+	DefaultPauseCommand []string
+
 	// DefaultLinuxPodLabels default pod labels for Linux
 	DefaultLinuxPodLabels = map[string]string{}
 
 	// DefaultLinuxContainerCommand default container command for Linux
 	DefaultLinuxContainerCommand = []string{"top"}
+
+	// DefaultLinuxPauseCommand default container command for Linux pause
+	DefaultLinuxPauseCommand = []string{"sh", "-c", "top"}
 
 	// DefaultLcowPodLabels default pod labels for Linux containers on Windows
 	DefaultLcowPodLabels = map[string]string{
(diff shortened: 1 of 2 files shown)


mikebrow:
@hickeyma heads up.. I expected a fail for containerd and possibly dockershim here .. but containerd passed. I may not have the testcase right yet, or my assumption of fail may have been wrong. However, I checked out the logs and it seems this PR's change is only being tested on cri-o not containerd. So there's a logic bug in the actions test for containerd somewhere, such that containerd is running on a different version of critest than it should be. Not sure if this is a regression.. I thought we checked the version tag that was being run for containerd. Maybe we should be outputting the build tag in the critest output if possible.

cri-o:
```
[It] runtime should support removing created container [Conformance]
```
containerd:
```
[It] runtime should support removing container [Conformance]
```

hickeyma:
> I expected a fail for containerd and possibly dockershim here .. but containerd passed.

@mikebrow Not sure why this isn't tested on containerd. It calls `make test-cri` in https://github.com/kubernetes-sigs/cri-tools/blob/master/.github/workflows/containerd.yml#L107 which calls the target in https://github.com/containerd/cri/blob/master/hack/test-cri.sh. Looking at this, it does not skip any of the tests.

Re dockershim, it does skip some tests as follows: https://github.com/kubernetes-sigs/cri-tools/blob/master/hack/run-critest.sh#L43

mikebrow:
> > I expected a fail for containerd and possibly dockershim here .. but containerd passed.
> 
> @mikebrow Not sure why this isn't tested on containerd. It calls `make test-cri` in https://github.com/kubernetes-sigs/cri-tools/blob/master/.github/workflows/containerd.yml#L107 which calls the target in https://github.com/containerd/cri/blob/master/hack/test-cri.sh. Looking at this, it does not skip any of the tests.
> 
> Re dockershim, it does skip some tests as follows: https://github.com/kubernetes-sigs/cri-tools/blob/master/hack/run-critest.sh#L43

yeah it's not a skip issue.. for containerd it's not testing on the PR itself.. could be on the parent commit not sure..

hickeyma:
LGTM NB

PR #643, opened 2020-07-28
ps: document --pod <id> must be a full ID

haircommander:
#### What type of PR is this?

> Uncomment only one `/kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
 /kind documentation

Since ps is simply passing a filter, and filters don't check for substring matches, but rather exact matches,
the ID must be full.

Before we pass the id as a filter, translate truncated id to full id with a PodSandboxStatus request

Signed-off-by: Peter Hunt <pehunt@redhat.com>

Fixes #

```release-note
none
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -869,7 +869,18 @@ func ListContainers(runtimeClient pb.RuntimeServiceClient, imageClient pb.ImageS
 		filter.Id = opts.id
 	}
 	if opts.podID != "" {
-		filter.PodSandboxId = opts.podID
+		// translate to full pod ID
+		request := &pb.PodSandboxStatusRequest{
+			PodSandboxId: opts.podID,
+		}
+		logrus.Debugf("PodSandboxStatusRequest to find full pod ID: %v", request)
+		r, err := runtimeClient.PodSandboxStatus(context.Background(), request)
+		logrus.Debugf("PodSandboxStatusReponse to find full pod ID: %v", r)
+		if err != nil {
+			return errors.Wrapf(err, "failed to translate pod ID %s to full pod ID", opts.podID)
+		}
+
+		filter.PodSandboxId = r.Status.Id
 	}
 	st := &pb.ContainerStateValue{}
 	if !opts.all && opts.state == "" {


mikebrow:
> The other way around, I think it would be nice to change the behavior to accept a short pod ID, wouldn't it?

Agreed. This seems like an omission.

mikebrow:
Drafted a pull request for containerd: https://github.com/containerd/cri/pull/1547

haircommander:
the other option is we could do a PodSandboxStatus request to translate the truncated ID to a full ID, as I've updated this with :smiley: I'm fine either way!

saschagrunert:
> the other option is we could do a PodSandboxStatus request to translate the truncated ID to a full ID, as I've updated this with 😃 I'm fine either way!

Hm, yeah now I’m wondering what the better way is. Changing the behavior in crictl seems more explicit, right? WDYT @mikebrow ?

mikebrow:
> > the other option is we could do a PodSandboxStatus request to translate the truncated ID to a full ID, as I've updated this with 😃 I'm fine either way!
> 
> Hm, yeah now I’m wondering what the better way is. Changing the behavior in crictl seems more explicit, right? WDYT @mikebrow ?

IMO we should apply this with consistency for all of the apis that include an id as a lookup/filter field. Had there been a set of CRI services for doing short to long name conversions or some detail in the api to say where said conversions should be used and where not then we'd probably want test buckets for the rule :-)  Note: this was the only place that in containerd/cri, that I found, where we did a get for the pod id and did not do a lookup for matching purposes. Thus my belief this was an omission. The lack of a rule in the API.. makes me feel like this is a feature of the runtime to support or not. A history quick check shows there was an expectation that the lookup of the id in the filter was expected to be added later, but was not done right away because crictl didn't have the option to set the filter.. this going back 3y :-)

haircommander:
> > > the other option is we could do a PodSandboxStatus request to translate the truncated ID to a full ID, as I've updated this with smiley I'm fine either way!
> > 
> > 
> > Hm, yeah now I’m wondering what the better way is. Changing the behavior in crictl seems more explicit, right? WDYT @mikebrow ?
> 
> IMO we should apply this with consistency for all of the apis that include an id as a lookup/filter field. Had there been a set of CRI services for doing short to long name conversions or some detail in the api to say where said conversions should be used and where not then we'd probably want test buckets for the rule :-) Note: this was the only place in containerd/cri, that I found, where we did a get for the pod id and did not do a lookup for matching purposes. Thus my belief this was an omission. The lack of a rule in the API.. makes me feel like this is a feature of the runtime to support or not. A quick check in history shows there was an expectation that the lookup of the id in the filter was to be added later, but was not done right away (was missed only added to stats not also for container list) .. this going back 3y :-) [containerd/cri#222 (comment)](https://github.com/containerd/cri/issues/222#issuecomment-335896263)

sorry, to be clear, are you saying that we should follow the example in this current PR, but apply it to every option that passes in an ID that could have a short or long name? Or are you saying the CRI implementations should handle the filter more gracefully?

mikebrow:
> sorry, to be clear, are you saying that we should follow the example in this current PR, but apply it to every option that passes in an ID that could have a short or long name? Or are you saying the CRI implementations should handle the filter more gracefully?

"should handle the filter more gracefully" is a rule we do not enforce.. Container runtimes are not required to support short ids. But they should, arguably, be consistent across all of the services. 

IMO it would be confusing, when debugging crictl, if for every container or pod id... we did a CRI status check first just to get the long name then call back to the runtime for the service we are using. 

I pushed a PR into containerd/cri that makes the behavior of container list with pod id filter consistent with all the other pod id use cases.  Thus for containerd, at least, we will not have to change the behavior of crictl to first do a pod status request to lookup the long name to substitute for the filter's pod id.

mikebrow:
That said.. a nice feature to add to the crictl client would be the addition of tab shortcuts.

saschagrunert:
Proposed a fix for CRI-O https://github.com/cri-o/cri-o/pull/4032

feiskyer:
> Hm, yeah now I’m wondering what the better way is. Changing the behavior in crictl seems more explicit, right? WDYT @mikebrow ?

+1 to do this on the crictl client side. Supporting the substring sandbox ID or container ID is not defined from CRI interface. What about we list the IDs and match by substring from crictl?

mikebrow:
> > Hm, yeah now I’m wondering what the better way is. Changing the behavior in crictl seems more explicit, right? WDYT @mikebrow ?
> 
> +1 to do this on the crictl client side. Supporting the substring sandbox ID or container ID is not defined from CRI interface. What about we list the IDs and match by substring from crictl?

via bash completion makes sense, as that is using the users expectation of fill out as is done for most linux client tools.  

Putting the behavior in crictl is the almost the same as setting the expected behavior for the CRI api only it just does it for the client tool, not for the service requests. 

Short IDs have already been supported via container runtime(s) for years. Circumventing the container runtime short id processing by adding the extra CRI calls to get each list... for every id used by every crctl command and sub command.. is expensive resource wise and changes behavior.

saschagrunert:
@haircommander is this something we still want to integrate?

haircommander:
no I don't think so, as it seems we've decided to let the cri implementation handle it

PR #662, opened 2020-09-15
Implement checkpoint/restore commands

adrianreber:
#### What type of PR is this?

/kind feature
/kind api-change

This enables the checkpoint/restore commands. Right now it can be used like this:
```
# crictl checkpoint CTR-ID
# crictl restore CTR-ID
```
To checkpoint and restore a container. It also offers the possibility to restore a container in another pod:
```
# crictl checkpoint CTR-ID
# crictl restore -p POD-ID CTR-ID # POD-ID being the destination pod
```

This PR is just to make sure the changes in https://github.com/cri-o/cri-o/pull/4199 can be tested.


vendor/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto
```diff
@@ -87,6 +87,11 @@ service RuntimeService {
     // Once it returns error, new container log file MUST NOT be created.
     rpc ReopenContainerLog(ReopenContainerLogRequest) returns (ReopenContainerLogResponse) {}
 
+    // CheckpointContainer checkpoints a container
+    rpc CheckpointContainer(CheckpointContainerRequest) returns (CheckpointContainerResponse) {}
```
feiskyer, line 91: where did these two API come from? I didn't find it from cri-api repo.

adrianreber: @feiskyer I was not aware of the correct process, but I now created a KEP for this: https://github.com/kubernetes/enhancements/pull/1990

feiskyer: thanks, it needs to be in the cri-api first

Issue #661, opened 2020-09-14
OCI Image and Distribution conformance


mikebrow:
A suite of CI tests focusing on OCI distribution and OCI image conformance is desired.

At the start of this project the number of clients and registries were fairly low compared to today. Testing but one docker image format (schema 2) on one registry (docker.io) is insufficient for today's CRI environment.  

Support exists already for changing the prefix, registry, for the images used in test. These images are pushed via docker client, then pulled and tested against the CRI runtimes in CI.

Suggest adding OCI specific tests for some number of clients to some number of registries, to begin exercising container image build, push, and pull tooling for more than just one type of container image.

PR #665, opened 2020-09-24
Add test oci images

hickeyma:
#### What type of PR is this?

/kind feature

This adds OCI images which can be used to expand ihe test buckets to test more than one container image. it is currently testing just a docker image.

Partial #661 

Tested against a local docker registry, Docker Hub and GCR. 
TODO: It will need to be run against the `gcr.io/cri-tools` project to load the OCI images for use by the test buckets. I don't have access to this project.

```release-note
NONE
```

### DIFF
--- a/hack/install-buildah.sh
+++ b/hack/install-buildah.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Install buildah
+. /etc/os-release
+sudo sh -c "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/x${ID^}_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list"
+wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/x${ID^}_${VERSION_ID}/Release.key -O Release.key
+sudo apt-key add - < Release.key
+rm -f Release.key
+sudo apt-get update -qq
+sudo apt-get -qq -y install buildah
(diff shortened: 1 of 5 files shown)


hickeyma:
Something to note:

-  This will need to be run against the gcr.io/cri-tools project to load the OCI images for use by the test buckets. Does someone have access to the project?

hickeyma:
@saschagrunert @mikebrow and I had a discussion on slack about pushing images to the cri-tools GCR service account for image tests. Currently we need someone with the authentication to login, pull this PR and run make all-push to refresh the test images. This is all a bit manual and difficult to make changes. A good suggestion by @saschagrunert is to setup a process to automatically build the images continuously via [cloud builder](https://cloud.google.com/cloud-build).

I did some investigation around cloud builder and my feedback is as follows wrt updaing the current process (inc. adding OCI support):

1) Looks like will need a cloud builder file per image supported within each image directory e.g. in `image-user` directory there maybe need for the following build files: `cloudbuild-uid.yaml`, `cloudbuild-username.yaml`, `cloudbuild-pushmanifest.yaml` etc. This is because I cannot see any easy way to layout the build files using lists per architecture. It means therefore that the file with steps per architecture for each image becomes large quickly without adding other images to it. That is the docker images, would also need files for the OCI images.

2) Need to add a step for each architecture build of an image as cannot use a list as per the existing build in the Makefile. Step is similar to the following:

```yaml
 - name: gcr.io/cloud-builders/docker
    id : 'build-amd64'
    args:
      - build
      - --tag=gcr.io/cont-reg-project1/test-image-user-uid-amd64
      - --build-arg=ARCH=amd64
      - --build-arg=USER=1002;
      - .
```

3) May need to push docker images for the different build architectures to GCR repository (e.g. `arm` etc.) as not readily available

4) For building OCI images, would need to push [buildah](https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/buildah) image to GCR repository as it is a [community-contributed](https://github.com/GoogleCloudPlatform/cloud-builders-community) Docker image

5) [Push manifest](https://github.com/estesp/manifest-tool) would need to be contributed as a [community-contributed](https://github.com/GoogleCloudPlatform/cloud-builders-community) Docker image and then pushed to GCR repository to use it

saschagrunert:
@hickeyma we could promote the images directly via adding them to k8s.gcr.io:
https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io

For that we need:
1. A postsubmit job which pushes every image to our staging location, like:
  https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-seccomp-operator.yaml
2. With releases we can promote those images to k8s.gcr.io, like:
  https://github.com/kubernetes/k8s.io/pull/1149/files

Our tests can refer the `:latest` tag or a a release we decide to promote.

Every image (probably architecture) still needs a `cloudbuild.yaml`, which will look like this:
https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/cloudbuild.yaml

PR #666, opened 2020-09-24
Add timeout option to CreateContainer, RunContainer, RunPodSandbox

haircommander:
#### What type of PR is this?

/kind feature

Add timeout option to CreateContainer, RunContainer, RunPodSandbox

This option allows users to test their CRI implementation with a timeout,
similarly to how the kubelet runs requests.

Signed-off-by: Peter Hunt <pehunt@redhat.com>

```release-note
Add `timeout` option to create, run, runp requests, allowing users to cancel CRI requests after a timeout
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -57,6 +57,9 @@ type runOptions struct {
 	// podConfig is path to the config for sandbox
 	podConfig string
 
+	// the create timeout
+	timeout time.Duration
+
 	// the image pull options
 	*pullOptions
 }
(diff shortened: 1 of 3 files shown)


haircommander:
haha @ the PR number :smiling_imp:

cmd/crictl/container.go
```diff
@@ -530,6 +539,11 @@ var runContainerCommand = &cli.Command{
 		Name:    "runtime",
 		Aliases: []string{"r"},
 		Usage:   "Runtime handler to use. Available options are defined by the container runtime.",
+	}, &cli.Int64Flag{
```
saschagrunert, line 542: Let's use `cli.TimestampFlag` here.

saschagrunert (changes requested):
Generally LGTM, just a suggestion around the type :point_down:

cmd/crictl/container.go
haircommander: I am not so sure I like that. it requires users to do a diff of time at the command line, which seems more complicated than just specifying seconds.

saschagrunert: Ah I think I meant `DurationFlag`. Would this fit better? :innocent:

haircommander: ah yes that's better! updated as suggested

saschagrunert: Thank you! :pray:

saschagrunert (approved):
/lgtm

saschagrunert (approved):
/approve

PR #668, opened 2020-09-30
Timeout fixes

haircommander:
#### What type of PR is this?

/kind bug
/kind api-change

Adds two fixes to the runp/create `--timeout` feature
1: correctly use the timeout passed in by not multiplying it by time.Seconds (a relic from when it was originally specified as an int, oops)
2: change the option to `--cancel-timeout` and `-T` to not overload `--timeout` and `-t` (Which has a different meaning in the top level flags, and could confuse users)

```release-note
Update the `timeout` option in create, run, runp requests to be `cancel-timeout`
```


saschagrunert (approved):
/lgtm

PR #674, opened 2020-10-28
CI: bring back port forward test

kolyshkin:
#### What type of PR is this?

/kind feature

(is adding back a CI test a feature? IDK)

This reverts commit 2688ed34bf00a7bc2fe718c748b0fbaf9ab6cb79 (PR #626),
and adds the following change on top:

```diff
-          sudo iptables -t nat -I POSTROUTING -s 127.0.0.1 ! -d 127.0.0.1 -j MASQUERADE
+          sudo iptables -t nat -I POSTROUTING -s 127.0.0.0/8 ! -d 127.0.0.0/8 -j MASQUERADE
```

The original DNS problem was caused by the fact that
systemd-resolved is listening at the IP address 127.0.0.53
and thus it was affected by the above rule.

Change the address to the address block (yes, the whole /8 block
is reserved for localhost addresses!) to fix this.

None

```release-note
None
```

### DIFF
--- a/.github/workflows/crio.yml
+++ b/.github/workflows/crio.yml
@@ -23,6 +23,17 @@ jobs:
           echo "::set-env name=GOPATH::${{ github.workspace }}"
           echo "::add-path::${{ github.workspace }}/bin"
 
+      - name: Setup system
+        run: |
+          # enable necessary kernel modules
+          sudo ip6tables --list >/dev/null
+
+          # enable necessary sysctls
+          sudo sysctl -w net.ipv4.conf.all.route_localnet=1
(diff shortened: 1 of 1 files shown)


kolyshkin:
Draft for now as I'm not sure it will work, just hope so.

kolyshkin:
@saschagrunert PTAL

kolyshkin:
It is working at https://github.com/cri-o/cri-o/pull/4309 so should work here, too. :crossed_fingers:

saschagrunert (approved):
NIce, thank you for the fix! :pray: 
/lgtm
/approve

Issue #678, opened 2020-11-09
Add tests for CRI security profiles (seccomp/AppArmor)


saschagrunert:
The Container Runtime Interface will feature AppArmor and Seccomp security profiles as replacement for the now deprecated profile paths. We should update the test suite to verify those new fields, too.

PR #681, opened 2020-11-09
WIP: Update seccomp tests to use CRI security profile field

saschagrunert:
#### What type of PR is this?

/kind feature

The CRI validation for the security context now uses the seccomp field
instead of the deprecated seccomp profile path. This means that the
docker/default tests are now testing for `RuntimeDefault`. Invalid test
cases have been removed, too.

Refers to #678 

```release-note
critest now uses the new security profile field instead of the seccomp profile path for validation testing
```


saschagrunert:
cc @mrunalp @mikebrow 

I expect the CI to fail here…

mikebrow:
> cc @mrunalp @mikebrow
> 
> I expect the CI to fail here… As an alternative, we could also duplicate the tests to still preserve the old ones.

I think it makes sense to test both, yes.

PR #682, opened 2020-11-12
pkg/validate/apparmor: fix path

kolyshkin:
#### What type of PR is this?

/kind bug
/kind failing-test

A change introduced by [1] makes it impossible to create a directory
entry right under container's root directory, so the test case
"should enforce a permissive profile" fails.

For the purpose of this test it does not matter where we create the
file, so let's do it under /tmp.

[1] https://github.com/containers/storage/pull/743

None

n/a 

```release-note
NONE
```


kolyshkin:
Looks like it's working (tested in https://github.com/cri-o/cri-o/pull/4366) so removed the WIP.

kolyshkin:
@saschagrunert @haircommander PTAL

PR #684, opened 2020-11-14
Makefile: assorted improvements

kolyshkin:
#### What type of PR is this?

/kind cleanup

Optionally add one or more of the following kinds if applicable:

#### What this PR does / why we need it:

Assorted Makefile improvements. Please see individual commits for details.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

~~CI is failing now, the fix is in https://github.com/kubernetes-sigs/cri-tools/pull/689. Once that one is merged, this needs to be rebased to pass CI.~~ Rebased

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #689, opened 2020-11-18
.github/workflows: fix GOROOT under sudo etc

kolyshkin:
#### What type of PR is this?

/kind bug
/kind failing-test

##### 1. .github/workflows/containerd.yml: simplify
    
Remove `BDIR=$BDIR` from sudo invocations, as it is not used by commands
run by sudo, and the command line of sudo itself is built before running
sudo. Hope this explanation makes sense.
    
Also remove `PATH=$PATH GOPATH=$GOPATH` assignments. Those could make
sense with sudo, but do not make any sense without.

##### 2. github/workflows: fix GOROOT under sudo

It was found that using sudo (as in e.g. sudo make install) results in
wrong GOROOT setting, which leads to various weird compilation bugs.

Here is how GOROOT is changed:

```diff
-GOROOT="/opt/hostedtoolcache/go/1.15.4/x64"
+GOROOT="/opt/hostedtoolcache/go/1.14.11/x64"
```
Here are some of the errors caused by it:

	# runtime/internal/sys
	compile: version "go1.14.11" does not match go tool version "go1.15.4"
	# sync/atomic
	flag provided but not defined: -p

Let's use `sudo -E PATH=$PATH` for all cases which might call go.
This will make sure that proper go version is used (from $PATH), and
also that all the environment variables that it might need are correctly
set.

##### 4. .github/workflows/crio.yml: fix a deprecation notice

Fixes this:
    
> GoogleCloudPlatform/github-actions/setup-gcloud has been deprecated.
> Please use google-github-actions/setup-gcloud

None

* Please see individual commits for details.
* This should fix CI failure observed in https://github.com/kubernetes-sigs/cri-tools/pull/684
* This PR is additionally tested in https://github.com/kubernetes-sigs/cri-tools/pull/690

```release-note
NONE
```

### DIFF
--- a/.github/workflows/containerd.yml
+++ b/.github/workflows/containerd.yml
@@ -273,11 +273,11 @@ jobs:
             [plugins.cri.containerd.default_runtime]
               runtime_type = \"${{matrix.runtime}}\"
           EOF
-          sudo PATH=$PATH BDIR=$BDIR /usr/local/bin/containerd -a ${BDIR}/c.sock -root ${BDIR}/root -state ${BDIR}/state -log-level debug &> ${BDIR}/containerd-cri.log &
-          sudo BDIR=$BDIR /usr/local/bin/ctr -a ${BDIR}/c.sock version
-          sudo BDIR=$BDIR /usr/local/sbin/runc --version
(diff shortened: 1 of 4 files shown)


kolyshkin:
@saschagrunert @feiskyer PTAL

feiskyer (commented):
This is great! thanks for the fix.

/lgtm
/approve

PR #691, opened 2020-11-18
.github/workflows: add go mod and build cache

kolyshkin:
#### What type of PR is this?

/kind feature

_(can we have kind ci please?)_

Adds caching of go mod and go build, to presumably speed up github actions.

None

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,16 @@ jobs:
           echo "GOPATH=${{ github.workspace }}" >> $GITHUB_ENV
           echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
 
+      - name: Cache go modules and build cache
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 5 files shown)


kolyshkin:
close/reopen to re-trigger CI

feiskyer:
interesting. @kolyshkin have you checked how much time would it help to reduce?

kolyshkin:
Hmm, I missed the fact that this repo uses vendor. In this case having go mod cache makes less sense.

Or perhaps we can drop vendor.

kolyshkin:
> have you checked how much time would it help to reduce?

I will do some measurements later this week. I think the impact could be bigger without vendor/

kolyshkin:
Let's close and open to re-run ci. Now we do have cache more or less populated and it might show.

saschagrunert (approved):
/lgtm

PR #693, opened 2020-11-18
Makefile: avoid rebuilding binaries

kolyshkin:
#### What type of PR is this?

/kind cleanup
/kind regression

_(this is a continuation of #684; sorry for the mess)_

It is a common scenario in which someone runs

	make
	sudo make install

Problem is, with the current Makefile, which has install-crictl depend
on a phony target crictl (and same for critest), it leads to making
crictl (and critest) targets twice. It is OK per se, since go is very
effective in caching dependencies and build artefacts, but it can be
problematic with sudo in case go is installed in some non-standard
path and/or requires some environment variables that sudo skips
(unless -E is used). As a result, binaries can be rebuilt again with
the wrong version of go, or some wrong environment, or maybe go binary
can't be found at all due to changed $PATH.

Anyway, to fix all this, let's create real (rather than phony) targets,
and have install depend on it.

At the same time, it would be good to have make always try to rebuild
the binaries, since it's not aware of their dependencies (but go does).
So, have the phony targets `crictl` and `critest` which run `make -B`
to force rebuild.

In other words, `make` (or `make binaries`, or `make crictl`) will now
always rebuild the binaries (same as before), but `make install` will
only try to build binaries if they are absent.

While at it:
 - put the built binaries into BUILD_BIN_DIR
 - consolidate and simplify some targets
 - remove non-existent targets from .PHONY

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

None

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+MAKEFLAGS += --no-print-directory
 GO ?= go
 
 # test for go module support
@@ -47,6 +48,9 @@ endef
 GINKGO := $(BUILD_BIN_PATH)/ginkgo
 GOLANGCI_LINT := $(BUILD_BIN_PATH)/golangci-lint
 
+CRITEST := $(BUILD_BIN_PATH)/critest$(BIN_EXT)
+CRICTL := $(BUILD_BIN_PATH)/crictl$(BIN_EXT)
+
 all: binaries
 
 help:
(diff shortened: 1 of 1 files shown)


kolyshkin:
This is still messed up. Ideally, I'd love to have `make` rebuild the binaries, but `make install` to use the prebuilt ones, if available, and rebuild them if not. Not sure how to implement it though. Need to think.

kolyshkin:
Rebased to remove the stale flag.

Is there anything that I need to address?

kolyshkin:
@saschagrunert PTAL

saschagrunert (approved):
/lgtm

Issue #687, opened 2020-11-17
Promote test images to k8s.gcr.io


hickeyma:
#### What would you like to be added:

- [x] Add cri-tools to k8s.gcr.io: https://github.com/kubernetes/k8s.io/pull/1431
- [ ] Add new [cloud builder](https://cloud.google.com/cloud-build) process to build test images and push images to staging GCR repos:
     - [x] Test user images: #696 
     - [ ] Test images
     - [ ] Test win images
     - [ ] Test hostnet images
- [ ] Testinfra postsubmit job which pushes images to staging location:
     - [x] Test user images prow job: https://github.com/kubernetes/test-infra/pull/20092, https://github.com/kubernetes/test-infra/pull/20200, https://github.com/kubernetes/test-infra/pull/20211, https://github.com/kubernetes/test-infra/pull/20250
     - [ ] Test images  prow job
     - [ ] Test win images  prow job
     - [ ] Test hostnet images  prow job
- [ ] Promote the test images
- [ ] Add all test images to the k8s image list
- [ ] Remove the old test images process

#### Why is this needed:

The test images are pushed to the cri-tools GCR service account for unit tests. Currently we need someone with the authentication to login, pull this PR and run make all-push to refresh the test images. This is all a bit manual and difficult to make changes. 

One way of doing this is to automatically build the images continuously via [cloud builder](https://cloud.google.com/cloud-build) . The process however seems to require a lot of overhead to setup as described in https://github.com/kubernetes-sigs/cri-tools/pull/665#issuecomment-724912726.

Another approach is to promote the images by adding them to k8s.gcr.io and using the existing k8s registry infrastructure. This is described in https://github.com/kubernetes-sigs/cri-tools/pull/665#issuecomment-725281993. This is the approach to be used.

Thanks to @saschagrunert for feedback on different approaches.

PR #696, opened 2020-11-25
Add mechanism for building and pushing test user images to staging location

hickeyma:
#### What type of PR is this?

/kind feature

Optionally add one or more of the following kinds if applicable:
/kind test-infra

The test images are pushed to the cri-tools GCR service account for unit tests. Currently we need someone with the authentication to login, pull this PR and run make all-push to refresh the test images. This is all a bit manual and difficult to make changes. This PR adds mechanism to use staging from Kubernetes test-infra to automate the process.

Partial #687 

This is a first draft at finding a pattern to build and push test images to staging location. At the moment, there is some repetition especially in manifest Makefile as [make GCB process](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#makefile-build-example) doesn't seem to handle including other makefiles like in [hack](https://github.com/kubernetes-sigs/cri-tools/tree/master/hack/make-rules). It seems to require a make file to be named `Makefile` also.
 
This can be tested locally by connecting to you GCR instance from CLI and then running in order:

1. gcloud builds submit --config cloudbuild-images.yaml --substitutions=_IMAGE_TYPE="uid"
2. gcloud builds submit --config cloudbuild-images.yaml --substitutions=_IMAGE_TYPE="username"
3. gcloud builds submit --config cloudbuild-images.yaml --substitutions=_IMAGE_TYPE="uid-group"
4. gcloud builds submit --config cloudbuild-images.yaml --substitutions=_IMAGE_TYPE="username-group"
5. gcloud builds submit --config cloudbuild-manifests.yaml

```release-note
NONE
```

### DIFF
--- a/images/image-user/cloudbuild-images.yaml
+++ b/images/image-user/cloudbuild-images.yaml
@@ -1,0 +1,57 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+timeout: 1200s
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: gcr.io/cloud-builders/docker
+    id : 'build-amd64'
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-amd64
+      - --build-arg=ARCH=amd64
+      - --build-arg=USER=1002;
+      - .
+  - name: gcr.io/cloud-builders/docker
+    id : 'build-arm'
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-arm
+      - --build-arg=ARCH=arm32v6
+      - --build-arg=USER=1002;
+      - .
+  - name: gcr.io/cloud-builders/docker
+    id : 'build-arm64'
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-arm64
+      - --build-arg=ARCH=arm64v8
+      - --build-arg=USER=1002;
+      - .
+  - name: gcr.io/cloud-builders/docker
+    id : 'build-ppc64le'
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-ppc64le
+      - --build-arg=ARCH=ppc64le
+      - --build-arg=USER=1002;
+      - .
+  - name: gcr.io/cloud-builders/docker
+    id : 'build-s390x'
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-s390x
+      - --build-arg=ARCH=s390x
+      - --build-arg=USER=1002;
+      - .
+substitutions:
+  # _IMAGE_TYPE needs to be set to the image that is to be built
+  # Can be one of the following: 'uid', 'username', 'uid-group'
+  # or 'username-group'
+  _IMAGE_TYPE: 'uid'
+images:
+  - 'gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-amd64'
+  - 'gcr.io/$PROJECT_ID/test-image-user-$_IMAGE_TYPE-arm'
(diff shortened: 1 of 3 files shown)


images/image-user/Makefile-orig
```diff
@@ -0,0 +1,45 @@
+# Copyright 2018 The Kubernetes Authors.
```
saschagrunert, line 1: I'm wondering why we should keep that file.

images/image-user/cloudbuild-images.yaml
```diff
@@ -0,0 +1,57 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
```
saschagrunert, line 1: I think this file has to be called `cloudbuild.yaml` and we need one for each image (in a dedicated directory)

images/image-user/cloudbuild-manifests.yaml
```diff
@@ -0,0 +1,10 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
```
saschagrunert, line 1: I think we do not need that file in this repository. What is the intention of it?

images/image-user/Makefile-orig
hickeyma: Good point, just left it in incase for reference. Will remove it in next commit.

images/image-user/cloudbuild-images.yaml
hickeyma: I can change the name and add it to images sub-directory. Do we need a file per image? Can it be called with the flag `--substitutions=_IMAGE_TYPE="username-group"`, changing the type for each image?

images/image-user/cloudbuild-manifests.yaml
hickeyma: It creates multi-arch images for non-x86 architecture. Refer to #287 for more details.

hickeyma:
Thanks for the review @saschagrunert. I have replied to your comments. Let me know what you think.

images/image-user/cloudbuild-images.yaml
saschagrunert: It might work, yeah. The image builder (like `gcr.io/k8s-testimages/image-builder:v20200901-ab141a0`) runs a `/builder` binary, which supports the following CLI flags:
```
bash-4.4# ./builder -h
Usage of ./builder:
  -allow-dirty
        If true, allow pushing dirty builds.
  -build-dir string
        If provided, this directory will be uploaded as the source for the Google Cloud Build run.
  -env-passthrough string
        Comma-separated list of specified environment variables to be passed to GCB as substitutions with an _ prefix. If the variable doesn't exist, the substitution will exist but be empty.
  -gcb-config string
        If provided, this will be used as the name of the Google Cloud Build config file. (default "cloudbuild.yaml")
  -log-dir string
        If provided, build logs will be sent to files in this directory instead of to stdout/stderr.
  -no-source
        If true, no source will be uploaded with this build.
  -project string
        If specified, use a non-default GCP project.
  -scratch-bucket string
        The complete GCS path for Cloud Build to store scratch files (sources, logs).
  -variant string
        If specified, build only the given variant. An error if no variants are defined.
  -with-git-dir
        If true, upload the .git directory to GCB, so we can e.g. get the git log and tag.
```

images/image-user/Makefile-orig
saschagrunert: Thanks! :pray:

images/image-user/cloudbuild-images.yaml
hickeyma: Could use `-env-passthrough` to pass `IMAGE_TYPE` flag which means that can use the one cloud build file creating the different images for `image-user`.  The `-gcb-config` flag means can pass the cloud build filename. There should be able to keep names `cloudbuild-images.yaml` and `cloudbuild-manifests.yaml`. What do you think @saschagrunert?

images/image-user/Makefile-orig
hickeyma: Done

images/image-user/cloudbuild-images.yaml
saschagrunert: Sounds good to me :+1:

saschagrunert (approved):
/lgtm

hickeyma:
Thanks for the review and merge @saschagrunert

PR #708, opened 2020-12-19
fix a bug in closure function

yuqitao:
#### What type of PR is this?

/kind bug

fix a bug in closure function , when run `crictl rmp -all`, only remove one pod not all pods 

None

no

no


feiskyer (commented):
thanks for the fix
/lgtm
/approve

Issue #716, opened 2021-01-26
Provide .sha256 files in release artifiacts


micahhausler:
#### What would you like to be added:

#322 and #323 addressed producing .sha256 output to the release text, but I'd like `<filename>.sha256` files produced as well.

#### Why is this needed:

It'd be helpful to be able to programmatically fetch the shasums of artifacts so separate build processes that verify the integrity of the download. I'm working on kubernetes-sigs/kind#2033 and would like to fetch these programmatically.

saschagrunert:
I agree this seems useful. :+1:

MaXinjian:
Seems now still no .sha256 files  in release artifiacts.  [Modify hack/release.sh script to include sha256 sum files #717] (https://github.com/kubernetes-sigs/cri-tools/pull/717) only show sha256 context

PR #717, opened 2021-01-27
Modify hack/release.sh script to include sha256 sum files

saschagrunert:
#### What type of PR is this?

/kind feature

We now add a *.sha256 file side by side to the tarball, which can be
uploaded as GitHub artifact. Beside this, the script now directly
outputs the Markdown table to lazy copy-paste. The script also now
includes the full file which has been build and not only it's name to
provide a more useful and verbose output.

Closes https://github.com/kubernetes-sigs/cri-tools/issues/716
None

```release-note
Added *.sha256 sum files to release artifacts
```

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Copyright 2018 The Kubernetes Authors.
 #
@@ -20,8 +20,6 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-UPLOAD_GCS=${UPLOAD_GCS:-false}
-GCS_PATH=${GCS_PATH:-"kubernetes-release/crictl"}
 VERSION=$(git describe --abbrev=0 --tag)
 CRI_CTL_PLATFORMS=(
     linux/amd64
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -90,22 +86,23 @@ for platform in "${CRI_TEST_PLATFORMS[@]}"; do
...
+printf "| file | sha256 |\n"
+printf "| ---- | ------ |\n"
+
 # Show sha256 for release files
 if [[ "${OSTYPE}" == "darwin"* ]]; then
-  for file in $(ls ${CRI_TOOLS_ROOT}/_output/releases); do
-    echo "$file $(shasum -a 256 ${CRI_TOOLS_ROOT}/_output/releases/$file | awk '{print $1}')"
-  done
+    for file in $(ls "$OUTPUTDIR"); do
+        echo "| $file | $(shasum -a 256 "$OUTPUTDIR/$file" | awk '{print $1}' | tee "$OUTPUTDIR/$file.sha256") |"
```
micahhausler, line 102: What do you think about writing the full output so verifiers can run `sha256sum -c`? Adding the `sed` will trim out the absolute path so its just `<hash>  $file`

```suggestion
        echo "| $file | $(shasum -a 256 "$OUTPUTDIR/$file" | sed -e "s,$OUTPUTDIR,,"' | tee "$OUTPUTDIR/$file.sha256" | awk '{print $1}' ) |"
```

saschagrunert: Sure, added it as suggested :+1:

micahhausler (approved):
/lgtm

PR #724, opened 2021-03-01
fix StartedAt and FinishedAt of the container status 

Iceber:
Signed-off-by: Iceber Gu <wei.cai-nat@daocloud.io>

#### What type of PR is this?

/kind bug

Finished time is always `1754-08-31T06:49:24.128654848+08:05` in the run and create states, because `FinishedAt` is `-6795364578871345152`
```sh
[root@iceber-3 ~]# crictl inspect 587444fc34ec9
{
  "status": {
    "id": "587444fc34ec9846c163d0c74bbf985ac73662817a89e16b7328b955a03ef902",
    "metadata": {
      "attempt": 6,
      "name": "dx-arch-stolon-sentinel"
    },
    "state": "CONTAINER_RUNNING",
    "createdAt": "2021-03-01T16:45:23.390180141+08:00",
    "startedAt": "2021-03-01T16:45:23.544299436+08:00",
    "finishedAt": "1754-08-31T06:49:24.128654848+08:05",
    "exitCode": 0,
```
Fixed output in the running state:
```sh
[root@iceber-3 ~]# ./crictl inspect 587444fc34ec9
{
  "status": {
    "id": "587444fc34ec9846c163d0c74bbf985ac73662817a89e16b7328b955a03ef902",
    "metadata": {
      "attempt": 6,
      "name": "dx-arch-stolon-sentinel"
    },
    "state": "CONTAINER_RUNNING",
    "createdAt": "2021-03-01T16:45:23.390180141+08:00",
    "startedAt": "2021-03-01T16:45:23.544299436+08:00",
    "finishedAt": "0001-01-01T00:00:00Z",
    "exitCode": 0,
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -794,9 +794,22 @@ func marshalContainerStatus(cs *pb.ContainerStatus) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
 	jsonMap["createdAt"] = time.Unix(0, cs.CreatedAt).Format(time.RFC3339Nano)
-	jsonMap["startedAt"] = time.Unix(0, cs.StartedAt).Format(time.RFC3339Nano)
-	jsonMap["finishedAt"] = time.Unix(0, cs.FinishedAt).Format(time.RFC3339Nano)
+	var startedAt, finishedAt time.Time
+	if cs.State != pb.ContainerState_CONTAINER_CREATED {
+		// If container is not in the created state, we have tried and
+		// started the container. Set the startedAt.
+		startedAt = time.Unix(0, cs.StartedAt)
+	}
+	if cs.State == pb.ContainerState_CONTAINER_EXITED ||
+		(cs.State == pb.ContainerState_CONTAINER_UNKNOWN && cs.FinishedAt > 0) {
+		// If container is in the exit state, set the finishedAt.
+		// Or if container is in the unknown state and FinishedAt > 0, set the finishedAt
+		finishedAt = time.Unix(0, cs.FinishedAt)
+	}
+	jsonMap["startedAt"] = startedAt.Format(time.RFC3339Nano)
+	jsonMap["finishedAt"] = finishedAt.Format(time.RFC3339Nano)
 	return marshalMapInOrder(jsonMap, *cs)
 }
 


cmd/crictl/container.go
```diff
@@ -794,9 +794,20 @@ func marshalContainerStatus(cs *pb.ContainerStatus) (string, error) {
...
 	}
+
 	jsonMap["createdAt"] = time.Unix(0, cs.CreatedAt).Format(time.RFC3339Nano)
-	jsonMap["startedAt"] = time.Unix(0, cs.StartedAt).Format(time.RFC3339Nano)
-	jsonMap["finishedAt"] = time.Unix(0, cs.FinishedAt).Format(time.RFC3339Nano)
+	jsonMap["startedAt"], jsonMap["finishedAt"] = "", ""
+	if cs.State != pb.ContainerState_CONTAINER_CREATED {
+		// If container is not in the created state, we have tried and
+		// started the container. Set the startedAt.
+		jsonMap["startedAt"] = time.Unix(0, cs.FinishedAt).Format(time.RFC3339Nano)
```
saschagrunert, line 803:
```suggestion
		jsonMap["startedAt"] = time.Unix(0, cs.StartedAt).Format(time.RFC3339Nano)
```

saschagrunert (changes requested):

saschagrunert (commented):
Which container runtime did you use for the tests? I did a local test with CRI-O (master) and the resulting dates for a created container were:

```json
"startedAt": "1970-01-01T01:00:00+01:00",
"finishedAt": "1970-01-01T01:00:00+01:00",
```

I think we're good to move forward with that change. We could say that it's a breaking change but I would not expect someone to rely on the date parsing output if the container state does not match.

Iceber:
> Which container runtime did you use for the tests? I did a local test with CRI-O (master) and the resulting dates for a created container were:
> 
> ```json
> "startedAt": "1970-01-01T01:00:00+01:00",
> "finishedAt": "1970-01-01T01:00:00+01:00",
> ```
> 
> I think we're good to move forward with that change. We could say that it's a breaking change but I would not expect someone to rely on the date parsing output if the container state does not match.

I did the tests with the `dockershim`.
You get the `1970-01-01T01:00:00+01:00` because the started and finished timestamp is 0 (zero value)
https://github.com/cri-o/cri-o/blob/6ee8d68771c24a045aef1f5372eae377928a3530/server/container_status.go#L68-L84

You can use --debug to see the returned data

Iceber:
And kubernetes sometimes ignores startedAt and finishedAt when dealing with `CRI container status`, as seen [here](https://github.com/kubernetes/kubernetes/blob/793390e13be91d79150b57abad5710477ab96bd7/pkg/kubelet/kuberuntime/kuberuntime_container.go#L510-L522)

saschagrunert:
> And kubernetes sometimes ignores startedAt and finishedAt when dealing with `CRI container status`, as seen [here](https://github.com/kubernetes/kubernetes/blob/793390e13be91d79150b57abad5710477ab96bd7/pkg/kubelet/kuberuntime/kuberuntime_container.go#L510-L522)
> 
> I'm thinking about whether to show `""` or `0001-01-01T00:00:00Z`(`time.Time{}.Format(time.RFC3339Nano)`) for empty `startedAt` and `finishedAt`
> `0001-01-01T00:00:00Z` has the advantage that it can be parsed by `time.Parse` without returning errors. And `result.IsZero()` is true

In that case it would't be a "breaking change". I like that. :+1:

Iceber:
@saschagrunert  update to show `0001-01-01T00:00:00Z`.
PTAL, Thanks

saschagrunert (approved):
/lgtm
/hold for the tests

saschagrunert:
/hold cancel

the CI failures are not related to this change, I'll follow-up on those.

saschagrunert (approved):
/lgtm

PR #799, opened 2021-07-21
crictl: Adds support for updating resource limits for Windows Containers

claudiubelu:
#### What type of PR is this?

/kind feature
/sig windows

The fields supported in the WindowsContainerResources are:

- CpuCount: Number of CPUs available to the container. Default: 0 (not specified).
- CpuMaximum: Specifies the portion of processor cycles that this container can use as a percentage times 100.
- CpuShares: CPU shares (relative weight vs. other containers). Default: 0 (not specified).
- MemoryLimitInBytes: Memory limit in bytes. Default: 0 (not specified).

```release-note
Added support to update the resource limits to Windows Containers in crictl.
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	goruntime "runtime"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


feiskyer:
this looks good, but have you validated whether any runtimes (e.g. containerd) support this behavior?

claudiubelu:
Support is being added here: https://github.com/containerd/containerd/pull/5778

PR #864, opened 2021-12-17
[Do not merge] Try to fix windows CI

dcantah:
Trying out changes to fix the windows CI. Not sure why the windows shim is getting placed in /bin instead of $pwd/bin


dcantah:
Huzzah ok this worked

PR #865, opened 2021-12-21
Fix Containerd main branch CI for Windows

dcantah:
#### What type of PR is this?
/kind bug

Due to the containerd makefile using $(PWD)/bin to determine where to place the windows shim, it doesn't seem to like being run with powershell. I'm not sure if it's just not inherited or what the reason is, but instead of having the current dir replaced it simply would be an empty string. The result of this is that the windows shim would get placed in /bin on the machine instead of the bin in the current directory. 'make install' right afterwards would then fail to find the file. To remedy this, use bash.exe
as the shell to run the CI. This shell change should be harmless and is intended to get the CI working again for windows finally, but in Containerd using CURDIR instead of PWD may be the solution to this problem. There shouldn't be a need to revert this if a fix is put in place in Containerd

This change additionally adds a small bit of logic to only build the windows shim ourselves if we're running the CI for containerd < 1.6. In 1.6 (there's only beta versions out so far but) the makefile has support for building the shim itself. A simple 'make' or 'make binaries' invocation will build the shim and place it in the same location as containerd and ctr.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/containerd.yml
+++ b/.github/workflows/containerd.yml
@@ -102,10 +102,11 @@ jobs:
         working-directory: ${{ github.workspace }}/src/github.com/containerd/containerd
 
       - name: Install containerd on Windows
+        shell: bash
         if: startsWith(matrix.os, 'windows')
         run: |
-          make
-          make install
+          mingw32-make.exe 
+          mingw32-make.exe install
         working-directory: ${{ github.workspace }}/src/github.com/containerd/containerd
 
       - name: Install ${{matrix.runc}} runtime engine on Linux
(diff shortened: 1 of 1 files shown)


dcantah:
Change to fix the core issue in containerd as well, CURDIR seemed to work. https://github.com/containerd/containerd/pull/6394. Let me know if it'd be better to close this and I'll repurpose this (or open a new pr) so that it just contains the change to skip building the shim ourselves on 1.6+

feiskyer:
Thanks for the fix, LGTM.

Would you like to fix the dockershim building in a separate PR?

feiskyer:
Merged this manually because dockershim CI is also failing.

dcantah:
@feiskyer Thanks! And I'll have to check and see what the reason for those failures are, but currently on vacation so may be a bit delayed :/

PR #867, opened 2021-12-29
Fix cri-dockerd CI runs

dcantah:
#### What type of PR is this?
/kind bug

github.com/Mirantis/cri-dockerd has changed a couple things that were affecting the CI. The code has moved to a src/ subdirectory so running 'go install .' at the root of the repo fails on finding anything. The default addresses have also changed from /var/run/dockershim.sock (and //./pipe/dockershim) to /var/run/cri-dockerd.sock. A couple of flags also don't seem to exist anymore, namely --logtostderr and --v. All in all this change cds to the src subdirectory before trying to build/install anything, makes cri-dockerd launch with the old dockershim address, and gets rid of the now nonexistent flags.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/dockershim.yml
+++ b/.github/workflows/dockershim.yml
@@ -71,7 +71,7 @@ jobs:
           set -o pipefail
 
           git clone https://github.com/Mirantis/cri-dockerd $GOPATH/src/github.com/Mirantis/cri-dockerd
-          cd $GOPATH/src/github.com/Mirantis/cri-dockerd
+          cd $GOPATH/src/github.com/Mirantis/cri-dockerd/src
           go mod tidy
 
           # Need the flag for errors building on Windows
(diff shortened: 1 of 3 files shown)


dcantah:
Ok this looks like it's ready after testing this locally @feiskyer

dcantah:
We could go either way in this PR also for the address. Pass the new default address to critest under the -runtime-endpoint and -image-endpoint flags or do what we have currently and launch cri-dockerd with the old default address. Or change these https://github.com/kubernetes-sigs/cri-tools/blob/master/pkg/framework/test_context.go#L74-L78 but I'm not sure if this makes sense at the moment.

feiskyer (commented):
LGTM with setting the ep for the tests and thanks for the fix.

/lgtm
/approve

Issue #856, opened 2021-12-13
Release v1.23.0


thiguetta:
#### What would you like to be added:
k8s 1.23 was released on Dec 7th and still we don't have a cri-tools release for 1.23

saschagrunert:
I'd like to get the CI back before doing a release. Right now there seems to be two in flight:
- https://github.com/kubernetes-sigs/cri-tools/pull/858
- https://github.com/kubernetes-sigs/cri-tools/pull/853

We're not in a hurry since we would be still backwards compatible with Kubernetes. We also do not have CRI v1 support merged yet, so the changes between 1.22 and 1.23 seem marginal.

saschagrunert:
@feiskyer @mrunalp do we wanna switch to CRI v1 before releasing 1.23? This would be a breaking change.

Or do we wanna support both CRI versions in parallel? We could make it configurable, but it would increase the implementation effort.

hswong3i:
With Kubernetes usual N / N-1 support style, shall critool 1.23 support both Kubernetes 1.23 / 1.22?

If yes, does OCI only supported by Kubernetes 1.22, or still required for legacy Docker format support?

afbjorklund:
The packages are still on 1.19, so I think `crictl` has a longer life span.

The [homepage](https://github.com/kubernetes-sigs/cri-tools#current-status) says that it supports anything from k8s 1.16.x and up

afbjorklund:
If you use cri-o, it sets up 1.21 to be used.

```
 cri-tools |   1.21.0~2 | http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  Packages
```

If you use containerd, you get 1.19 instead

```
 cri-tools |  1.19.0-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
```

----

Assuming that one wants to use modern packaging, like apt.

And not fall back to having to install odd things from tarballs.

hswong3i:
For people who interested with latest up-to-date but non-official deb/rpm for Ubuntu/Debian/CentOS/RHEL/Fedora/openSUSE, I have a OBS repo for try: https://build.opensuse.org/project/show/home:alvistack

afbjorklund:
We are having big stability problems with OBS when it comes to installing cri-o, would be nice with something a bit more stable (and official!) for a tool that is now required by the default installer. Especially if multiple versions

hswong3i:
[...]

PR #871, opened 2022-01-12
Remove dockershim from crictl

### LINKED ISSUE #870: [1.24] Remove dockershim from crictl code and docs
`crictl` currently defaults to `dockershim.sock`. We should update the default in advance of 1.24. Additionally, remove mentions of dockershim from the repo:

https://github.com/kubernetes-sigs/cri-tools/search?q=dockershim.sock

shannonxtreme:
/kind feature
/kind deprecation

Removes dockershim from the crictl docs and main_unix / main_windows in preparation for the 1.24 deprecation of dockershim. 

- Remove `dockershim.sock` from default runtime endpoints for unix and windows
- Change `dockershim.sock` to `containerd.sock` in the default description for `--runtime-endpoints`.
- Remove `dockershim` from the crictl docs (including code examples in the docs)

This PR doesn't address `critest` mentions of `dockershim` as I'm not sure how to do that (but I'm happy to learn).

Fixes #870

```release-note
`dockershim` is no longer one of the runtime endpoints for `crictl`. 
```

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -60,13 +60,14 @@ This will
 - Run the benchmark tests using `ginkgo`
 - Output the test results to STDOUT
 
-critest connects to Unix: `unix:///var/run/containerd.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+critest connects to Unix: `unix:///run/containerd/containerd.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
 
 ## Additional options
 
 - `-ginkgo.focus`: Only run the tests that match the regular expression.
 - `-image-endpoint`: Set the endpoint of image service. Same with runtime-endpoint if not specified.
-- `-runtime-endpoint`: Set the endpoint of runtime service. Default to Unix: `unix:///var/run/containerd.sock` or Windows: `tcp://localhost:3735`.
+- `-runtime-endpoint`: Set the endpoint of runtime service. If not set, attempt
+  to search a list of known endpoints.
 
    **Note:** The default endpoints are deprecated. Always set the
    `--runtime-endpoint`. Performance might be affected as each default
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -46,18 +46,18 @@ This will
 - Run the tests using `ginkgo`
 - Output the test results to STDOUT
 
-critest connects to Unix: `unix:///var/run/containerd.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+For specific runtimes, the endpoint can be set by flags
+`--runtime-endpoint` and `--image-endpoint`.
 
 ## Additional options
 
 - `-ginkgo.focus`: Only run the tests that match the regular expression.
 - `-image-endpoint`: Set the endpoint of image service. Same with runtime-endpoint if not specified.
 - `-test-images-file`: Optional path to a YAML file containing references to custom container images to be used in tests.
-- `-runtime-endpoint`: Set the endpoint of runtime service. Default to
-  `unix:///var/run/containerd.sock` or Windows: `tcp://localhost:3735`.
+- `-runtime-endpoint`: Set the endpoint of runtime service. If not set, attempt to search a list of known endpoints.
 
-  **Note:** The default endpoints are deprecated. Always set the
-  `--runtime-endpoint`. Performance might be affected as each default
+  **Note:** The backup endpoints are deprecated. Always set the
+  `--runtime-endpoint`. Performance might be affected as each backup
   connection attempt takes n-seconds to complete before timing out and going
   to the next in sequence.
 
--- a/hack/run-dockershim-critest.sh
+++ b/hack/run-dockershim-critest.sh
@@ -16,9 +16,6 @@
 
 # Run critest with default dockershim.
 
-# Works only with https://github.com/Mirantis/cri-dockerd - other CRI
-# implementations aren't supported.
-
 set -o errexit
 set -o nounset
 set -o pipefail
--- a/cmd/crictl/main_unix.go
+++ b/cmd/crictl/main_unix.go
@@ -1,4 +1,4 @@
-// +build !windows
+//go:build !windows
 
 /*
 Copyright 2017 The Kubernetes Authors.
@@ -27,6 +27,6 @@ const (
 	defaultConfigPath = "/etc/crictl.yaml"
 )
 
-var defaultRuntimeEndpoints = []string{"unix:///run/containerd/containerd.sock", "unix:///run/crio/crio.sock", "unix:///var/run/cri-dockerd.sock"}
+var defaultRuntimeEndpoints = []string{"unix:///run/containerd/containerd.sock", "unix:///run/crio/crio.sock", "unix:///var/run/cri-dockerd.sock", "unix:///var/run/dockershim.sock"}
 
 var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
--- a/cmd/crictl/main_windows.go
+++ b/cmd/crictl/main_windows.go
@@ -1,5 +1,4 @@
 //go:build windows
-// +build windows
 
 /*
 Copyright 2017 The Kubernetes Authors.
@@ -24,7 +23,7 @@ import (
 	"path/filepath"
 )
 
-var defaultRuntimeEndpoints = []string{"npipe:////./pipe/containerd", "npipe:////./pipe/crio", "npipe:////./pipe/cri-dockerd"}
+var defaultRuntimeEndpoints = []string{"npipe:////./pipe/containerd", "npipe:////./pipe/crio", "npipe:////./pipe/cri-dockerd", "npipe:////./pipe/dockershim"}
 var defaultConfigPath string
 
 var shutdownSignals = []os.Signal{os.Interrupt}
--- a/docs/crictl.md
+++ b/docs/crictl.md
@@ -63,51 +63,65 @@ COMMANDS:
 - `completion`:         Output bash shell completion code
 - `help, h`:            Shows a list of commands or help for one command
 
-crictl by default connects on Unix to:
+## Connecting to endpoints
 
-- `unix:///run/containerd/containerd.sock` or
-- `unix:///run/crio/crio.sock` or
+You should specify an endpoint for `crictl` to use. The endpoint is based on your
+container runtime. You can set the endpoint in one of the following ways: 
+
+- Set global option flags `--runtime-endpoint` (`-r`) and `--image-endpoint`
+  (`-i`)
+- Set environment variables `CONTAINER_RUNTIME_ENDPOINT` and
+  `IMAGE_SERVICE_ENDPOINT`
+- Set the endpoint in the config file located at `/etc/crictl.yaml`
+
+If you don't set the endpoint, the following applies:
+
+- If the runtime endpoint is not set, `crictl` tries to connect using:
+  - containerd
+  - cri-o
+  - cri-dockerd
+  - dockershim (unsupported in Kubernetes 1.24)
+- If the image endpoint is not set, `crictl` uses the runtime endpoint setting
+
+`crictl` tries to connect to the following endpoints in order if you don't set one:
+
+**Unix**
+
+- `unix:///run/containerd/containerd.sock`
+- `unix:///run/crio/crio.sock`
 - `unix:///var/run/cri-dockerd.sock`
+- `unix:///var/run/dockershim.sock` (unsupported in Kubernetes 1.24)
 
-or on Windows to:
+**Windows**
 
-- `npipe:////./pipe/containerd` or
-- `npipe:////./pipe/crio` or
+- `npipe:////./pipe/containerd`
+- `npipe:////./pipe/crio`
 - `npipe:////./pipe/cri-dockerd`
+- `npipe:////./pipe/dockershim` (unsupported in Kubernetes 1.24)
 
 For other runtimes, use:
 
 - [frakti](https://github.com/kubernetes/frakti): `unix:///var/run/frakti.sock`
 
-The endpoint can be set in three ways:
+> Note: These backup endpoints are deprecated. You should **always** set a
+> runtime endpoint. If you don't, performance may be affected as each
+> backup connection attempt takes some time to complete before timing out
+> and going to the next in sequence.
 
-- By setting global option flags `--runtime-endpoint` (`-r`) and `--image-endpoint` (`-i`)
-- By setting environment variables `CONTAINER_RUNTIME_ENDPOINT` and `IMAGE_SERVICE_ENDPOINT`
-- By setting the endpoint in the config file `--config=/etc/crictl.yaml`
+You can view the current endpoint configuration in the `crictl` config file.
 
-If the endpoint is not set then it works as follows:
-
-- If the runtime endpoint is not set, `crictl` will by default try to connect using:
-  - containerd
-  - cri-o
-  - cri-dockerd
-- If the image endpoint is not set, `crictl` will by default use the runtime endpoint setting
-
-> Note: The default endpoints are now deprecated and the runtime endpoint should always be set instead.
-The performance maybe affected as each default connection attempt takes n-seconds to complete before timing out and going to the next in sequence.
-
-Unix:
+**Unix**
 
 ```sh
 $ cat /etc/crictl.yaml
-runtime-endpoint: unix:///var/run/containerd.sock
-image-endpoint: unix:///var/run/containerd.sock
+runtime-endpoint: unix:///run/containerd/containerd.sock
+image-endpoint: unix:///run/containerd/containerd.sock
 timeout: 2
 debug: true
 pull-image-on-create: false
 ```
 
-Windows:
+**Windows**
 
 ```cmd
 C:\> type %USERPROFILE%\.crictl\crictl.yaml


dims:
looks like all the bases are covered. thanks @shannonxtreme 

https://cs.k8s.io/?q=dockershim.sock&i=nope&files=&excludeFiles=&repos=kubernetes-sigs/cri-tools

/assign @adisky 

one thing though ... for hack/run-dockershim-critest.sh we could add some comments that i can only work with https://github.com/Mirantis/cri-dockerd and not with any other CRI implementation. (either in this PR or in a follow up, i don't mind)

SergeyKanzhelev:
Do we want `crictl` to be able to talk to earlier versions of k8s clusters while they are still supported? Once crictl is moved to use v1 of CRI API only, this will be totally reasonable PR. Before that, I'd suggest we keep it.

SergeyKanzhelev:
> Do we want `crictl` to be able to talk to earlier versions of k8s clusters while they are still supported?

One thing is that the matrix is not up to date: https://github.com/kubernetes-sigs/cri-tools#current-status

adisky:
> Do we want `crictl` to be able to talk to earlier versions of k8s clusters while they are still supported? Once crictl is moved to use v1 of CRI API only, this will be totally reasonable PR. Before that, I'd suggest we keep it.

@SergeyKanzhelev doesn't users expected to use matching version of crictl & k8s, if K8s 1.24 removed dockershim crictl 1.24 also expected to do the same?possiblt we should release a 1.23 tag for crictl then merge this PR?

Also this isn't essentially removing any support, only removing dockershim from default endpoints users can still configure `unix:///var/run/dockershim.sock`  manually

SergeyKanzhelev:
> @SergeyKanzhelev doesn't users expected to use matching version of crictl & k8s

as a general statement, this wouldn't be ideal. `crictl` may be used alongside the containerd. For instance, it is installed on COS. And it works independently on the version of k8s installed on that COS image.

Since the `dockershim.sock` is only exposed by k8s itself, this statement has more sense. Also, defaults are deprecated and customers are encouraged to specify the socket explicitly (#599). So even if we will break customer scenario here, where the fallback to default is expected, it will not be completely unexpected. It also will be a good notification for customers to migrate :-).

docs/benchmark.md
```diff
@@ -38,12 +38,12 @@ This will
...
 
-critest connects to Unix: `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+critest connects to Unix: `unix:///var/run/containerd.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
 
 ## Additional options
 
 - `-ginkgo.focus`: Only run the tests that match the regular expression.
 - `-image-endpoint`: Set the endpoint of image service. Same with runtime-endpoint if not specified.
-- `-runtime-endpoint`: Set the endpoint of runtime service. Default to Unix: `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735`.
+- `-runtime-endpoint`: Set the endpoint of runtime service. Default to Unix: `unix:///var/run/containerd.sock` or Windows: `tcp://localhost:3735`.
```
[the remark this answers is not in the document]
shannonxtreme: It attempts to connect to each default endpoint in order as well doesn't it? So containerd, cri-o, cri-dockerd?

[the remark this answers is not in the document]
SergeyKanzhelev: that is correct: https://github.com/kubernetes-sigs/cri-tools/blob/89f77d831c900431a85d42c28fe81dbf2f4d3f9e/cmd/crictl/main.go#L108

saschagrunert:
Let's discuss the release related bits in https://github.com/kubernetes-sigs/cri-tools/issues/856

afbjorklund:
We would have hate having to use different versions of crictl in minikube, for different versions of kubernetes.

afbjorklund:
Note that the packages are still installing crictl 1.19.0. (never mind 1.23.0)

```
 cri-tools |  1.19.0-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
 cri-tools |  1.13.0-01 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
 cri-tools |  1.13.0-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
 cri-tools |  1.12.0-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
 cri-tools |  1.11.1-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
 cri-tools |  1.11.0-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
 cri-tools | 1.0.0-beta.1-00 | https://apt.kubernetes.io kubernetes-xenial/main amd64 Packages
```

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

shannonxtreme:
Since the defaults are deprecated anyway, isn't this loss of functionality
more of an incentive to move off relying on just the defaults, and instead
specifying explicit endpoints?

Idk, someone from sig-node probably knows better than me

afbjorklund:
> Since the defaults are deprecated anyway, isn't this loss of functionality more of an incentive to move off relying on just the defaults, and instead specifying explicit endpoints?

But then you could remove all endpoints, instead of just one.

shannonxtreme:
I'm not sure why that hasn't happened. Maybe a deprecation policy thing?

I guess a middle ground could be to move dockershim.sock to last on the
list so that it isn't the first endpoint tried by default

afbjorklund:
> as a general statement, this wouldn't be ideal. `crictl` may be used alongside the containerd. For instance, it is installed on COS. And it works independently on the version of k8s installed on that COS image.

Same thing for minikube, currently using a random version 1.21.0 for the crictl installation

docs/benchmark.md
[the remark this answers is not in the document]
afbjorklund: there is no default: https://github.com/kubernetes-sigs/cri-tools/issues/597

afbjorklund:
> It also will be a good notification for customers to migrate :-).

They don't have to migrate, they just have to start using CRI

shannonxtreme:
Is it that the defaults currently do not work? I thought the case is that
the defaults are deprecated and will be removed at some point, but still
work for now

afbjorklund:
> Is it that the defaults currently do not work?

The project doesn't dare to pick a new favorite, so recommends using any container runtime that follows the CRI

And if crictl needs to try all the possible locations, that introduces a latency when checking the various sockets...

shannonxtreme:
@SergeyKanzhelev and @afbjorklund in this case, should we leave dockershim.sock in the list of fallback options? If we keep it, we should mention in the docs that the endpoint won't work in 1.24+. The main changes for this PR would then just be to update the language around "default" behaviour in the warnings. 

WDYT?

afbjorklund:
The question is: can you use crictl 1.24 to control a cluster running kubernetes 1.23 ?

We currently install crictl as part of the OS (together with the container runtime).
So it is the same crictl that is being used for **all** versions of Kubernetes, later on.

We do set the /etc/crictl.yaml (as instructed), so if it is "only" the list of default sockets...

But once you start changing the CRI version, this rule will become more important.
Unless it is always backward compatible, so that the new tool can talk to old clusters ?

afbjorklund:
And like I mentioned elsewhere, currently crictl 1.19+ is being packaged and required.

So there is no mapping between crictl version and kubernetes version, in requirements.

shannonxtreme:
@afbjorklund I'm honestly not sure about the versioning thing, sorry. I'd assume if we left the list of fallback endpoints as-is, the behaviour wouldn't change, but I'm not positive. 

Maybe someone from SIG node can confirm, or @SergeyKanzhelev.

adisky:
I still think this is a nice workaround
```
I guess a middle ground could be to move dockershim.sock to last on the
list so that it isn't the first endpoint tried by default

**Comment by shannonxtreme** [comment]
I made a few changes @afbjorklund and @SergeyKanzhelev:

* Add dockershim.sock as the last item in the list of "default" endpoints
* In the docs, position setting the endpoint manually as the recommended approach
* For the description of default endpoints in docs, add "unsupported in Kubernetes 1.25" to dockershim.sock

This should keep crictl working for previous versions of Kubernetes. We should remove the default endpoints at some point, preferable in a version release of crictl. 

I'd like to get this change merged. This no longer breaks users on previous versions of K8s, makes the removal of dockershim in 1.25 clear, and doesn't require crictl to be versioned to work.

**Comment by afbjorklund** [comment]
> makes the removal of dockershim in 1.25 clear,

was it moved **again** ? It was supposed to be 1.24

**Comment by shannonxtreme** [comment]
Lol it was not moved again, I just apparently had a brain fart. Thank you. That's also why I am trying to change the wording to be more "the fallback (or backup) endpoints are deprecated".

**Comment by afbjorklund** [comment]
> IDK why VS Code is adding the extra line 1 in main_unix.go and main_windows.go (//go:build !windows). Will it affect the file? I can't save the file without it being auto added

If your gofmt is "too new", it will update the line to the new syntax (you can remove the old syntax)

**Comment by afbjorklund** [comment]
> "the fallback (or backup) endpoints are deprecated"

It's deprecated alright, just not entirely removed yet.

**Comment by shannonxtreme** [comment]
@afbjorklund thanks for the clarity. I opened #893 as a suggestion to remove these fallbacks with minimal impact.

**Comment by afbjorklund** [comment]
We enabled a feature to minikube, to start the extra background service (or socket) that is required with the new setup.

For containerd and cri-o, nothing needs starting. For docker, nothing needed started with dockershim but now it does:

`sudo systemctl start cri-docker.service`

Ideally cri-dockerd should be "started" with socket activation of `/var/run/cri-dockerd.sock`, but there are no packages yet.

So the user needs to remember to start it, after starting "docker.service" (which is maybe more obvious that is required)

File: docs/benchmark.md [hunk 2]
```diff
@@ -38,12 +38,19 @@ This will
...
-critest connects to Unix: `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+critest connects to Unix: `unix:///run/containerd/containerd.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
 
 ## Additional options
 
 - `-ginkgo.focus`: Only run the tests that match the regular expression.
 - `-image-endpoint`: Set the endpoint of image service. Same with runtime-endpoint if not specified.
-- `-runtime-endpoint`: Set the endpoint of runtime service. Default to Unix: `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735`.
+- `-runtime-endpoint`: Set the endpoint of runtime service. Default to Unix:
+  `unix:///run/containerd/containerd.sock` or Windows: `tcp://localhost:3735`.
```
afbjorklund, line 48: remove this info (about containerd and docker tcp). Should be something like:

`-runtime-endpoint`: Set the endpoint of runtime service. Default to searching a list of known endpoints.

docs/validation.md
```diff
@@ -46,14 +46,24 @@ This will
 - Run the tests using `ginkgo`
 - Output the test results to STDOUT
 
-critest connects to Unix: `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+For specific runtimes, the endpoint can be set by flags
+`--runtime-endpoint` and `--image-endpoint`. critest connects to Unix:
```
afbjorklund, line 50: Again, this "default endpoint" is probably not true. Also critest should search all of them.

docs/validation.md
```diff
@@ -46,14 +46,24 @@ This will
...
+`tcp://localhost:3735` if you don't specify a runtime endpoint. 
 
 ## Additional options
 
 - `-ginkgo.focus`: Only run the tests that match the regular expression.
 - `-image-endpoint`: Set the endpoint of image service. Same with runtime-endpoint if not specified.
 - `-test-images-file`: Optional path to a YAML file containing references to custom container images to be used in tests.
-- `-runtime-endpoint`: Set the endpoint of runtime service. Default to `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735`.
+- `-runtime-endpoint`: Set the endpoint of runtime service. If not set, attempt
+  to connect to `unix:///run/containerd/containerd.sock` or Windows: `tcp://localhost:3735`.
```
afbjorklund, line 60: And again.

hack/run-dockershim-critest.sh
```diff
@@ -16,6 +16,9 @@
 
 # Run critest with default dockershim.
 
+# Works only with https://github.com/Mirantis/cri-dockerd - other CRI
```
afbjorklund, line 19: This information is incorrect, it still sets the endpoint to dockershim (not cri-dockerd) below.

docs/validation.md
shannonxtreme: Not sure what the correct one is here then. Revert to dockershim.sock?

shannonxtreme:
@afbjorklund I changed most of what you pointed out except the info about the critest default endpoint. LMK if that needs to be reverted or not. Also the info about starting cri-dockerd is not in scope for this PR. Not sure where it needs to go in the cri-tools docs (if it does at all).

afbjorklund:
> Also the info about starting cri-dockerd is not in scope for this PR. 

Makes sense, it was just that I thought it was still confusion about the location - but it turned out that it wasn't started...

docs/validation.md
afbjorklund: Can't you just leave it out ?

For specific runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.

saschagrunert:
Have we reached any consensus with this one?

afbjorklund:
Do you want to make crictl part of the release ?

Or is it OK to use the same system installation for all versions of k8s, like both 1.23 and 1.24 ?

saschagrunert:
> Or is it OK to use the same system installation for all versions of k8s, like both 1.23 and 1.24 ?

I'd prefer that. But we can still introduce breaking changes if we document them accordingly. For example we will have to remove support for CRI v1alpha2 when we vendor in Kubernetes which also does not support the conversion any more.

afbjorklund:
I don't see any version requirements on cri-tools and cni-plugins, and they seem to frequently be lagging behind ?

mikebrow (approved):
LGTM

when we remove the default loop check ..would be good to add a check on not provided and ask if they want to attempt to connect.. check known defaults... ask them if they want to save one in the config..

afbjorklund:
Currently (v1.24.0), it is just failing. It never gets around to trying `unix:///var/run/cri-dockerd.sock`

```
WARN[0000] runtime connect using default endpoints: [unix:///var/run/dockershim.sock unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]. As the default settings are now deprecated, you should set the endpoint instead. 
ERRO[0000] unable to determine runtime API version: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix /var/run/dockershim.sock: connect: no such file or directory" 
E0510 08:00:35.184124   12046 remote_runtime.go:925] "Status from runtime service failed" err="rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService"
FATA[0000] getting status of runtime: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService 
```

feiskyer:
make a rebase again?

mikebrow:
> @mikebrow
> 
> Currently (v1.24.0), it is just failing. It never gets around to trying `unix:///var/run/cri-dockerd.sock`
> 
> ```
> WARN[0000] runtime connect using default endpoints: [unix:///var/run/dockershim.sock unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]. As the default settings are now deprecated, you should set the endpoint instead. 
> ERRO[0000] unable to determine runtime API version: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix /var/run/dockershim.sock: connect: no such file or directory" 
> E0510 08:00:35.184124   12046 remote_runtime.go:925] "Status from runtime service failed" err="rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService"
> FATA[0000] getting status of runtime: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService 
> ```
> 
> It works fine with the `-r` parameter. I think it is done that way in `kubeadm`, instead of relying on CRI config.
> 
> ```
> $ crictl -r unix:///run/cri-dockerd.sock version
> Version:  0.1.0
> RuntimeName:  docker
> RuntimeVersion:  20.10.12
> RuntimeApiVersion:  1.41.0
> ```

ok.. thx for the heads up.. need some more debug so we can see which container runtime on which socket doesn't have a matching runtime.v1alpha2.RuntimeService...

afbjorklund:
It was either 1.4 or 1.5, should be easy to reproduce

mikebrow:
> It was either 1.4 or 1.5 (whatever docker used), should be easy to reproduce
> 
> There were some weird ghost reports of it "failing" even with containerd 1.6, since the error message can be misleading if the daemon is not answering/running for some reason...
> 
> By providing it explicitly, the error is more clear.
> 
> And by using configuration, it is less cluttered than having to lug that `-r` parameter around. So I think that kubeadm should require /etc/crictl.yaml use well as `crictl`

If containerd is installed but you have disabled the cri plugin on purpose (standard for docker install's version of the containerd config.toml) the containerd sock will not have any cri services to serve and crictl will see that as a failure.  It shouldn't stop it's loop rather it should continue on reporting that the containerd sock does not have cri enabled.

mikebrow:
The error here https://github.com/kubernetes-sigs/cri-tools/blob/master/cmd/crictl/main.go#L82 isn't very useful when the err returned does not mention the endpoint that failed...

mikebrow:
IMO the loop thing should be a oh .. you don't have a container runtime configured in your crictl config and you have not set one at the command line or in the ENV.. here are your options we've found at the typical default configurations.. which would you like to set as your default in the config...: grab stdin wait for response for 30sec.. or fail to prompt

afbjorklund:
I tried again, with the latest minikube beta - but with k8s version 1.24.0 and crictl version 1.24.0:

```
docker@minikube:~$ sudo rm /etc/crictl.yaml 
docker@minikube:~$ sudo crictl version
WARN[0000] runtime connect using default endpoints: [unix:///var/run/dockershim.sock unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock]. As the default settings are now deprecated, you should set the endpoint instead. 
ERRO[0002] connect endpoint 'unix:///var/run/dockershim.sock', make sure you are running as root and the endpoint has been started: context deadline exceeded 
FATA[0002] getting the runtime version: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService 
```

shannonxtreme:
Ok so I rebased, but I'd really like someone to take another look at these changes to make sure this is right

mikebrow:
> I tried again, with the latest minikube beta - but with k8s version 1.24.0 and crictl version 1.24.0:
> 
> `minikube start --kubernetes-version=v1.24.0`
> 
> ```
> docker@minikube:~$ sudo rm /etc/crictl.yaml 
> docker@minikube:~$ sudo ./crictl --version
> crictl version 1.24.0
> docker@minikube:~$ sudo ./crictl version
> WARN[0000] runtime connect using default endpoints: [unix:///var/run/dockershim.sock unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]. As the default settings are now deprecated, you should set the endpoint instead. 
> ERRO[0000] unable to determine runtime API version: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix /var/run/dockershim.sock: connect: no such file or directory" 
> E0525 18:30:45.757607   51352 remote_runtime.go:168] "Version from runtime service failed" err="rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService"
> FATA[0000] getting the runtime version: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService 
> docker@minikube:~$ containerd --version
> containerd containerd.io 1.6.4 212e8b6fa2f44b9c21b2798135fc6fb7c53efc16
> docker@minikube:~$ dockerd --version
> Docker version 20.10.15, build 4433bf6
> docker@minikube:~$ cri-dockerd --version
> cri-dockerd 0.2.0 (a4d1895)
> docker@minikube:~$ sudo ./crictl -r unix:///var/run/cri-dockerd.sock version
> Version:  0.1.0
> RuntimeName:  docker
> RuntimeVersion:  20.10.15
> RuntimeApiVersion:  1.41.0
> ```
> 
> So we always set up the /etc/crictl.yaml config.

you may have tried it again but did you fix the broken config.toml yet?

afbjorklund:
It using docker, not containerd

afbjorklund:
Here is the output from the version in this PR:

```console
docker@minikube:~$ sudo ./crictl --version
crictl version 1.24.0-38-g43477844
docker@minikube:~$ sudo ./crictl --debug version
DEBU[0000] get runtime connection                       
WARN[0000] runtime connect using default endpoints: [unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock unix:///var/run/dockershim.sock]. As the default settings are now deprecated, you should set the endpoint instead. 
DEBU[0000] Note that performance maybe affected as each default connection attempt takes n-seconds to complete before timing out and going to the next in sequence. 
DEBU[0000] Connect using endpoint "unix:///run/containerd/containerd.sock" with "2s" timeout 
DEBU[0000] Connected successfully using endpoint: unix:///run/containerd/containerd.sock 
DEBU[0000] VersionRequest: &VersionRequest{Version:v1,} 
E0526 10:31:27.234639     568 remote_runtime.go:168] "Version from runtime service failed" err="rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService"
DEBU[0000] VersionResponse: nil                         
FATA[0000] getting the runtime version: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService 
docker@minikube:~$ 
```

mikebrow:
> It never tries the other endpoints.

correct if docker's edit of the containerd config has disabled the cri services the containerd sock won't serve cri services

> If I enable the cri plugin, then it will think that it should use the containerd runtime (instead of the docker runtime)

this is your view.. makes sense to you.. may not make sense to people who are not using a dockershim version of kubelet on their workernode or development environment

crictl has a config option and a cli option to let users specify which container runtime they want to use

for ages the defaulting sequence list in crictl was dockershim first and the list was deprecated .. I believe with the intention of removing it and letting people choose manually via config or cli.. given the number of options available for choosing it makes sense to let users choose which container runtime they want. 

choices: either delete the default code and force manual config, or keep it as it is with a fix to continue to the next when the sock is there but docker has disabled the cri api, or get docker to stop disabling the cri api in containerd, and/or give crictl users a list to choose from and ask them which one they want..

> So then it won't match the CRI configuration from kubeadm, which will confuse things later.

why does kubeadm require containerd not be configured correctly for working with kubeadm? I believe kubeadm also works with containerd/cri-o?

> `/var/lib/kubelet/kubeadm-flags.env:KUBELET_KUBEADM_ARGS="--container-runtime=remote --container-runtime-endpoint=unix:///var/run/cri-dockerd.sock --node-ip=192.168.49.2 --pod-infra-container-image=k8s.gcr.io/pause:3.7"`
> `Found multiple CRI endpoints on the host. Please define which one do you wish to use by setting the 'criSocket' field in the kubeadm configuration file: unix:///var/run/containerd/containerd.sock, unix:///var/run/cri-dockerd.sock`
> So I think it is better to leave the default docker configuration for containerd, and configure crictl ?

it is better to have containerd, and the other container runtimes, configured correctly for their clients, and configure crictl for your specific use case.  There are a great number of clients of the container runtimes... not just this particular kubeadm use case.

> /etc/crictl.yaml
> 
> ```
> runtime-endpoint: unix:///var/run/cri-dockerd.sock
> ```

yup and 
```
crictl config --get runtime-endpoint 
unix:///run/containerd/containerd.sock
```
or 
```
crictl config --set runtime-endpoint unix:///var/run/cri-dockerd.sock
```

afbjorklund:
> why does kubeadm require containerd not be configured correctly for working with kubeadm?

It uses a separate flag / config item for kubeadm, and then uses `crictl -r` with it for all the calls.

PR #873, opened 2022-01-15
Use same grpc max message size as Kubelet

### LINKED ISSUE #872: crictl pods fails when response message size is big
#### What happened:

`crictl pods` returns failure when the response message size exceeds 4M.

`crictl pods` continues to work with big response message.

Create lots of pods through crictl then run `crictl pods`. The command will return unsuccessful code with error message "listing pod sandboxes: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4564414 vs. 4194304)"

containerd & kubelet had increased the max message size to 16M for their grpc client.

https://github.com/containerd/containerd/blob/main/defaults/defaults.go#L22
https://github.com/kubernetes/kubernetes/blob/7bde4baac48d702aabb03123a342d3c11323c4ef/pkg/kubelet/cri/remote/utils.go#L29

linxiulei:
Fixes #872

#### What type of PR is this?

/kind bug

Fixes #872

Fixes #872

```release-note
NONE
```

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -38,6 +38,8 @@ import (
 
 const (
 	defaultTimeout = 2 * time.Second
+	// use same message size as cri remote client in kubelet.
+	maxMsgSize = 1024 * 1024 * 16
 )
 
 var (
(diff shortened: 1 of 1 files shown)


feiskyer (commented):
thanks for the improvements, Triggering github action CIs ...

saschagrunert (approved):
/lgtm

PR #857, opened 2021-12-13
add tests for ListContainerStats

minbaev:
Signed-off-by: Alexander Minbaev <alexander.minbaev@ibm.com>

#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
PR adds tests for `ListContainerStats` RPC
Adds tests covering the RPC behavior after this change:
 https://github.com/containerd/containerd/pull/6373

#### Special notes for your reviewer:
None

#### Does this PR introduce a user-facing change?
None

PR #880, opened 2022-01-21
use commit ref

minbaev:
##What this PR does / why we need it:

updates containerd version to commit ref this is needed to make
#857
pass


minbaev:
closing this pr as the commit ref was added to #857 for testing

Issue #883, opened 2022-01-26
crictl must work with containerd 1.5 - allow the fallback logic to CRI API v1alpha2


SergeyKanzhelev:
This is a follow up from https://github.com/kubernetes-sigs/cri-tools/pull/878#issuecomment-1022493135

Containerd 1.5 only implements CRI API `v1alpha2` and cri tools must support the same runtimes k8s supports. So there should be a fallback logic to `v1alpha2`.

SergeyKanzhelev:
@saschagrunert what is the versioning story of CRI tools for CRI-O? How many versions back customers will expect to be supported?

SergeyKanzhelev:
cc: @qiutongs

saschagrunert:
> @saschagrunert what is the versioning story of CRI tools for CRI-O? How many versions back customers will expect to be supported?

Since CRI-O versionizes together with Kubernetes, we would have any issue with a breaking change here. We also support CRI v1 since a couple of releases (v1.21 I guess), so we're already good to go.

saschagrunert:
One way to approach this would be to switch to the internal runtime and image services completely:

https://github.com/kubernetes-sigs/cri-tools/blob/8553b616d79b7753f4eda5fb9144cd8ca9eb9764/cmd/crictl/main.go#L135-L137

This service already features the conversion of the types, but it would mean that we bound the kubelet code tighter to crictl.

afbjorklund:
Same problem with Docker (cri-dockerd), I guess it deserves a special issue.

```console
docker@minikube:~$ crictl --version
crictl version v1.21.0
docker@minikube:~$ crictl version
Version:  0.1.0
RuntimeName:  docker
RuntimeVersion:  20.10.12
RuntimeApiVersion:  1.41.0
docker@minikube:~$ cri-dockerd --version
cri-dockerd 0.2.0 (a4d1895)
docker@minikube:~$ crictl info
{
  "status": {
    "conditions": [
      {
        "type": "RuntimeReady",
        "status": true,
        "reason": "",
        "message": ""
      },
      {
        "type": "NetworkReady",
        "status": true,
        "reason": "",
        "message": ""
      }
    ]
  }
}
```

When upgrading cri-tools to latest/greatest (i.e. 1.24 pre-release), it no longer works:

```console
docker@minikube:~$ ./crictl-1.23.0 --version
crictl version v1.23.0
docker@minikube:~$ ./crictl-1.23.0 info
{
  "status": {
[...]

PR #884, opened 2022-01-27
Add backwards compatiblity to `Version` and `ListContainerStats` RPCs

saschagrunert:
#### What type of PR is this?

/kind feature

This allows backwards compatiblity with previous runtimes not supporting
CRI v1.

Refers to https://github.com/kubernetes-sigs/cri-tools/issues/883
None

```release-note
None
```

### DIFF
--- a/cmd/crictl/version.go
+++ b/cmd/crictl/version.go
@@ -22,23 +22,20 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
-	"golang.org/x/net/context"
+	internalapi "k8s.io/cri-api/pkg/apis"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
-const (
-	criClientVersion = "v1"
-)
+const criClientVersion = "v1"
 
 var runtimeVersionCommand = &cli.Command{
 	Name:  "version",
 	Usage: "Display runtime version information",
 	Action: func(context *cli.Context) error {
-		runtimeClient, runtimeConn, err := getRuntimeClient(context)
+		runtimeClient, err := getRuntimeService(context)
 		if err != nil {
 			return err
 		}
-		defer closeConnection(context, runtimeConn)
 		err = Version(runtimeClient, criClientVersion)
 		if err != nil {
 			return errors.Wrap(err, "getting the runtime version")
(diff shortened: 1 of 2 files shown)


saschagrunert:
Let's see what the CI thinks.

saschagrunert:
@feiskyer @SergeyKanzhelev @Random-Liu PTAL

cmd/crictl/version.go
```diff
@@ -48,10 +45,10 @@ var runtimeVersionCommand = &cli.Command{
 }
 
 // Version sends a VersionRequest to the server, and parses the returned VersionResponse.
-func Version(client pb.RuntimeServiceClient, version string) error {
+func Version(client internalapi.RuntimeService, version string) error {
 	request := &pb.VersionRequest{Version: version}
```
SergeyKanzhelev, line 49: I wonder what is the history of this field. Both Containerd and Cri-O seems to be simply ignoring this value

cmd/crictl/version.go
```diff
@@ -22,23 +22,20 @@ import (
...
 	"github.com/urfave/cli/v2"
-	"golang.org/x/net/context"
+	internalapi "k8s.io/cri-api/pkg/apis"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
-const (
-	criClientVersion = "v1"
-)
+const criClientVersion = "v1"
```
SergeyKanzhelev, line 29: since we are using internlapi, maybe use `internalApi.CRIVersionV1`?

This const is meaningless, only recommending for consistency

SergeyKanzhelev (approved):
/lgtm

SergeyKanzhelev:
sorry if anybody else wanted to review this, i didn't realize this will merge with lgtm.

I think this is the right approach for back compat

cmd/crictl/version.go
saschagrunert: Yes, this makes sense I'll fix it in the next PR.

cmd/crictl/version.go
afbjorklund: I think it was for the Docker API version, no ?

```go
        typedVersion, err := r.runtimeClient.Version(ctx, &runtimeapi.VersionRequest{
                Version: apiVersion,
        })
```

```
docker@minikube:~$ sudo crictl version
Version:  0.1.0
RuntimeName:  docker
RuntimeVersion:  20.10.12
RuntimeApiVersion:  1.41.0
```

afbjorklund: @SergeyKanzhelev : apparently the apiVersion was used to make the Docker API Version semver-compatible:

```go
       // Docker API version (e.g., 1.23) is not semver compatible. Add a ".0"
       // suffix to remedy this.
       apiVersion := fmt.Sprintf("%s.0", v.APIVersion)
```

https://github.com/kubernetes/kubernetes/commit/13b23c51cb8539ec83e17474618b5ecb34f5bf28

Then it was moved out to a separate track, but of course the field remained in the API (not used for Docker either).

https://github.com/kubernetes/kubernetes/commit/d8e29e782f4087f810ae543d038772c503ce2d4b

PR #895, opened 2022-02-22
Type cast error messages to `string`

saschagrunert:
#### What type of PR is this?

/kind bug

We should force string types for those error messages otherwise they're
not readable by the ginkgo output when running critest.

None
None

```release-note
Fixed critest `ExecSync` output to print strings when `stderr` should be empty.
```


SergeyKanzhelev (approved):
/lgtm

there are also couple placed stderr is being formatted into string., but `%q` qualifier used there should be OK.

PR #906, opened 2022-03-22
Add image signature verification support on pull

saschagrunert:
#### What type of PR is this?

/kind feature

The `crictl pull` subcommand now supports the `--verify-signature, -s` command to verify cosign signatures, which will be shipped for the official  Kubernetes images from v1.24.0. This allows other tools to verify the signatures on the fly before pulling the image.

Please note that providing the flag will not fail if no signature is present, only if it's invalid.

None
Refers to https://github.com/kubernetes/release/issues/2383

```release-note
Added `crictl pull --verify-signature, -s` subcommand to verify cosign signatures of container images.
Please note that providing the flag will not fail if no signature is present, only if it's invalid.
```


saschagrunert:
@feiskyer @mrunalp @mikebrow  what do you think, do we wanna take such a feature in?

feiskyer:
+1 to have this

saschagrunert:
Rebased, PTAL @feiskyer :)

mikebrow:
> @feiskyer @mrunalp @mikebrow what do you think, do we wanna take such a feature in?

a few thoughts/questions.. apologies for not being up to date with your progress on this

1) +1 to have this done somewhere in the stack, just not sure it's the right fit for crictl

2) This project scope change probably needs approval of some sort from the sponsor: https://github.com/kubernetes-sigs/cri-tools#what-is-the-scope-of-this-project

3) Doing the check in crictl is no guarantee the runtime will use the same resolve path or get the same image unless we at least do some sha verification between what is being pre-checked and the image pulled. 

4) On error should the image be removed? Should it have been unpacked if it's invalid?

5) In my mind this was either going to be done in a higher level client context (kubelet image manager or admission controller) and/or through a set of policies passed down through the CRI and possibly handled through an extension in the distribution api enabled by with the registries/artifact providers to reduce traffic overload.

haircommander:
FWIW now that this issue's gone stale: I agree this seems like more of something one would configure in the runtime, unless the CRI becomes signature aware

Issue #925, opened 2022-05-06
ginkgo version aligning


mikebrow:
We, containerd, are having problems moving up to v1.24.0 and post v1.24.0 critest with both parallel on and a skip and/or focus applied.

History here: https://github.com/containerd/containerd/pull/6894

Noting also that k/k is on a different version of ginkgo and ginkgo has the requirement of the installed version matching the library. This may cause some versioning issues when doing both critest and e2e buckets.  Let's discuss.

Cheers, Mike

saschagrunert:
> We, containerd, are having problems moving up to v1.24.0 and post v1.24.0 critest with both parallel on and a skip and/or focus applied.

Is this something which has to be fixed in ginkgo?

> Noting also that k/k is on a different version of ginkgo and ginkgo has the requirement of the installed version matching the library. This may cause some versioning issues when doing both critest and e2e buckets. Let's discuss.

There are also efforts of moving k/k to ginkgo v2: https://github.com/kubernetes/kubernetes/pull/109111

mikebrow:
missed this question @saschagrunert .. fu wei was taking a look, see pr ^

saschagrunert:
@mikebrow if #930 got merged, would you need a new release of cri-tools?

PR #930, opened 2022-05-11
cmd/critest: fix empty ginkgo flag's value issue

fuweid:
#### What type of PR is this?

/kind bug

In commit 53ad8bb7f97e1b1d1c0c0634e43a3c2b8b07b718 [1], the ginkgo uses
flags.StringVar for some flags, for instance, the --skip or --focus.

But in the following release of ginkgo(>=v1.15.0 [2]), the ginkgo starts to
use string slice instead of the StringVar, which means we can't get the
user's input by the flags.Value.String() [3]. The String() always returns
empty value.

Based on that, we should use ginkgo API to generate flags instead of
manually handler in cri-test.

[1]: https://github.com/kubernetes-sigs/cri-tools/blob/53ad8bb7f97e1b1d1c0c0634e43a3c2b8b07b718/vendor/github.com/onsi/ginkgo/config/config.go#L79
[2]: https://github.com/onsi/ginkgo/blob/v1.15.0/config/config.go#L84
[3]: https://github.com/onsi/ginkgo/blob/v1.15.0/config/config.go#L68-L71

Fixes: #925

```release-note

```

### DIFF
--- a/cmd/critest/cri_test.go
+++ b/cmd/critest/cri_test.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/ginkgo/v2/reporters"
+	ginkgotypes "github.com/onsi/ginkgo/v2/types"
 	"github.com/onsi/gomega"
 
 	_ "github.com/kubernetes-sigs/cri-tools/pkg/benchmark"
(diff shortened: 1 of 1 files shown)


fuweid:
My local test result is like

```
➜  cri-tools git:(support-string-slice-var) sudo build/bin/critest -parallel $(nproc) -ginkgo.skip="Image"            
critest version: 1.24.0-7-g651ff15e
Running Suite: CRI validation - /tmp/cri-test2158160727
=======================================================
Random Seed: 1652281139

Will run 69 of 76 specs
Running in parallel across 8 processes
```

Both the skip and parallel mode are working.

fuweid:
Ping @feiskyer @saschagrunert

saschagrunert (approved):
Thank you!
/lgtm

PR #938, opened 2022-05-24
Migrate to native ginkgo v2

saschagrunert:
#### What type of PR is this?

/kind cleanup

Two migrations needs to be done to avoid the following warnings:

```
You're using deprecated Ginkgo functionality:
=============================================
  Support for custom reporters has been removed in V2.  Please read the documentation linked to below for Ginkgo's new behavior and for a migration path:
  Learn more at: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-custom-reporters
  Measure is deprecated and will be removed in Ginkgo V2.  Please migrate to gomega/gmeasure.
  Learn more at: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-measure
    github.com/kubernetes-sigs/cri-tools/pkg/benchmark/pod_container.go:60
```

None
None

```release-note
Fixed warnings when running with ginkgo v2.
```

### DIFF
--- a/pkg/benchmark/pod_container.go
+++ b/pkg/benchmark/pod_container.go
@@ -23,6 +23,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gmeasure"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/benchmark/pod_container.go
```diff
@@ -69,18 +66,31 @@ var _ = framework.KubeDescribe("PodSandbox", func() {
...
+				By("stop PodSandbox")
+				rc.StopPodSandbox(podID)
+				By("delete PodSandbox")
+				rc.RemovePodSandbox(podID)
+			}
+
+			// Run a single test to ensure images are available and everything works
+			benchmark()
+
+			// Do the benchmark
```
mikebrow, line 90: immediately after running a benchmark the runtimes are likely to be busy removing image snapshots, hitting golang gc, and such... running the benchmark twice in a row is a different benchmark.

saschagrunert: May I ask you to clarify for me what we should do here? We pre-run the benchmark once to ensure everything is cached and in place (images for example).

saschagrunert:
@feiskyer PTAL

pkg/benchmark/pod_container.go
feiskyer: @mikebrow do you mean we should do this in a different way (e.g. invoking image pull API) to avoid delete operations here?

mikebrow: Just pointing out that gc of image and golang objects/artifacts, are async tasks that lazily follow removing the references (count of containers using an image and snapshots, references in memory to golang memory objects... that lazy activity is likely to run during the 2nd run if the 2nd run immediately follows. Said gc activities will probably use locks that will delay the 2nd run. You could wait between runs for the system to settle down.. or measure both runs as users rarely start a dummy pod to get the system ready to run a second pod.

if you turn on debug you'll see the gc outputs

if you want to pre pull the images, and they have a specific tag (not latest, and not untagged which defaults to latest) then you can use no-pull as that will eliminate the timing issue related to remote image pull from the registry. However then you are not benchmarking the image pull, possibly not the unpack (I'd have to check) :-) (image pull, unpack, and snapshot for a particular container before adding it's volumes for the roots are the expensive image resource operations)  

benchmarking is hard...

saschagrunert:
How about getting this in before v1.25.0?

feiskyer (commented):
sorry, missed this PR

/lgtm
/approve

PR #962, opened 2022-08-05
No truncate PodID in ps command if --no-trunc exists

kebe7jun:
#### What type of PR is this?
/kind bug

`crictl ps --no-trunc` prints truncated PodID, it should be no-trunc.

```release-note
The `crictl ps --no-trunc` command will output the full PodID.
```


saschagrunert (approved):
/lgtm

PR #979, opened 2022-09-02
Print deleted digest when repo tags empty

### LINKED ISSUE #976: `crictl rmi --prune` shows no output for untagged images removed
#### What happened:

Running `crictl rmi --prune` does not log information about all the images pruned as it only prints those that are tagged and images that are referenced by digest are pruned, but this is not reflected in the logs.

This looks to be because the [loop to print this info](https://github.com/kubernetes-sigs/cri-tools/blob/85c21d32b32681271bac1e8602dab9725b32f89f/cmd/crictl/image.go#L409-L411) is over the repoTags:

```go
for _, repoTag := range status.Image.RepoTags {
	fmt.Printf("Deleted: %s\n", repoTag)
}
```

But for images pulled by digest, RepoTags is nil.

If the digest is known, it would be great to have this print `Deleted: <repo>@<digest>\n`, but if that is not possible there should at least be some output that an image was deleted e.g. `Deleted: untagged <imagename> with id <imageid>\n`

1. Reference image by digest
2. Remove container referencing image
3. Run `crictl rmi --prune`
4. See no output related to above image.

thomasf7:
#### What type of PR is this?

/kind bug

This PR updates the `crictl rmi` command to print the repoDigests of deleted images where the image has no repoTags, which can happen when the image digest is used for pulling the image. Currently when this happens the images are deleted with no output whatsoever to indicate this has happened.

Fixes #976

```release-note
Added output to `crictl rmi` when deleting images that were pulled by digest and therefore have no tags.
```

### DIFF
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -406,6 +406,14 @@ var removeImageCommand = &cli.Command{
 				}
 				continue
 			}
+			if len(status.Image.RepoTags) == 0 {
+				// RepoTags is nil when pulling image by repoDigest,
+				// so print deleted using that instead.
+				for _, repoDigest := range status.Image.RepoDigests {
+					fmt.Printf("Deleted: %s\n", repoDigest)
+				}
+				continue
+			}
 			for _, repoTag := range status.Image.RepoTags {
 				fmt.Printf("Deleted: %s\n", repoTag)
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
/lgtm

PR #997, opened 2022-10-05
Implement password input on the command line

### LINKED ISSUE #996: Read password from the command line when pulling images

Currently the pull command accepts username and password either using a based64 encoded auth string or the credentials put directly on the command line.

I would like to be able to pass the password without having to write it in some fashion that will store it on disk like, for example, in the shell history.

#### Why is this needed:

It would help avoid credentials spreading when people need to pull images manually.

sgaist:
#### What type of PR is this?

/kind feature

Rather than writing full credentials on the command line when pulling an image this option allows user to pass a username and get asked for the password as can be done with the ssh command line application.

Fixes #996

```release-note
Added: enter password on the command line for pull related commands

A new option has been introduced to allow user to enter the registry password securely on the command line rather than as part of the command itself.

Example:
    crictl pull --username foobar registry.example.com/myimage:mytag
```

### DIFF
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -21,10 +21,12 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"syscall"
 
 	"github.com/docker/go-units"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
+	"golang.org/x/term"
 	internalapi "k8s.io/cri-api/pkg/apis"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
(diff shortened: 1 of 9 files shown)


saschagrunert:
@haircommander PTAL

cmd/crictl/image.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
haircommander, line 8: nit: if you end up updating something, can you revert this change?

haircommander: actually, the boilerplate linter is failing because of this

sgaist: Strange, it is something that was changed when I used one of `go mod vendor` or `gofmt` but running either of them again after reverting the change does not mention it anymore.

haircommander: thanks!

PR #1005, opened 2022-10-18
critest: added a validation case for SupplementalGroups with the pre-defined groups in the container image

everpeace:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

In most cri implementations, at least containerd and cri-o, when the container's primary UID belongs to some groups in the container image, `SupplementalGroups` are just added to them even if `SupplementalGroups` does not include them (see https://github.com/kubernetes/kubernetes/issues/112879). 

However, the behavior is not clearly stated in the API descriptions of `SupplementalGroups` both in [kubernetes](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3563-L3568) and [cri-api](https://github.com/kubernetes/cri-api/blob/v0.25.3/pkg/apis/runtime/v1/api.proto#L821-L823).

To improve the situation, in https://github.com/kubernetes/kubernetes/pull/113047, I try to clarify the behavior in the API description of `SupplementalGroups` both in kubernetes and cri-api. 

This PR added a test case to critest to validate the behavior as @haircommander recommended in  [the k/k PR](https://github.com/kubernetes/kubernetes/pull/113047#discussion_r997217751).

#### Which issue(s) this PR fixes:

ref: https://github.com/kubernetes/kubernetes/pull/113047#discussion_r997217751

#### Does this PR introduce a user-facing change?

```release-note

```

PR #1008, opened 2022-10-19
test-image: images/image-predefined-groups: stop specifying group in USER directive in Dockerfile

everpeace:
#### What type of PR is this?

/kind bug

It is just because this is not suitable for the test case for SupplementalGroups with pre-defined groups in the container image. In the test case, default user needs to belong to supplementary group in the container image.

ref: https://docs.docker.com/engine/reference/builder/#user
> When specifying a group for the user, the user will have
> only the specified group membership. Any other configured group
> memberships will be ignored.

The image is used in #1005.

#1005 

```release-note

```


everpeace (commented):
@haircommander @saschagrunert I need a small fix in the image. would you mind reviewing this?

saschagrunert (approved):

Issue #1014, opened 2022-10-27
Accept detach sequence for attached containers


Jenkins-J:
#### What would you like to be added:

This is the continuation of a feature request that was made in the [containerd](https://github.com/containerd/containerd) project.

When running `crictl attach -i -t`, there should be a way to detach from the session without ending the process.
Docker provides a way to detach from an attached container, using a specified key sequence [(the default key sequence
being `ctrl-p ctrl-q`)](https://docs.docker.com/engine/reference/commandline/attach/#description). This could be implemented in the `crictl` tool as well.

#### Why is this needed:
Improves user experience and usability. See [this feature request](https://github.com/containerd/containerd/issues/7412) in the containerd repository for more details.

PR #1024, opened 2022-11-10
Add detach key sequence support

Jenkins-J:
Allow detaching from attached container using 'ctrl-p ctrl-q' key sequence. This is the same as the default docker detach key sequence. This only applies when the `-i` and `-t` options are set. Resolves issue #1014.

Signed-off by James Jenkins <James.Jenkins@ibm.com>

```release-note

```

### DIFF
--- a/cmd/crictl/exec.go
+++ b/cmd/crictl/exec.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"time"
@@ -36,6 +37,8 @@ const (
 	kubeletURLSchema = "http"
 	kubeletURLHost   = "http://127.0.0.1:10250"
 )
+
+const detachSequence = "ctrl-p,ctrl-q"
 
 var runtimeExecCommand = &cli.Command{
 	Name:                   "exec",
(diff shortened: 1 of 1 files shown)


cmd/crictl/exec.go
```diff
@@ -173,6 +173,14 @@ func stream(in, tty bool, url *url.URL) error {
 	logrus.Debugf("StreamOptions: %v", streamOptions)
 	if !tty {
 		return executor.Stream(streamOptions)
+	} else {
+		var detachKeys []byte
+		detachKeys, err = mobyterm.ToBytes("ctrl-p,ctrl-q")
```
haircommander, line 178: nit: could this be a const named variable so it's easier to reference

haircommander:
It makes sense to me to have a detach sequence. I find myself worried this is a breaking behavior change. it's not likely, but it's possible there are uses somewhere that specify these keys and don't currently detach. Maybe it should be a `crictl run/attach` argument that defaults to ""? WDYT @saschagrunert

cmd/crictl/exec.go
```diff
@@ -173,6 +173,14 @@ func stream(in, tty bool, url *url.URL) error {
...
+		if err != nil {
+			return fmt.Errorf("could not bind detach keys")
```
saschagrunert, line 180:
```suggestion
			return errors.New("could not bind detach keys")
```

saschagrunert:
> Maybe it should be a `crictl run/attach` argument that defaults to ""?

This would make it more complicated from an end user perspective, right?

cmd/crictl/exec.go
Jenkins-J: Just clarifying, do you mean change `detchKeys` to a const or make the key sequence string (`"ctrl-p,ctrl-q"`) a const?

haircommander: the sequence string please :)

cmd/crictl/exec.go
Jenkins-J: Done

cmd/crictl/exec.go
Jenkins-J: Done

haircommander:
/lgtm

thanks!

saschagrunert (approved):

PR #1058, opened 2023-01-06
Increase default timeout on Windows

jsturtevant:
Signed-off-by: James Sturtevant <jstur@microsoft.com>

#### What type of PR is this?
/kind bug

Most interactions against Windows containers take longer than 2s. This increases the default so the user experience on Windows is better without having to modify every command or know to create a config file with the timeout.  

When creating the first pod it can take a long time to pull due to image container sizes so the timeout here is very high to handle this scenario.  Starting a Pod can take longer than 2s even when the image is already pulled.

/cc @dcantah @marosset 
/sig windows

```release-note
Increase default timeout to 200s when running on Windows 
```

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"sort"
 	"time"
 
@@ -34,6 +35,7 @@ import (
 )
 
 const defaultTimeout = 2 * time.Second
+const defaultTimeoutWindows = 200 * time.Second
 
 var (
 	// RuntimeEndpoint is CRI server runtime endpoint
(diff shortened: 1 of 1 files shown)


dcantah:
200s is just to account for the first pull of the pause image on RunPodSandbox I assume?

dcantah:
Ahh, it's in the description

dcantah (approved):
This seems fine.. although perhaps a comment above the timeout stating the reasoning for the discrepancy between linux and win

marosset:
I'd also like to add that if you forget to increase the timeout when creating a new pod or starting a container on Windows if the operation times out it is very tedious to try and clean up the failed operation.
This has personally caused me some frustration in the past.

marosset (approved):
/lgtm

dcantah:
> I'd also like to add that if you forget to increase the timeout when creating a new pod or starting a container on Windows if the operation times out it is very tedious to try and clean up the failed operation.
> 
> This has personally caused me some frustration in the past.

In what way? Was this the "doesn't have {insert specific metadata field}" error we'd ran into before when testing things via crictl? I'd think that'd also hold true for Linux. I just remember crictl rm not functioning because it whined we didn't fill in every field, but nothing Windows specific

marosset:
> e "doesn't have {insert specific metadata field}" error we'd ran in

Yes it was that error. It may hold try for Linux but I have never encountered the issue on Linux w/ the default timeout :P

dcantah:
Fair enough LOL

haircommander:
/lgtm

@saschagrunert PTAL

feiskyer (approved):

PR #1063, opened 2023-01-16
fix(docs): add tip on latest version

profnandaa:
#### What type of PR is this?

/kind documentation

I noticed that the version is put here more of definitive than a suggestion. This PR adds a tip/comment to guide user to get the latest version in the releases page.

No

### DIFF
--- a/docs/crictl.md
+++ b/docs/crictl.md
@@ -11,7 +11,7 @@ crictl is currently in Beta and still under quick iterations. It is hosted at th
 - using `wget`:
 
 ```sh
-VERSION="v1.24.1"
+VERSION="v1.24.1" # check latest version in /releases page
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
/lgtm

Thank you!

profnandaa:
Thanks, that was quick! @saschagrunert 

I'd like to take on adding a note about installing on Windows too. I know people can figure out but having something explicit or a note, will be helpful.  For new-comers, sometimes when they don't see a platform mentioned, they often think it's not supported.

Something in the lines of:

> The below steps are based on `linux-amd64`, however you can get downloads for all other platforms (Windows, ARM, etc) in the [releases](https://github.com/kubernetes-sigs/cri-tools/releases) page.

This is ok?

PR #1065, opened 2023-01-16
fix(docs): add a note about other supported platforms

profnandaa:
#### What type of PR is this?

/kind documentation

I know people can figure out but having an explicit note about the supported platforms will be helpful. For new-comers, sometimes when they don't see a platform mentioned, they often think it's not supported, especially in the case of Windows.

Sorry should have added this in the previous PR #1063  but came as an after-thought.

No


haircommander:
/lgtm

makes sense!

feiskyer (approved):

PR #1099, opened 2023-02-20
Run GitHub actions by `push` only on `master` and tags

saschagrunert:
#### What type of PR is this?

/kind cleanup

Dependabot creates branches which runs the CI twice per PR. To avoid that, we now filter the branches and tags for the `push` event.

None
None

```release-note
None
```


feiskyer (commented):
Good catch!

/lgtm

PR #1151, opened 2023-05-03
Replace dockershim with cri-dockerd

### LINKED ISSUE #870: [1.24] Remove dockershim from crictl code and docs

### LINKED ISSUE #1162: Restore tests for docker runtime

Need to run tests for docker, similar to cri-o.

#### Why is this needed:

The tests were removed because of "dockershim":

* #1040

But no tests were added for the actual "cri-dockerd".

----

Note: It wasn't actually using dockershim for 1.24 and 1.25 either, since it had been removed from the codebase.

It just renamed the remote socket endpoint, instead of removing dockershim from cri-tools (as per Issue #870).

commit 175fbd1fdff9534f2592cce9b83e5610b198949b

afbjorklund:
#### What type of PR is this?

/kind cleanup

Fixes #870
Fixes #1162

```release-note
dockershim was removed in k8s 1.24
```

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -60,13 +60,13 @@ This will
 - Run the benchmark tests using `ginkgo`
 - Output the test results to STDOUT
 
-critest connects to Unix: `unix:///var/run/cri-dockerd.sock` or Windows: `npipe:////./pipe/cri-dockerd` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+For all runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
 
 ## Additional options
 
 - `-ginkgo.focus`: Only run the tests that match the regular expression.
 - `-image-endpoint`: Set the endpoint of image service. Same with runtime-endpoint if not specified.
-- `-runtime-endpoint`: Set the endpoint of runtime service. Default to Unix: `unix:///var/run/cri-dockerd.sock` or Windows: `npipe:////./pipe/cri-dockerd`.
+- `-runtime-endpoint`: Set the endpoint of runtime service.
 - `-benchmarking-params-file`: optional path to a YAML file containing parameters describing which
 benchmarks should be run.
 - `-benchmarking-output-dir`: optional path to a pre-existing directory in which to write JSON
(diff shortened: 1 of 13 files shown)


docs/benchmark.md
```diff
@@ -60,13 +61,13 @@ This will
 - Run the benchmark tests using `ginkgo`
 - Output the test results to STDOUT
 
-critest connects to Unix: `unix:///var/run/dockershim.sock` or Windows: `tcp://localhost:3735` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
+critest connects to Unix: `unix:///var/run/cri-dockerd.sock` or Windows: `npipe:////./pipe/cri-dockerd` by default. For other runtimes, the endpoint can be set by flags `--runtime-endpoint` and `--image-endpoint`.
```
SergeyKanzhelev, line 64: Is this statement true? Defaults are "unix:///run/containerd/containerd.sock", "unix:///run/crio/crio.sock", "unix:///var/run/cri-dockerd.sock"

afbjorklund: Those defaults are only for `crictl`, while this documentation was for `critest`.

The defaults are deprecated anyway, so both crictl and critest _should_ be configured...

afbjorklund:
The reason it didn't fail sooner, was because the cri-dockerd socket was moved to the old dockershim location.

```shell
sudo /usr/local/bin/cri-dockerd --network-plugin="" --container-runtime-endpoint=unix:///var/run/dockershim.sock
```

For normal installations, there should be no need for dockershim.sock as it doesn't work anymore (since 1.27)

commit f2091fc3e7be70dcb9a02c325781bffe204029ed

afbjorklund:
Looks like the test had been disabled anyway ? This is probably why it breaks in production instead

commit baefda37cab2e7574ef6a893e9caccb365af7c69

docs/benchmark.md
afbjorklund: There was no change in behavior, it was running cri-dockerd also before - it just moved the endpoint location to use the old dockershim path...

Removing the defaults could definitely be done.

PR #1224, opened 2023-07-25
Use `UserSpecifiedImage` for seccomp tests

saschagrunert:
#### What type of PR is this?

/kind bug

The seccomp tests use their own container creation logic, which is why we forgot to add the `UserSpecifiedImage` there as well. This is now fixed.

None
No released code, so we can apply the fix without a release note.

```release-note
None
```


saschagrunert:

PR #1228, opened 2023-08-02
crictl: add metrics{,p} command

haircommander:
#### What type of PR is this?

/kind feature
which are aliases to each other
currently json and yaml output are supported

```release-note
Add `crictl metricsp` command, that allows a user to display pod metrics from the CRI
```


cmd/crictl/pod_metrics.go
```diff
@@ -0,0 +1,155 @@
...
+	output string
+
+	// live watch
+	watch bool
+}
+
+var podMetricsCommand = &cli.Command{
+	Name:                   "metricsp",
+	Aliases:                []string{"metrics"},
+	Usage:                  "List pod resource usage statistics",
```
SergeyKanzhelev, line 41: the description is exactly like `statsp` https://github.com/kubernetes-sigs/cri-tools/blob/94d44a2463ce016540effcc4cf219a2e6630753e/cmd/crictl/pod_stats.go#L52

Should there be any differenece?

haircommander: good call, I updated to `List pod resource usage metrics` which isn't super informative but I think will differentiate enough

SergeyKanzhelev: we really need to expand on both descriptions. I cannot imagine anybody figuring the difference on their own.

cmd/crictl/pod_metrics.go
```diff
@@ -0,0 +1,124 @@
...
+	// live watch
+	watch bool
+}
+
+var podMetricsCommand = &cli.Command{
+	Name:                   "metricsp",
+	Aliases:                []string{"metrics"},
+	Usage:                  "List pod resource usage metrics",
+	UseShortOptionHandling: true,
+	ArgsUsage:              "[ID]",
```
SergeyKanzhelev, line 42: we don't need the ID, right? It is metrics across all pods

cmd/crictl/pod_metrics.go
```diff
@@ -0,0 +1,124 @@
...
+	// output format
+	output string
+
+	// live watch
+	watch bool
+}
+
+var podMetricsCommand = &cli.Command{
+	Name:                   "metricsp",
+	Aliases:                []string{"metrics"},
```
SergeyKanzhelev, line 39: why do we need this alias? It may be confusing as all pod-specific commands should have `p` suffix

haircommander:
updated as suggested @SergeyKanzhelev

SergeyKanzhelev (approved):
/lgtm

PR #1181, opened 2023-06-01
Migrate to field based seccomp tests

saschagrunert:
#### What type of PR is this?

/kind api-change

#### What this PR does / why we need it:
The seccomp profile path field is deprecated since many releases within Kubernetes, where CRI API consumers should use the native field.

We now migrate the tests to use them.

#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
Refers to https://github.com/cri-o/cri-o/pull/7006
#### Does this PR introduce a user-facing change?

```release-note
Switched to the `SecurityProfile` field for the seccomp related critest suite, making `SeccompProfilePath` obsolete.
```

PR #1236, opened 2023-08-10
Fix seccomp value in test

saschagrunert:
#### What type of PR is this?

/kind regression

The seccomp value should be `2` rather than `1`. The test succeeded because we also have the value `Seccomp_filters` which gets matched by "ecc", too. We now check for "Seccomp:" to actually match the right field.

Refers to https://github.com/kubernetes-sigs/cri-tools/pull/1181/files#r1288562050
The field was always `Seccomp:` since https://github.com/torvalds/linux/commit/2f4b3bf6b2318cfaa177ec5a802f4d8d6afbd816 and has never changed.

The `Seccomp_filters:` field has been added in https://github.com/torvalds/linux/commit/c818c03b661cd769e035e41673d5543ba2ebda64

```release-note
Fixed seccomp filter regression for test "should support seccomp default on the container".
```

### DIFF
--- a/pkg/validate/security_context_linux.go
+++ b/pkg/validate/security_context_linux.go
@@ -657,10 +657,23 @@ var _ = framework.KubeDescribe("Security Context", func() {
          }
      ]
 }`
+
+			// seccompProcStatusField is the field of /proc/$PID/status referencing the seccomp filter type.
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev (approved):
/lgtm

PR #1239, opened 2023-08-14
Use previous and latest tag from git in release script

saschagrunert:
#### What type of PR is this?

/kind failing-test

We cannot rely on the discover modes of the release notes tool, because we cut patch releases from time to time from `master`. Means we now simply select the latest and previous tag from the git command.

None
None

```release-note
None
```


feiskyer (commented):
/lgtm

PR #1241, opened 2023-08-16
Add crictl events command

surik:
#### What type of PR is this?

/kind feature

This pull request introduces the `crictl events` command, designed to consume events. It comes in response to the recent developments in both `cri-o` (since version 1.26) and `containerd` (since version 1.7), both of which have integrated endpoints for streaming Evented PLEG. Notably, `containerd` currently lacks multi-subscriber support which I'm trying to address in https://github.com/containerd/containerd/pull/8916.

None

```release-note
Add `crictl` events command to fetch and print container events
```

### DIFF
--- a/cmd/crictl/events.go
+++ b/cmd/crictl/events.go
@@ -92,7 +92,7 @@ func Events(cliContext *cli.Context, client internalapi.RuntimeService) error {
 		case e := <-containerEventsCh:
 			err := outputEvent(e, cliContext.String("output"), cliContext.String("template"))
 			if err != nil {
-				fmt.Printf("formatting container event: %s\n", err)
+				fmt.Printf("failed to format container event with the error: %s\n", err)
 			}
 		}
 	}
--- a/cmd/crictl/events.go
+++ b/cmd/crictl/events.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+	internalapi "k8s.io/cri-api/pkg/apis"
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
+)
+
+var eventsCommand = &cli.Command{
+	Name:                   "events",
+	Usage:                  "Fetch the events of containers",
+	Aliases:                []string{"event"},
+	UseShortOptionHandling: true,
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "output",
+			Aliases: []string{"o"},
+			Value:   "json",
+			Usage:   "Output format, One of: json|yaml|go-template",
+		},
+		&cli.StringFlag{
+			Name:  "template",
+			Usage: "The template string is only used when output is go-template; The Template format is golang template",
+		},
+	},
+	Action: func(c *cli.Context) error {
+		if c.NArg() != 0 {
+			return cli.ShowSubcommandHelp(c)
+		}
+
+		runtimeClient, err := getRuntimeService(c, 0)
+		if err != nil {
+			return err
+		}
+
+		if err = Events(c, runtimeClient); err != nil {
+			return fmt.Errorf("getting container events: %w", err)
+		}
+
+		return nil
+	},
+}
+
+func Events(cliContext *cli.Context, client internalapi.RuntimeService) error {
+	errCh := make(chan error, 1)
+
+	containerEventsCh := make(chan *pb.ContainerEventResponse)
+	go func() {
+		logrus.Debug("getting container events")
+		err := client.GetContainerEvents(containerEventsCh)
+		if err == io.EOF {
+			errCh <- nil
+			return
+		}
+		errCh <- err
+	}()
+
+	for {
+		select {
+		case err := <-errCh:
+			return err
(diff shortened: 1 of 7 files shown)


saschagrunert (approved):
Thank you!

cmd/crictl/events.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		errCh <- err
+	}()
+
+	for {
+		select {
+		case err := <-errCh:
+			return err
+		case e := <-containerEventsCh:
+			err := outputEvent(e, cliContext.String("output"), cliContext.String("template"))
```
SergeyKanzhelev, line 83: should the single formatting error fail the whole command?

docs/crictl.md
```diff
@@ -65,6 +65,7 @@ COMMANDS:
 - `statsp`:             List pod(s) resource usage statistics
 - `completion`:         Output bash shell completion code
 - `checkpoint`:         Checkpoint one or more running containers
+- `events, event`:      Fetch the events of containers
```
SergeyKanzhelev, line 68: should we say "Stream the events of containers"?

```suggestion
- `events, event`:      Stream the events of containers
```

test/e2e/events_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// The actual test suite
+var _ = t.Describe("events", func() {
+	It("should succeed", func() {
+		// Given
+		endpoint, testDir, crio := t.StartCrio()
+
+		// When
+		session := t.CrictlWithEndpointNoWait(endpoint, "events")
```
SergeyKanzhelev, line 32: why no wait? I think the wait is very much intentional to avoid flakes of slow bootstrap

cmd/crictl/events.go
```diff
@@ -0,0 +1,89 @@
...
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "output",
+			Aliases: []string{"o"},
+			Value:   "json",
+			Usage:   "Output format, One of: json|yaml|go-template",
+		},
+		&cli.StringFlag{
+			Name:  "template",
+			Usage: "The template string is only used when output is go-template; The Template format is golang template",
```
SergeyKanzhelev, line 43: let's add explicit validation that it is not specified without `go-template` to avoid confusion

cmd/crictl/events.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+	internalapi "k8s.io/cri-api/pkg/apis"
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
+)
+
+var eventsCommand = &cli.Command{
+	Name:                   "events",
+	Usage:                  "Fetch the events of containers",
```
SergeyKanzhelev, line 31:
```suggestion
	Usage:                  "Stream the events of containers",
```

test/e2e/events_test.go
surik: I don't have much experience with this testing framework but what I see in [the docs](https://pkg.go.dev/github.com/onsi/gomega@v1.27.10/gexec#Session.Wait) for the `Wait` method:
```
Wait waits until the wrapped command exits. It can be passed an optional timeout. If the command does not exit within the timeout, Wait will trigger a test failure.
```
As the `events` command doesn't terminate we'll probably have the test fail after waiting for 1 minute.

Maybe I can add a short timeout before `Expect(session.Out).ToNot(Say("unknown method GetContainerEvents")) // no errors` to help with slow bootstrap?

cmd/crictl/events.go
```diff
@@ -48,6 +48,16 @@ var eventsCommand = &cli.Command{
 			return cli.ShowSubcommandHelp(c)
 		}
 
+		switch format := c.String("output"); format {
+		case "json", "yaml":
+			if len(c.String("template")) > 0 {
+				return fmt.Errorf("template can't be used with %q format", format)
+			}
+		case "go-template":
```
SergeyKanzhelev, line 56: should we confirm that the `template` is defined?

cmd/crictl/events.go
```diff
@@ -0,0 +1,99 @@
...
+	}()
+
+	for {
+		select {
+		case err := <-errCh:
+			return err
+		case e := <-containerEventsCh:
+			err := outputEvent(e, cliContext.String("output"), cliContext.String("template"))
+			if err != nil {
+				fmt.Printf("formatting container event: %s\n", err)
```
SergeyKanzhelev, line 95:
```suggestion
				fmt.Printf("failed to format container event with the error: %s\n", err)
```

cmd/crictl/events.go
surik: Yes, I believe that would make sense. Although I don't see that there is extensive validation in other commands. Would it be beneficial to address that once and for all?

SergeyKanzhelev: I'd suggest to concentrate on one command at a time. If you want to improve other commands - please do. 

For this command specifically I think it is important to fail on validation early as it will wait for events and only fail trying to format first of them. It will be quite annoying

test/e2e/events_test.go
```diff
@@ -0,0 +1,47 @@
...
+		Expect(session.Out).ToNot(Say("unknown method GetContainerEvents")) // no errors
+
+		// Then
+		session.Terminate()
```
SergeyKanzhelev, line 44: defer this?

test/e2e/events_test.go
SergeyKanzhelev: oh, you are right. I misundertood.

test/e2e/events_test.go
```diff
@@ -41,7 +45,7 @@ var _ = t.Describe("events", func() {
 		Expect(session.Out).ToNot(Say("unknown method GetContainerEvents")) // no errors
 
 		// Then
-		session.Terminate()
+		defer session.Terminate()
```
SergeyKanzhelev, line 48: need to clean up tear down sequence.

1. call defer right after getting the session. SO it will be terminated no matter what
2. defer will execute after stopping crio as written now. Need to move stopping crio somewhere. I think the best can be done is moving StopCrio to AfterEach.

surik: This is done. I had to split tests into two suites: 1) one that relies on cri-o 2) a set of options validation tests that don't need to start cri-o. I hope this is fine.

SergeyKanzhelev (approved):
/lgtm

PR #1247, opened 2023-08-21
Add show pinned information for images

Iceber:
#### What type of PR is this?

/kind feature

Add showing of pinned information when using the `crictl images` command.

1. `crictl images -v` prints Pinned when image is pinned
```bash
$ crictl images -v
// ignore some print...

ID: sha256:e6f1816883972d4be47bd48879a08919b96afcd344132622e4d444987919323c
RepoTags: k8s.m.daocloud.io/pause:3.9
RepoDigests: k8s.m.daocloud.io/pause@sha256:7031c1b283388d2c2e09b57badb803c05ebed362dc88d84b480cc47f72a21097
Size: 321520
Uid: &Int64Value{Value:65535,}

ID: sha256:261e6f4fa9213ab2648fe05c7c0abefe50a3ddd561dfa4547a415cd62d8e5fe2
RepoTags: quay.io/tigera/operator:v1.30.4
RepoDigests: quay.io/tigera/operator@sha256:780eeab342e62bd200e533a960b548216b23b8bde7a672c4527f29eec9ce2d79
Size: 21216944
Uid: &Int64Value{Value:10001,}

ID: sha256:ed210e3e4a5bae1237f1bb44d72a05a2f1e5c6bfe7a7e73da179e2534269c459
RepoTags: registry.aliyuncs.com/google_containers/pause:3.5
RepoDigests: registry.aliyuncs.com/google_containers/pause@sha256:1ff6c18fbef2045af6b9c16bf034cc421a29027b800e4f9b68ae9b1cb3e9ae07
Size: 301416
Uid: &Int64Value{Value:65535,}
Pinned: true
```

2. add --pinned flag for `crictl images`
```bash
$ crictl images --pinned
IMAGE                                               TAG                 IMAGE ID            SIZE                PINNED
// ignore some print ....
k8s.m.daocloud.io/kube-scheduler                    v1.27.4             98ef2570f3cde       18.2MB              false
k8s.m.daocloud.io/pause                             3.9                 e6f1816883972       322kB               false
quay.io/tigera/operator                             v1.30.4             261e6f4fa9213       21.2MB              false
registry.aliyuncs.com/google_containers/pause       3.5                 ed210e3e4a5ba       301kB               true
```
```bash
crictl images --digests --pinned
IMAGE                                               TAG                 DIGEST              IMAGE ID            SIZE                PINNED
k8s.m.daocloud.io/kube-proxy                        v1.27.4             4bcb707da9898       6848d7eda0341       23.9MB              false
k8s.m.daocloud.io/kube-scheduler                    v1.27.4             5897d7a97d23d       98ef2570f3cde       18.2MB              false
k8s.m.daocloud.io/pause                             3.9                 7031c1b283388       e6f1816883972       322kB               false
quay.io/tigera/operator                             v1.30.4             780eeab342e62       261e6f4fa9213       21.2MB              false
registry.aliyuncs.com/google_containers/pause       3.5                 1ff6c18fbef20       ed210e3e4a5ba       301kB               true
```

```release-note
Add showing of pinned information when using the `crictl images` command.
```

### DIFF
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -231,6 +231,9 @@ var listImageCommand = &cli.Command{
 			if image.Username != "" {
 				fmt.Printf("Username: %v\n", image.Username)
 			}
+			if image.Pinned {
+				fmt.Printf("Pinned: %v\n", image.Pinned)
+			}
 			fmt.Printf("\n")
 		}
 		display.Flush()
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -155,6 +155,10 @@ var listImageCommand = &cli.Command{
 			Name:  "no-trunc",
 			Usage: "Show output without truncating the ID",
 		},
+		&cli.BoolFlag{
+			Name:  "pinned",
+			Usage: "Show whether the image is pinned or not",
+		},
 	},
 	Action: func(c *cli.Context) error {
 		if c.NArg() > 1 {
@@ -183,14 +187,19 @@ var listImageCommand = &cli.Command{
 		display := newTableDisplay(20, 1, 3, ' ', 0)
 		verbose := c.Bool("verbose")
 		showDigest := c.Bool("digests")
+		showPinned := c.Bool("pinned")
 		quiet := c.Bool("quiet")
 		noTrunc := c.Bool("no-trunc")
 		if !verbose && !quiet {
+			row := []string{columnImage, columnTag}
 			if showDigest {
-				display.AddRow([]string{columnImage, columnTag, columnDigest, columnImageID, columnSize})
-			} else {
-				display.AddRow([]string{columnImage, columnTag, columnImageID, columnSize})
+				row = append(row, columnDigest)
 			}
+			row = append(row, columnImageID, columnSize)
+			if showPinned {
+				row = append(row, columnPinned)
+			}
+			display.AddRow(row)
 		}
 		for _, image := range r.Images {
 			if quiet {
@@ -207,11 +216,15 @@ var listImageCommand = &cli.Command{
 					repoDigest = getTruncatedID(repoDigest, "sha256:")
 				}
 				for _, repoTagPair := range repoTagPairs {
+					row := []string{repoTagPair[0], repoTagPair[1]}
 					if showDigest {
-						display.AddRow([]string{repoTagPair[0], repoTagPair[1], repoDigest, id, size})
-					} else {
-						display.AddRow([]string{repoTagPair[0], repoTagPair[1], id, size})
+						row = append(row, repoDigest)
 					}
+					row = append(row, id, size)
+					if showPinned {
+						row = append(row, fmt.Sprintf("%v", image.Pinned))
+					}
+					display.AddRow(row)
 				}
 				continue
 			}
--- a/cmd/crictl/display.go
+++ b/cmd/crictl/display.go
@@ -37,6 +37,7 @@ const (
 	columnNamespace  = "NAMESPACE"
 	columnSize       = "SIZE"
 	columnTag        = "TAG"
+	columnPinned     = "PINNED"
 	columnDigest     = "DIGEST"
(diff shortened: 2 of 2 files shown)


saschagrunert (commented):
Code LGTM

Can we add a release note to the body of this PR to document the change?

Iceber:
@saschagrunert updated, ptal

afbjorklund:
There is a"typo" in the  description and commits, it says `ctr images` (containerd) when it means `crictl images`?

Verified that it also works with Docker, presuming that https://github.com/Mirantis/cri-dockerd/pull/194 is patched into `cri-dockerd` first.

```
ID: sha256:e6f1816883972d4be47bd48879a08919b96afcd344132622e4d444987919323c
RepoTags: registry.k8s.io/pause:3.9
RepoDigests: registry.k8s.io/pause@sha256:7031c1b283388d2c2e09b57badb803c05ebed362dc88d84b480cc47f72a21097
Size: 743952
Pinned: true
```

Iceber:
> There is a"typo" in the description and commits, it says ctr images (containerd) when it means crictl images?

Sorry, I misspelled the name of the command.

cmd/crictl/image.go
```diff
@@ -182,15 +186,19 @@ var listImageCommand = &cli.Command{
...
 		display := newTableDisplay(20, 1, 3, ' ', 0)
 		verbose := c.Bool("verbose")
+		showPinned := c.Bool("pinned")
 		showDigest := c.Bool("digests")
 		quiet := c.Bool("quiet")
 		noTrunc := c.Bool("no-trunc")
 		if !verbose && !quiet {
+			row := []string{columnImage, columnTag}
+			if showPinned {
+				row = append(row, columnPinned)
```
SergeyKanzhelev, line 196: can we add a new column to the end? The column is not that important and in most cases will only be true for a single image

cmd/crictl/image.go
```diff
@@ -155,6 +155,10 @@ var listImageCommand = &cli.Command{
 			Name:  "no-trunc",
 			Usage: "Show output without truncating the ID",
 		},
+		&cli.BoolFlag{
```
SergeyKanzhelev, line 158: why wouldn't we just output it by default without an explicit flag?

cmd/crictl/image.go
Iceber: My thinking is that when the user passes the --pinned flag, maybe he is trying to view the pinned information, so I put this column in front of it.
 if I put it at the end this is what it would look like, one more line of code:
```go
			row := []string{columnImage, columnTag}
			if showPinned {
				row = append(row, columnPinned)
			}
			row = append(row, columnImageID, columnSize)      
			if showDigest {
				row = append(row, columnDigest)
			}
			display.AddRow(row)
```

Of course, I'm keeping an open mind about the placement of this column, if you suggest putting it at the end , then I'll update this

cmd/crictl/image.go
Iceber: I didn't think this field was that important, so I left the default output format unchanged.

Of course I'm not particularly sure, I'll take everyone's advice on this

cmd/crictl/image.go
SergeyKanzhelev: Trying to understand the history here. I think the `--digests` was added to match `docker cli`. And docker has it because there were no digests for older images: https://docs.docker.com/engine/reference/commandline/images/#digests

If we want to keep compatibility with `docker`, having an additional flag and not changing the default output makes sense.

It may be a good idea to create a test for this and explicitly mention that the ordering of columns and the default list of columns is driven by compatibility with the docker cli.

Ordering-wise I really don't like the idea of inserting column in the middle. If customer wants it - they can use template to specify the exact order of columns.

My suggestion will be to put pinned to be the last column.

Iceber:
@SergeyKanzhelev I've updated the description and code, PTAL, Thanks

SergeyKanzhelev (approved):
/lgtm

Issue #1296, opened 2023-11-23
when we transition from SPDY to WebSockets in exec and attach


duguhaotian:
#### What would you like to be added:

K8s have merged pr: https://github.com/kubernetes/kubernetes/pull/119186

KEP: https://github.com/kubernetes/enhancements/issues/4006

#### Why is this needed:
container runtime need to test this new feature

saschagrunert:
@duguhaotian thank you for the request. Do you plan to contribute the feature to cri-tools?

duguhaotian:
> @duguhaotian thank you for the request. Do you plan to contribute the feature to cri-tools?

sorry, i just want to know our plan.

PR #1305, opened 2023-12-05
support websocket for exec and attach

duguhaotian:
#### What type of PR is this?

Add one of the following kinds:
/kind feature

Optionally add one or more of the following kinds if applicable:
/kind api-change

#1296 

add new option for `exec/attach`, manual is: `crictl exec/attach -w -it ${container-id} ls`

#### test

i use container engine is [iSulad](https://github.com/openeuler-mirror/iSulad) which implement stream server is websocket.

```
$ export CRICTL_REMOTE_COMMAND_WEBSOCKETS=true
$ mycrictl exec -w -it b277916602c7 ls
bin   dev   etc   home  proc  root  sys   tmp   usr   var

# tcpdump logs
08:53:15.674128 IP localhost.10350 > localhost.40462: Flags [.], ack 303, win 510, options [nop,nop,TS val 1395824017 ecr 1395824017], length 0
E..4o.@.@..,........(n...       ..T=.......(.....
S2..S2..
08:53:15.675140 IP localhost.10350 > localhost.40462: Flags [P.], seq 1:173, ack 303, win 512, options [nop,nop,TS val 1395824018 ecr 1395824017], length 172
E...o.@.@...........(n...       ..T=.............
S2..S2..HTTP/1.1 101 Switching Protocols
Upgrade: WebSocket
Connection: Upgrade
Sec-WebSocket-Accept: mvvvKmgjEhY8040cH4xt8YSYr+Y=
Sec-WebSocket-Protocol: v5.channel.k8s.io


08:53:15.675156 IP localhost.40462 > localhost.10350: Flags [.], ack 173, win 511, options [nop,nop,TS val 1395824018 ecr 1395824018], length 0
E..4|w@.@..J..........(nT=...   .......(.....
S2..S2..
08:53:15.675527 IP localhost.40462 > localhost.10350: Flags [P.], seq 303:336, ack 173, win 512, options [nop,nop,TS val 1395824018 ecr 1395824018], length 33
E..U|x@.@..(..........(nT=...   .......I.....
S2..S2....S..mW..::...q..^g..%6...'..^c..
08:53:15.716065 IP localhost.10350 > localhost.40462: Flags [.], ack 336, win 512, options [nop,nop,TS val 1395824059 ecr 1395824018], length 0
E..4o.@.@..*........(n...       ..T=.......(.....
S2..S2..
08:53:15.755820 IP localhost.10350 > localhost.40462: Flags [P.], seq 173:337, ack 336, win 512, options [nop,nop,TS val 1395824099 ecr 1395824018], length 164
E...o.@.@...........(n...       ..T=.............
S2..S2...~....[1;34mbin.[m   .[1;34mdev.[m   .[1;34metc.[m   .[1;34mhome.[m  .[1;34mproc.[m  .[1;34mroot.[m  .[1;34msys.[m   .[1;34mtmp.[m   .[1;34musr.[m   .[1;34mvar.[m

08:53:15.796061 IP localhost.40462 > localhost.10350: Flags [.], ack 337, win 512, options [nop,nop,TS val 1395824139 ecr 1395824099], length 0
E..4|y@.@..H..........(nT=...   .2.....(.....
S2..S2..
08:53:15.856882 IP localhost.10350 > localhost.40462: Flags [P.], seq 337:360, ack 336, win 512, options [nop,nop,TS val 1395824200 ecr 1395824139], length 23
E..Ko.@.@...........(n...       .2T=.......?.....
S2.HS2.....{"status":"Success"}
08:53:15.856891 IP localhost.40462 > localhost.10350: Flags [.], ack 360, win 512, options [nop,nop,TS val 1395824200 ecr 1395824200], length 0

```

### DIFF
--- a/cmd/crictl/exec.go
+++ b/cmd/crictl/exec.go
@@ -26,6 +26,7 @@ import (
 	mobyterm "github.com/moby/term"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
+	"k8s.io/apimachinery/pkg/util/httpstream"
 	restclient "k8s.io/client-go/rest"
 	remoteclient "k8s.io/client-go/tools/remotecommand"
 	internalapi "k8s.io/cri-api/pkg/apis"
@@ -163,8 +164,27 @@ func Exec(ctx context.Context, client internalapi.RuntimeService, opts execOptio
 	return stream(ctx, opts.stdin, opts.tty, URL)
 }
 
+func getExecutor(url *url.URL) (exec remoteclient.Executor, err error) {
+	config := &restclient.Config{TLSClientConfig: restclient.TLSClientConfig{Insecure: true}}
+
+	exec, err = remoteclient.NewSPDYExecutor(config, "POST", url)
+	if err != nil {
+		return
+	}
+	if !IsEnabledWebsockets() {
+		return
+	}
+	wexec, err := remoteclient.NewWebSocketExecutor(config, "GET", url.String())
+	if err != nil {
+		return
+	}
+
+	exec, err = remoteclient.NewFallbackExecutor(wexec, exec, httpstream.IsUpgradeFailure)
+	return
+}
+
 func stream(ctx context.Context, in, tty bool, url *url.URL) error {
-	executor, err := remoteclient.NewSPDYExecutor(&restclient.Config{TLSClientConfig: restclient.TLSClientConfig{Insecure: true}}, "POST", url)
+	executor, err := getExecutor(url)
 	if err != nil {
 		return err
 	}
--- a/cmd/crictl/util.go
+++ b/cmd/crictl/util.go
@@ -70,6 +70,10 @@ func SetupInterruptSignalHandler() <-chan struct{} {
 	return signalIntStopCh
 }
 
+func IsEnabledWebsockets() bool {
+	return strings.ToLower(os.Getenv("CRICTL_REMOTE_COMMAND_WEBSOCKETS")) == "true"
+}
+
 type listOptions struct {
 	// id of container or sandbox
 	id string


cmd/crictl/attach.go
```diff
@@ -63,9 +68,10 @@ var runtimeAttachCommand = &cli.Command{
 		defer cancel()
 
 		var opts = attachOptions{
-			id:    id,
-			tty:   c.Bool("tty"),
-			stdin: c.Bool("stdin"),
+			id:        id,
+			tty:       c.Bool("tty"),
+			stdin:     c.Bool("stdin"),
+			websocket: c.Bool("websocket"),
```
SergeyKanzhelev, line 74: would the name `use-websocket` be better? Or maybe have something with three options: `-- transport=spdy|websocket|default` where default will do websocket with the fallback to spdy?

cmd/crictl/exec.go
```diff
@@ -160,15 +167,27 @@ func Exec(ctx context.Context, client internalapi.RuntimeService, opts execOptio
...
 	if err != nil {
 		return err
 	}
 
+	if websocket {
+		websocketExecutor, err := remoteclient.NewWebSocketExecutor(config, "GET", url.String())
+		if err != nil {
+			return err
+		}
+		executor, err = remoteclient.NewFallbackExecutor(websocketExecutor, executor, httpstream.IsUpgradeFailure)
```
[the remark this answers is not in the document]
duguhaotian: fixed

cmd/crictl/attach.go
duguhaotian: fixed

duguhaotian:
i donot know why check of runc and cri-o critest failed?

the report error is spdy upgrade failed? why?

pkg/validate/streaming.go
```diff
@@ -189,6 +190,12 @@ func checkExec(c internalapi.RuntimeService, execServerURL, stdout string, stdou
 	e, err := remoteclient.NewSPDYExecutor(&rest.Config{TLSClientConfig: rest.TLSClientConfig{Insecure: true}}, "POST", url)
 	framework.ExpectNoError(err, "failed to create executor for %q", execServerURL)
 
+	we, err := remoteclient.NewWebSocketExecutor(&rest.Config{TLSClientConfig: rest.TLSClientConfig{Insecure: true}}, "GET", url.String())
+	framework.ExpectNoError(err, "failed to create websocket executor for %q", execServerURL)
+
+	e, err = remoteclient.NewFallbackExecutor(we, e, httpstream.IsUpgradeFailure)
+	framework.ExpectNoError(err, "failed to create fallback executor for %q", execServerURL)
+
```
saschagrunert, line 198: This is making the tests fail in the CI. Do runtimes have to support that?

duguhaotian: If websocket failed, will fallback to spdy that is same to before. Why make ci failed?
I think that IsUpgradeFailure will ignore error of websocket upgrade

duguhaotian: just remove this first, wait runtimes support it

cmd/crictl/exec.go
```diff
@@ -67,6 +68,12 @@ var runtimeExecCommand = &cli.Command{
 			Aliases: []string{"i"},
 			Usage:   "Keep STDIN open",
 		},
+		&cli.StringFlag{
+			Name:    "transport",
+			Aliases: []string{"p"},
+			Usage:   "transport protocol, One of: spdy|websocket|default",
+			Value:   "default",
```
saschagrunert, line 75: I think we should keep `spdy` as default value and remove `default`.

saschagrunert (commented):
How can we test this? Using CRI-O `main` always returns an error on exec:

```
sudo ./build/bin/linux/amd64/crictl exec -p websocket 20168cc1a0aefa251fed echo hi
FATA[0000] execing command in container: unable to upgrade streaming request: websocket: bad handshake
```
CRI-O logs:

```
ERRO[2023-12-08 09:28:57.322412994+01:00] unable to upgrade websocket connection: websocket server finished before becoming ready
```

The error is coming from:

https://github.com/cri-o/cri-o/blob/f40da54704c5c32fb24b2b0e204424afae5b0dc2/vendor/k8s.io/apimachinery/pkg/util/httpstream/wsstream/conn.go#L223-L225

We're vendoring Kubernetes v1.29.0-rc.1 right now, so I'm wondering which part is missing what :thinking:

cmd/crictl/exec.go
duguhaotian: fixed

saschagrunert:
> I think CRI-O need update code of websocket, like support 'v5.channel.k8s.io'

I may miss something here, but we use the code vendored from Kubernetes: https://github.com/cri-o/cri-o/blob/f40da54704c5c32fb24b2b0e204424afae5b0dc2/go.mod#L75-L80

saschagrunert:
@duguhaotian it returns there: https://github.com/cri-o/cri-o/blob/f40da54704c5c32fb24b2b0e204424afae5b0dc2/vendor/golang.org/x/net/websocket/server.go#L84

With the error message:
```
requested protocol(s) are not supported: [v5.channel.k8s.io]; supports [ channel.k8s.io base64.channel.k8s.io v4.channel.k8s.io v4.base64.channel.k8s.io]
```

duguhaotian:
> @duguhaotian it returns there: https://github.com/cri-o/cri-o/blob/f40da54704c5c32fb24b2b0e204424afae5b0dc2/vendor/golang.org/x/net/websocket/server.go#L84
> 
> With the error message:
> 
> ```
> requested protocol(s) are not supported: [v5.channel.k8s.io]; supports [ channel.k8s.io base64.channel.k8s.io v4.channel.k8s.io v4.base64.channel.k8s.io]
> ```
> 
> @duguhaotian is it intentional that runtimes only support v1-v4? Ref: https://github.com/kubernetes/client-go/blob/84a6fe7e4032ae1b8bc03b5208e771c5f7103549/tools/remotecommand/websocket.go#L89-L92

1. cri-o not support `v5.channel.k8s.io`;
2. i think runtimes just haven't done a good job of supporting `v5.channel.k8s.io`, and runtimes should support it.

cmd/crictl/exec.go
```diff
@@ -160,11 +168,35 @@ func Exec(ctx context.Context, client internalapi.RuntimeService, opts execOptio
...
+}
+
+func getExecutor(protocol string, url *url.URL) (exec remoteclient.Executor, err error) {
+	config := &restclient.Config{TLSClientConfig: restclient.TLSClientConfig{Insecure: true}}
+
+	switch protocol {
+	case "spdy":
+		exec, err = remoteclient.NewSPDYExecutor(config, "POST", url)
+	case "websocket":
+		exec, err = remoteclient.NewWebSocketExecutor(config, "GET", url.String())
```
saschagrunert, line 181: Can we use `remoteclient.NewWebSocketExecutorForProtocols` and make the protocols CLI configurable?

saschagrunert (changes requested):

cmd/crictl/exec.go
duguhaotian: ofcourse, default use StreamProtocolV5Name?

saschagrunert: @duguhaotian yeah

duguhaotian: change option to `crictl --websocket="x..channel.k8s.io"`, if not set `--websocket`, will use spdy

aojea:
There is a KEP and a plane to gradually move on to SPDY https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4006-transition-spdy-to-websockets , we should coordinate, @seans3 is leading this effort and if we implement different bits at different times we'll risk to drift

cmd/crictl/exec.go
duguhaotian: > @duguhaotian yeah, this way we'll not require to flip the default again in the future once [kubernetes/kubernetes#122263](https://github.com/kubernetes/kubernetes/issues/122263) got resolved.

fixed

cmd/crictl/attach.go
```diff
@@ -43,6 +43,17 @@ var runtimeAttachCommand = &cli.Command{
 			Aliases: []string{"i"},
 			Usage:   "Keep STDIN open",
 		},
+		&cli.StringFlag{
+			Name:    "transport-protocol",
+			Aliases: []string{"p"},
+			Usage:   "protocol of transport, One of: spdy|websocket",
+			Value:   "spdy",
+		},
+		&cli.BoolFlag{
```
aojea, line 52: why a flag and not using fall back , as we are doing with kubectl and is defined in the KEP?

flags are hard to remove in the future and become APIs for users, I also prefer an environment variable instead

aojea:
I think this should follow the same flow that the kubectl implementation designed in the KEP https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4006-transition-spdy-to-websockets#proposal-kubectl-websocket-executor-and-fallback-executor

use the legacy by default and have an environment variable to opt in to websockets by default but fallback to the legacy https://github.com/kubernetes/kubernetes/pull/119186/files#diff-62da7750fb41c9d53934dd9706ba719f0c0efdf931b6b36eb46498505e4d3520

cmd/crictl/attach.go
duguhaotian: thk you reply. fixed

duguhaotian:
> I think this should follow the same flow that the kubectl implementation designed in the KEP https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4006-transition-spdy-to-websockets#proposal-kubectl-websocket-executor-and-fallback-executor
> 
> use the legacy by default and have an environment variable to opt in to websockets by default but fallback to the legacy https://github.com/kubernetes/kubernetes/pull/119186/files#diff-62da7750fb41c9d53934dd9706ba719f0c0efdf931b6b36eb46498505e4d3520

I gree with you, my first version which use fallback executor.
But, current some runtime(like crio) do not support v5, so will cause CI testcase failed.

aojea:
> But, current some runtime(like crio) do not support v5, so will cause CI testcase failed.

Please check more in detail how this is implemented in Kubernetes, I linked the corresponding code and KEP for context

1. v5 will be only used when a special env variable is set (feature flag here is an env variable)
2. v5 will fallback to v4 if it fails because the container runtime does not implement it, so is not possible CI fails because the fallback has to work, otherwise is a bug in the implementation
3. if no env variable is set, no new code path is executed

saschagrunert:
I think we could at least support the websocket executor in crictl for users to play around with the feature and runtimes.

cmd/crictl/util.go
```diff
@@ -70,6 +70,10 @@ func SetupInterruptSignalHandler() <-chan struct{} {
 	return signalIntStopCh
 }
 
+func IsEnabledWebsockets() bool {
+	return strings.ToLower(os.Getenv("CRICTL_REMOTE_COMMAND_WEBSOCKETS")) == "true"
+}
```
saschagrunert, line 75: Let's remove this

cmd/crictl/exec.go
```diff
@@ -163,8 +164,27 @@ func Exec(ctx context.Context, client internalapi.RuntimeService, opts execOptio
...
+func getExecutor(url *url.URL) (exec remoteclient.Executor, err error) {
+	config := &restclient.Config{TLSClientConfig: restclient.TLSClientConfig{Insecure: true}}
+
+	exec, err = remoteclient.NewSPDYExecutor(config, "POST", url)
+	if err != nil {
+		return
+	}
+	if !IsEnabledWebsockets() {
+		return
+	}
```
saschagrunert, line 176:
```suggestion
```

cmd/crictl/exec.go
```diff
@@ -163,8 +164,27 @@ func Exec(ctx context.Context, client internalapi.RuntimeService, opts execOptio
 	return stream(ctx, opts.stdin, opts.tty, URL)
 }
 
+func getExecutor(url *url.URL) (exec remoteclient.Executor, err error) {
```
saschagrunert, line 167: I'd say we reintroduce the flag to select the executor.

saschagrunert (changes requested):

saschagrunert:
Thank you for the effort on this topic @duguhaotian, we really appreciate it. 🙏

PR #1355, opened 2024-02-14
Don't fail-fast containerd test matrix

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now set `fail-fast` to `false` for large matrixes like the containerd tests to not abort ongoing jobs which may succeed.

Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

None
None

```release-note
None
```


SergeyKanzhelev (approved):
/lgtm

PR #1360, opened 2024-02-22
Don't remove pinned images on image prune

### LINKED ISSUE #1356: Don't remove pinned images when pruning
After discussions in https://github.com/containerd/containerd/issues/9793, it was found that crictl does not honor pinned images, so they will erroneously get garbage collected. I would like crictl to keep pinned images when pruning.

#### Why is this needed:
This would normally only solve edge cases where people need pinned images for other reasons, but it was also found that, due to how crictl checks for its containers (using the runtime service instead of a container service), the pause container is never reported as running, and thus will get garbage collected. Implementing this will have the nice side effect of mitigating this problem. (I assume pause images should always be pinned in proper workflows, but I'm not super familiar with Kubernetes so I could be wrong in making this assumption.)

sondavidb:
#### What type of PR is this?
/kind feature

On `cricrl rmi --prune`, crictl ignores the pinned images label when removing images. Not only is this unexpected behavior, it also removes the pause container, because the container runtime endpoint doesn't return the pause container in its list of running containers.

This change honors the pinned label on `cricrl rmi --prune`, which also stops the pause container from being removed, as it is pinned in every workflow.

Fixes #1356 

I'd like to add testing for this, but I haven't been able to figure out how to make it work. I just ran a simple workflow of pulling an image, pinning it, and then running `crictl rmi --prune`.

```release-note
nerdctl rmi will not remove pinned images on --prune
```

### DIFF
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -390,7 +390,12 @@ var removeImageCommand = &cli.Command{
 				return err
 			}
 			for _, img := range r {
-				logrus.Debugf("Adding image to be removed: %v", img.GetId())
+				// Pinned images should not be removed on prune.
+				if prune && img.Pinned {
+					logrus.Debugf("Excluding pinned container image: %v", img.GetId())
+					continue
+				}
+				logrus.Debugf("Adding container image to be removed: %v", img.GetId())
 				ids[img.GetId()] = true
 			}
 		}


cmd/crictl/image.go
```diff
@@ -390,8 +390,11 @@ var removeImageCommand = &cli.Command{
 				return err
 			}
 			for _, img := range r {
-				logrus.Debugf("Adding image to be removed: %v", img.GetId())
-				ids[img.GetId()] = true
+				// Pinned images should not be removed on prune.
+				if all || (prune && !img.Pinned) {
+					logrus.Debugf("Adding image to be removed: %v", img.GetId())
+					ids[img.GetId()] = true
+				}
```
kwilczynski, line 397: What about something like this?

```suggestion
                                  // Pinned images should not be removed on prune.
                                  if prune && img.Pinned {
                                          logrus.Debugf("Excluding pinned container image: %v", img.GetId())
                                          continue
                                  }
                                  logrus.Debugf("Adding container image to be removed: %v", img.GetId())
                                  ids[img.GetId()] = true
```

I updated the wording to match other log messages.

kwilczynski:
@sondavidb, looks good!

For testing, have a look at the testes that were recently added as part of some other Pull Request. Perhaps it will give you an idea how to do it. This will cover unit tests, for end-to-end testes, it would be a bit more involved to do, albeit not impossible.

However, manual testing is also fine.

saschagrunert (approved):

kwilczynski:
/release-note-edit ```crictl rmi will not remove pinned images on --prune```

kwilczynski:
> /release-note-edit `crictl rmi will not remove pinned images on --prune`

@saschagrunert, this is probably something only you could do. Unless it's too late already.

saschagrunert:
> > /release-note-edit `crictl rmi will not remove pinned images on --prune`
> 
> @saschagrunert, this is probably something only you could do. Unless it's too late already.

Done :+1:

wangyuan0916:
Is there any plan to cherry pick this fix?/

kwilczynski:
> Is there any plan to cherry pick this fix?/

@wangyuan0916, cherry-pick to which version?

You should be able to pick up latest version of the CLI without any issues, it should work with older versions of CRI-O and such, given that we implement the specification and such.

What issues are you having?

wangyuan0916:
Hi @kwilczynski , thank you for your reply. We are in kubernetes 1.28.x. I check the cri-tools README about the Compatibility matrix and find that the kubernetes and cri-tool minor versions should be the same. So can I use the latest cri-tool with kubernetes 1.28.x? Or is it better to get this fix backported?

kwilczynski:
[...]
> So can I use the latest cri-tool with kubernetes 1.28.x? Or is it better to get this fix backported?

@wangyuan0916, try. :smile:

If things work, then you get the latest version with all the bug fixes and new features. If there are issues, then we can think of a backport.

PR #1367, opened 2024-02-26
1197: crictl rm to also remove container log files

### LINKED ISSUE #1197: Critctl doesn't delete container log files after container is deleted
I have created a container within an openshift pod, this leads to a log file created within directory
/var/log/pods/open_shift_pod_name/contianer_name
I see this file is created by crictl. But when I delete container via crictl the log file remains. When I do the same with another orchestrator the log files is deleted accordingly

fformica:
#### What type of PR is this?
/kind feature

Container log files are not deleted by the runtime, but the user might want to remove them alongside the container.
This PR changes the default behaviour to removing the logs and introduces a new option to the `rm` command to preserve the container log file and its rotations.
Failing to delete files doesn't fail the command and is only logged.

Fixes #1197 

```release-note
Changed default behaviour in `rm` command to also delete the container's log file.
Added `--keeplogs, -k` option to the `rm` command to preserve logs.
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -329,9 +329,9 @@ var removeContainerCommand = &cli.Command{
 			Usage:   "Force removal of the container, disregarding if running",
 		},
 		&cli.BoolFlag{
-			Name:    "deletelog",
-			Aliases: []string{"dl"},
-			Usage:   "Delete the container log file and its rotations as well",
+			Name:    "keep-logs",
+			Aliases: []string{"kl"},
+			Usage:   "Preserve the container log file and its rotations",
 		},
 		&cli.BoolFlag{
 			Name:    "all",
@@ -392,12 +392,12 @@ var removeContainerCommand = &cli.Command{
 				logrus.Errorf("removing container %q failed: %v", id, err)
 				errored = true
 				continue
-			} else if ctx.Bool("deletelog") {
-				logFileName := resp.GetStatus().GetLogPath()
-				if logFileName != "" {
-					logRotations, err := filepath.Glob(logFileName + "*")
+			} else if !ctx.Bool("keep-logs") {
+				logPath := resp.GetStatus().GetLogPath()
+				if logPath != "" {
+					logRotations, err := filepath.Glob(logPath + "*")
 					if err != nil {
-						logRotations = []string{logFileName}
+						logRotations = []string{logPath}
 					}
 
 					for _, logFile := range logRotations {
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -22,6 +22,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
+	"path/filepath"
 	goruntime "runtime"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


kwilczynski:
@fformica, thank you for taking this on! Appreciated.

However, we can have a different spin on the problem at hand.

Generally, most of the time, users _want_ to get everything (all the created artefacts) removed once the container is deleted, including the log files.

This should be the default behaviour, and it would follow the principle of least surprise here.

Now, the different spin I mentioned, would be now having an option to _not_ delete logs, if requested. This would allow users to preserve log files should these be needed for something.

So we would have `crictl rm --keep-logs` or perhaps `crictl rm --preseve-logs` instead here. 

What do you think?

@saschagrunert, thoughts?

saschagrunert:
I agree having it the other way around makes more sense. We just have to double check that we do not break anything in CI, and a highlighted release note that this is now the default behavior should be included.

cmd/crictl/container.go
```diff
@@ -326,6 +328,11 @@ var removeContainerCommand = &cli.Command{
 			Aliases: []string{"f"},
 			Usage:   "Force removal of the container, disregarding if running",
 		},
+		&cli.BoolFlag{
+			Name:    "keep-logs",
+			Aliases: []string{"kl"},
```
saschagrunert, line 333:
```suggestion
			Aliases: []string{"k"},
```

I'm not a huge fan of 2 char aliases because it could also mean `-k -l`.

fformica:
Thanks for the feedback!
Side note, it looks like containerd currently doesn't set logPath, so I'll test that a bit (#1197 was observed on OpenShift).

saschagrunert:
@fformica do we need to change / enhance docs for this?

fformica:
The crictl documentation doesn't explore the options for every command, so I don't think it's necessary to document this separately; I'm not sure though, so let me know if there's anything you think should be updated.

cmd/crictl/container.go
```diff
@@ -385,6 +392,21 @@ var removeContainerCommand = &cli.Command{
 				logrus.Errorf("removing container %q failed: %v", id, err)
 				errored = true
 				continue
+			} else if !ctx.Bool("keep-logs") {
+				logPath := resp.GetStatus().GetLogPath()
+				if logPath != "" {
+					logRotations, err := filepath.Glob(logPath + "*")
```
saschagrunert, line 398: Is this save for the log rotation? We would also possible remove files which are not rotated logs but other files, right?

fformica: It's possible, although the rotations attach to the file name so it's unlikely, e.g. this was from my test with cri-o:
```
-rw------- 1 root root 5.3M Feb 26 05:49 0.log
-rw-r--r-- 1 root root 1.2M Feb 26 05:44 0.log.20240226-053352.gz
-rw------- 1 root root  11M Feb 26 05:44 0.log.20240226-054404
```
Another container would be e.g. 1.log, but a user could have backed up a 0.log.bak, and this code would delete that too. I wasn't sure if the log rotation always follows this format though.
Reading https://kubernetes.io/docs/concepts/cluster-administration/logging/ the rotation in /var/log should only be performed by kubelet, so I could use that as a guide, removing files in this format in that directory and ignoring rotations outside it.

fformica: It looks like kubelet uses a very similar pattern: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/logs/container_log_manager.go#L312
For the time being I'll do the same and just add the missing dot; I'm averse to adding hardcoded references to the dir, especially because it seems a bit more variable under Windows (the doc mentions C:\var\logs and C:\var\log\pods).

kwilczynski:
> Sorry 😭 I only just noticed that by adding the dot I excluded the current log file...

No worries. 👍

Might be good to do some manual testing to confirm things work as expected.

saschagrunert (approved):

PR #1383, opened 2024-03-21
Support websocket transport for exec and attach

saschagrunert:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
This allows end users to choose the transport to be used as well as runtime developers to experiment on feature development.

#### Which issue(s) this PR fixes:

Supersedes: https://github.com/kubernetes-sigs/cri-tools/pull/1305

#### Special notes for your reviewer:
cc @duguhaotian
#### Does this PR introduce a user-facing change?

```release-note
Added `--transport / -r` flag to `crictl exec` and `attach` to select between the "spdy" (default) and "websocket" transport.
```

PR #1385, opened 2024-03-25
Add `--transport` option to port-forward

saschagrunert:
#### What type of PR is this?

/kind feature

The latest Kubernetes beta supports using websockets for port-forward. We now introduce a similar option to `attach` and `exec` to be able to select the new transport.

See: https://github.com/kubernetes/enhancements/issues/4006

Follow-up on https://github.com/kubernetes-sigs/cri-tools/pull/1383

None

```release-note
Added `--transport / -r` flag to `crictl port-forward` and `attach` to select between the "spdy" (default) and "websocket" transport.
```


saschagrunert:

cmd/crictl/portforward.go
```diff
@@ -36,6 +37,14 @@ var runtimePortForwardCommand = &cli.Command{
 	Name:      "port-forward",
 	Usage:     "Forward local port to a pod",
 	ArgsUsage: "POD-ID [LOCAL_PORT:]REMOTE_PORT",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    transportFlag,
+			Aliases: []string{"r"},
```
kwilczynski, line 43: Was `r` chosen here because `t` was already taken?

cmd/crictl/portforward.go
```diff
@@ -36,6 +37,14 @@ var runtimePortForwardCommand = &cli.Command{
...
+			Value:   transportSpdy,
+			Usage:   fmt.Sprintf("Transport protocol to be used, must be one of: %s, %s", transportSpdy, transportWebsocket),
```
kwilczynski, line 45: To align with the other help formats.

```suggestion
			Usage:   fmt.Sprintf("Transport protocol to use, can be one of: %s|%s", transportSpdy, transportWebsocket),
```

Even closer to the format:

```suggestion
			Usage:   fmt.Sprintf("Transport protocol to use, one of: %s|%s", transportSpdy, transportWebsocket),
```


Also, moved the "must" to "can" as there is a default set, which makes this option optional.

cmd/crictl/portforward.go
saschagrunert: Yes, on `exec` and `attach`.

cmd/crictl/portforward.go
saschagrunert: Fixed as suggested :+1:

cmd/crictl/portforward.go
kwilczynski: Ah bummer.

SergeyKanzhelev (approved):
/lgtm

PR #1435, opened 2024-05-23
Gomod tools

tallclair:
#### What type of PR is this?
/kind cleanup

Explicitly version the tools used in the `hack/` directory. This helps with supply chain provenance.

See https://github.com/kubernetes/kubernetes/tree/master/hack/tools for a similar pattern example.

```release-note
NONE
```


hack/tools/go.mod
```diff
@@ -0,0 +1,170 @@
...
+	k8s.io/kubernetes v1.27.13 // indirect
+	k8s.io/utils v0.0.0-20230209194617-a36077c30491 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.1.2 // indirect
+	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
+// Replace directives copied from
+// https://github.com/Mirantis/cri-dockerd/blob/683f70f69901e66d49dfac802841ff843171f131/go.mod#L150-L177
```
tallclair, line 143: Note to reviewers: this is needed because the cri-dockerd go.mod file mixes different k8s versions, and then overrides them in a replace directive like this. The replace directive is required to normalize the versions and ensure it builds.

saschagrunert (approved):

PR #1448, opened 2024-06-06
Add `--pull-timeout` flag to `crictl` `create`, `run` and `pull` commands

saschagrunert:
#### What type of PR is this?

/kind feature

This allows to set a pull context timeout for the supported commands. Runtimes may or may not use the timeout from the RPC for image pulls, but `crictl` should generally support that feature.

None
None

```release-note
Added `--pull-timeout` flag to `crictl` `create`, `run` and `pull` commands.
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -85,6 +85,9 @@ type pullOptions struct {
 	// Username to use for accessing the registry
 	// password will be requested on the command line
 	username string
+
+	// timeout is the maximum time used for the image pull
+	timeout time.Duration
 }
 
 var createPullFlags = []cli.Flag{
@@ -110,6 +113,17 @@ var createPullFlags = []cli.Flag{
 		Name:  "username",
 		Value: "",
 		Usage: "Use `USERNAME` for accessing the registry. The password will be requested on the command line",
+	},
+	&cli.DurationFlag{
+		Name:    "cancel-timeout",
+		Aliases: []string{"T"},
+		Usage:   "Seconds to wait for a container create request to complete before cancelling the request",
+	},
+	&cli.DurationFlag{
+		Name:    "pull-timeout",
+		Aliases: []string{"pt"},
+		Usage:   "Maximum time to be used for pulling the image, disabled if set to 0s",
+		EnvVars: []string{"CRICTL_PULL_TIMEOUT"},
 	},
 }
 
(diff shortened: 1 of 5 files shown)


haircommander:
LGTM

SergeyKanzhelev:
> Runtimes may or may not use the timeout from the RPC for image pulls

Should we make it a part of critest? Not sure how we will test the behavior, but at least runtime should stop processing and exit when requested.

cmd/crictl/image.go
```diff
@@ -633,7 +641,7 @@ func normalizeRepoDigest(repoDigests []string) (string, string) {
 
 // PullImageWithSandbox sends a PullImageRequest to the server, and parses
 // the returned PullImageResponse.
-func PullImageWithSandbox(client internalapi.ImageManagerService, image string, auth *pb.AuthConfig, sandbox *pb.PodSandboxConfig, ann map[string]string) (*pb.PullImageResponse, error) {
+func PullImageWithSandbox(client internalapi.ImageManagerService, image string, auth *pb.AuthConfig, sandbox *pb.PodSandboxConfig, ann map[string]string, timeout time.Duration) (*pb.PullImageResponse, error) {
```
SergeyKanzhelev, line 644: I wonder what should be a behavior on `Ctrl+C`. Should we have a context cancelled as well?

cmd/crictl/container.go
```diff
@@ -111,6 +114,17 @@ var createPullFlags = []cli.Flag{
 		Value: "",
 		Usage: "Use `USERNAME` for accessing the registry. The password will be requested on the command line",
 	},
+	&cli.DurationFlag{
+		Name:    "cancel-timeout",
+		Aliases: []string{"T"},
+		Usage:   "Seconds to wait for a container create request to complete before cancelling the request",
```
SergeyKanzhelev, line 120: should it say - "disabled if 0" as well?

SergeyKanzhelev: or anything about the default

cmd/crictl/container.go
```diff
@@ -111,6 +114,17 @@ var createPullFlags = []cli.Flag{
 		Value: "",
 		Usage: "Use `USERNAME` for accessing the registry. The password will be requested on the command line",
 	},
+	&cli.DurationFlag{
```
SergeyKanzhelev, line 117: I wonder if we need two of those or just one parameter `timeout` will be enough? Are those two having different semantics?

cmd/crictl/container.go
saschagrunert: The help mentions the default value:

```
   --pull-timeout value, --pt value                               Maximum time to be used for pulling the image, disabled if set to 0s (default: 0s) [$CRICTL_PULL_TIMEOUT]
```

cmd/crictl/image.go
saschagrunert: Interesting idea, I don't think we consider Ctrl+C in any RPC for now (means the RPC will run through but we will not catch the result). We can probably follow-up on that and add it to all commands.

saschagrunert:
> > Runtimes may or may not use the timeout from the RPC for image pulls
> 
> Should we make it a part of critest? Not sure how we will test the behavior, but at least runtime should stop processing and exit when requested.

Yes, I added a critest case for that.

SergeyKanzhelev (approved):
/lgtm
/approve

PR #1458, opened 2024-06-13
`crictl run`: Rename `--timeout`/`-t` to `--cancel-timeout`/`-T`

saschagrunert:
#### What type of PR is this?

/kind feature

This streamlines the behavior with `crictl create` by changing the `--timeout` to `--cancel-timeout`.

The timeout flag from the root command (`crictl -t`) can still be used as fallback.

None

_**This is a breaking CLI change. I assume we have no API guarantee and can do that, so please chime in if you think that's wrong.**_ 

```release-note
Renamed `--timeout`/`-t` flag to `--cancel-timeout`/`-T` for `crictl run` to streamline the behavior with `crictl create`.
```


SergeyKanzhelev (approved):
/lgtm

PR #1472, opened 2024-07-02
Fix invalid context on `Status` RPC

saschagrunert:
#### What type of PR is this?

/kind bug

We need to pass the context from the `InterruptableRPC`, otherwise it won't be interruptable.

None
None

```release-note
None
```


kwilczynski:
Nice!

PR #1489, opened 2024-07-05
userns: Skip tests if the host doesn't support idmap mounts

rata:
I'm testing containerd with runc 1.2.0-rc.2 and I found this issue. This (and other fixes that belong to the containerd repo) make the CI happy again.

It will be great if we can include this in a 1.30.1 crit-tools release, so we can update it in the containerd CI. We need to prepare for the runc 1.2.0 final release, so we can just update whenever it is released :).

The containerd PR is this: https://github.com/containerd/containerd/pull/9313. I did a hack (use my fork) to test the fixes indeed fix everything on containerd CI. It does, see here: https://github.com/containerd/containerd/actions/runs/9808474525/job/27084476978?pr=9313.

Without this commit, it fails on AlmaLinux 8 saying the syscall is not implemented: https://github.com/containerd/containerd/actions/runs/9797348691/job/27053863516?pr=9313#step:9:203

Is it possible to release cri-tools 1.30.1 soon with this change?

cc @saschagrunert 

---

critest is used in projects like containerd, that test against older distros (like AlmaLinux 8). In those distros, CI will fail when we upgrade to runc 1.2.0.

With runc 1.1 those test don't fail because runc doesn't support idmap mounts and the tests are skipped in that case. But with runc 1.2.0-rc.2, that supports idmap mounts, the tests are not skipped but fail on distros with older kernels that don't support idmap mounts.

This commit just tries to detect if the path used for the container rootfs supports idmap mounts. To do that it uses the Status() message from CRI with verbose param set to true. It parses the output that containerd sets (it's quite unspecified that field), and otherwise fallbacks to "/var/lib" as the path to test idmap mounts support.

#### What type of PR is this?
/kind bug
/kind failing-test

We need it to fix the CI in containerd, once we update runc to 1.2.
It might be needed here too, once runc 1.2 is used here

None

```release-note
Skip tests with idmap mounts on host that don't support it
```

### DIFF
--- a/pkg/validate/security_context_linux.go
+++ b/pkg/validate/security_context_linux.go
@@ -18,14 +18,17 @@ package validate
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
+	"golang.org/x/sys/unix"
 	internalapi "k8s.io/cri-api/pkg/apis"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
 	"sigs.k8s.io/cri-tools/pkg/common"
@@ -38,6 +41,8 @@ import (
 const (
 	nginxContainerImage string = framework.DefaultRegistryE2ETestImagesPrefix + "nginx:1.14-2"
 	noNewPrivsImage     string = framework.DefaultRegistryE2ETestImagesPrefix + "nonewprivs:1.3"
+	usernsSize          int    = 65536
+	usernsHostID        int    = 65536
 )
 
 var _ = framework.KubeDescribe("Security Context", func() {
@@ -860,7 +865,7 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			By("searching for runtime handler which supports user namespaces")
 			ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 			defer cancel()
-			resp, err := rc.Status(ctx, false)
+			resp, err := rc.Status(ctx, true) // Set verbose to true so the info field is populated.
 			framework.ExpectNoError(err, "failed to get runtime config: %v", err)
 
 			supportsUserNamespaces := false
@@ -875,6 +880,11 @@ var _ = framework.KubeDescribe("Security Context", func() {
 
 			if !supportsUserNamespaces {
 				Skip("no runtime handler found which supports user namespaces")
+			}
+
+			pathIDMap := rootfsPath(resp.GetInfo())
+			if err := supportsIDMap(pathIDMap); err != nil {
+				Skip("ID mapping is not supported" + " with path: " + pathIDMap + ": " + err.Error())
 			}
 		})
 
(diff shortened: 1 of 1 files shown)


pkg/validate/security_context_linux.go
```diff
@@ -876,6 +879,11 @@ var _ = framework.KubeDescribe("Security Context", func() {
 			if !supportsUserNamespaces {
 				Skip("no runtime handler found which supports user namespaces")
 			}
+
+			pathIDMap := rootfsPath(resp.GetInfo())
+			if err := supportsIDMap(pathIDMap); err != nil {
```
haircommander, line 884: why don't we use the [runtime handler features](https://github.com/kubernetes-sigs/cri-tools/pull/1489/files#diff-009f9008c3866ca7eb564aa00ef9855c8c612b5c19b293909c56d44aac667779R870) feature? shouldn't containerd not report the feature is available if it's not?

rata: The runtime handler features reports if the OCI runtime knows about idmap mounts and how to handle them. The thing is, OCI spec mandates that unrecognized lines are ignored. So, an old OCI runtime that doesn't know about idmap mounts, will just ignore the lines "uidMapping" and so on the mounts.

The end result if that happens is horrible, because there is no mapping applied and then you use the volume with the hostUID (the ID on the host the container is mapped to) and then create files. The hostUID changes every time you create the pod, so the volume will be with garbage in the UID/GID of files.

This is why we added to the runtime-spec the way to report in features if idmap mounts are supported: https://github.com/opencontainers/runtime-spec/pull/1219

That is to avoid that situation: you just don't give idmap mounts to runtime that don't know about it. Because they will ignore it and create the aforementioned mess.

So once you know the runtime will handle it properly (using the features you mentioned), you can pass it to the runtime. But idmap mount support is filesystem-dependant. That means, each filesystem used needs to support it. And more filesystems are adding support every kernel release.

So you send the request to the OCI runtime to use idmap mounts on some volumes, but the runtime can return an error if the given filesystem doesn't support idmap mounts on the kernel it is running.

In particular in AlmaLinux 8 what is happening is: runc is new enough to handle them correctly, so we can pass a config.json with idmap mounts to that runc version. So we do. But the kernel is old and doesn't support the syscalls needed (so it doesn't support idmap mounts), so runc returns an error when trying to set it up.

What this PR does is just testing for idmap mounts supports and, if it is not supported, it skips the test.

tl; dr: the features command is useful to know if the OCI runtime knows about idmap mounts at compile time (see more info on the problem it solves here: https://github.com/opencontainers/runtime-spec/pull/1219). However, we can run runc in filesystems that don't support idmap mounts (one fs can support it but another don't), and runc will return an error in those cases. We just skip the tests instead, to avoid the expected error.

rata: Maybe the shorter answer is: containerd can only report it will be used or fail, but not ignored (as it was before if containerd/runc didn't know about idmap mounts and userns).

But each file-system used by mounts on the container needs to support idmap mounts. So it can't report yes/no without knowing which pod (e.g. if the pod users NFS, then it will fail, but if only uses ext4 and tmpfs, it will work on lot of kernels).

pkg/validate/security_context_linux.go
```diff
@@ -1458,3 +1466,74 @@ func runUserNamespacePodWithError(
...
+
+	// We want to test if idmap mounts are supported.
+	// So we use just some random mapping, it doesn't really matter which one.
+	// For the helper command, we just need something that is alive while we
+	// test this, a sleep 5 will do it.
+	cmd := exec.Command("sleep", "5")
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Cloneflags:  syscall.CLONE_NEWUSER,
+		UidMappings: []syscall.SysProcIDMap{{ContainerID: 0, HostID: 65536, Size: 65536}},
+		GidMappings: []syscall.SysProcIDMap{{ContainerID: 0, HostID: 65536, Size: 65536}},
```
saschagrunert, line 1485: Let's make `65536` a const.

pkg/validate/security_context_linux.go
saschagrunert, line 884: Should this run once `BeforeAll` user namespace tests?

pkg/validate/security_context_linux.go
rata: Fixed, thanks!

pkg/validate/security_context_linux.go
rata: We can't add BeforeAll() for not ordered containers. And we don't need them to be ordered, it will slow it down for no reason.

I was reading the ginkgo manual, I think it makes sense to do the following:
1. Add a `When("Host idmap mount support is needed")` and move the tests that need it under that
2. Add a BeforeEach for that, that checks for idmap mount support
3. Add a `When("Host idmap mount support is not needed")` and move the tests that don't need it there (like multiple mappings and all, idmap mounts is not needed, validation is expected to fail).
4. Make the detection of the runtime handler support for userns inside a sync.Once, so we just query the runtime once and work with the answer later.

I'll push it in a few minutes

Let me know what you think :)

saschagrunert: Sounds good to me, thank you!

rata: Pushed this now

rata: If you see the commit diff with `git show -w`, it will be cleaner. It seems bigger, but it is only nesting some tests under the `When("Host idmap mount support is not needed")`

rata: I'll re-validate this on the containerd side, just in case. I'll let you know when it is ready for review

rata: While doing that, I found there is a test that needs more care, as containerd runs the tests inside a userns and the whole UID space was hardcoded and that test failed.

I'll double check tomorrow all is fine for that case before asking for another review.

haircommander: what file system are we trying to create a mount here? if runc thinks alma support idmapped mounts I'd expect all the mounts we create to work. like, we control what environments we run these tests in. We could make sure they're not nfs but instead ext4 or tmpfs (whatever the alma kernel supports at that time). What file systems are these tests running on that are failing, but runc is saying it's supporteD?

saschagrunert (approved):

rata:
The test fail was unrelated to this, some github issue:

```
fatal: unable to access 'https://github.com/containernetworking/plugins.git/': Failed to connect to github.com port 443 after 21 ms: Connection refused
```

I pushed again due to that.

@saschagrunert thanks a lot! I polished the code and fixed some edge cases now. I've also verified it passes all the tests in the containerd CI too(https://github.com/containerd/containerd/actions/runs/9859447900/job/27223414193?pr=9313) :)

saschagrunert (approved):

saschagrunert:
I created `release-1.30` and a corresponding cherry-pick in #1492

PR #1496, opened 2024-07-16
Enable more linters and update config

saschagrunert:
#### What type of PR is this?

/kind cleanup

Update the linter config and fix a bunch of lints.

None
None

```release-note
None
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,22 +6,39 @@ linters:
   disable-all: true
   enable:
     - asasalint
+    - asciicheck
     - bidichk
+    - bodyclose
     - canonicalheader
     - containedctx
+    - contextcheck
     - copyloopvar
     - decorder
+    - dogsled
     - dupl
     - dupword
     - durationcheck
+    - errchkjson
     - errname
+    - errorlint
+    - exhaustive
+    - exportloopref
     - fatcontext
     - forcetypeassert
+    - gci
+    - ginkgolinter
     - gocheckcompilerdirectives
     - gochecksumtype
     - gocritic
+    - godot
     - gofmt
+    - gofumpt
+    - goheader
     - goimports
+    - gomoddirectives
+    - gomodguard
+    - goprintffuncname
+    - gosimple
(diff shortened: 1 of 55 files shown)


saschagrunert:

cmd/crictl/config.go
```diff
@@ -106,7 +106,7 @@ CRICTL OPTIONS:
 			}
 			return common.WriteConfig(config, configFile)
 		} else if c.Bool("list") {
-			display := newTableDisplay(20, 1, 3, ' ', 0)
```
SergeyKanzhelev, line 109: I wonder what linter rule did it break?

.golangci.yml
```diff
@@ -38,81 +55,69 @@ linters:
...
+    - noctx
+    - nolintlint
     - nosprintfhostport
     - perfsprint
     - prealloc
     - predeclared
     - promlinter
     - protogetter
     - reassign
+    - rowserrcheck
```
kwilczynski, line 67: Not sure if we need to enable this one?

.golangci.yml
```diff
@@ -38,81 +55,69 @@ linters:
...
     - sloglint
     - spancheck
+    - sqlclosecheck
```
kwilczynski, line 70: Ditto as https://github.com/kubernetes-sigs/cri-tools/pull/1496/files#r1680376638.

saschagrunert: I'd say leave it enabled if we either add code which the linter may catch in the future.

.golangci.yml
saschagrunert: I'd say leave it enabled if we either add code which the linter may catch in the future.

cmd/crictl/config.go
saschagrunert: It's the [gosimple](https://golangci-lint.run/usage/linters/#gosimple) linter, but I could work around by introducing a `newDefaultTableDisplay` method and reusing it everywhere where it's needed.

PR #1503, opened 2024-07-17
WIP: Check AppArmor availability without vendored runc

saschagrunert:
#### What type of PR is this?

/kind cleanup

This allows getting away from the vendored package by refactoring the AppArmor related tests.

None
None

```release-note
None
```


pkg/validate/apparmor_linux.go
```diff
@@ -59,157 +59,172 @@ profile cri-validate-apparmor-test-audit-write flags=(attach_disconnected) {
...
+		checkAppArmorOnce sync.Once
+		appArmorEnabled   bool
+	)
+
+	appArmorIsEnabled := func() bool {
+		checkAppArmorOnce.Do(func() {
+			if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil {
+				buf, err := os.ReadFile("/sys/module/apparmor/parameters/enabled")
+				appArmorEnabled = err == nil && len(buf) > 1 && buf[0] == 'Y'
+			}
```
kwilczynski, line 74: You can reduce the check to just reading from the sysfs object.

If you get anything other than "Y", then AppArmor is not enabled. If you get an I/O error of any sort, especially `-ENOENT` (albeit, any I/O error would be sufficient, since this is sysfs object has permissions of 444, so anyone should be able to read it should it exists), then AppArmor is not built in the kernel.

kwilczynski:
Very nice!

/approve
/lgtm

PR #1535, opened 2024-07-30
Enable and fix `errcheck` linter

saschagrunert:
#### What type of PR is this?

/kind cleanup

Enable the linter and fix the reports.

None
None

```release-note
None
```

### DIFF
--- a/cmd/crictl/version.go
+++ b/cmd/crictl/version.go
@@ -32,7 +32,7 @@ var runtimeVersionCommand = &cli.Command{
 	Usage: "Display runtime version information",
 	Action: func(c *cli.Context) error {
 		if c.NArg() != 0 {
-			cli.ShowSubcommandHelp(c)
+			return cli.ShowSubcommandHelp(c)
 		}
 
 		runtimeClient, err := getRuntimeService(c, 0)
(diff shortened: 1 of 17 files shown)


cmd/crictl/version.go
```diff
@@ -32,7 +32,7 @@ var runtimeVersionCommand = &cli.Command{
 	Usage: "Display runtime version information",
 	Action: func(c *cli.Context) error {
 		if c.NArg() != 0 {
-			cli.ShowSubcommandHelp(c)
+			return cli.ShowSubcommandHelp(c)
```
saschagrunert, line 35: Another strange misbehavior because right now it still tries to connect even when showing the help.

pkg/validate/pod.go
```diff
@@ -75,6 +77,7 @@ var _ = framework.KubeDescribe("PodSandbox", func() {
 
 			By("test remove PodSandbox")
 			testRemovePodSandbox(rc, podID)
+			podID = "" // no need to cleanup pod
```
saschagrunert, line 80: Needs to be done because containerd does error on sandbox stop if the pod is not found: https://github.com/containerd/containerd/pull/10520

PR #1545, opened 2024-08-01
Add more CRI-O matrix tests

saschagrunert:
#### What type of PR is this?

/kind cleanup

Tweak the CRI-O config to allow using runc instead of crun as well as conmon-rs instead of conmon.

None
None

```release-note
None
```

### DIFF
--- a/.github/workflows/crio.yml
+++ b/.github/workflows/crio.yml
@@ -9,34 +9,27 @@ on:
 jobs:
   test:
     strategy:
+      fail-fast: false
       matrix:
         suite:
           - e2e
           - critest
-    name: ${{matrix.suite}}
+        oci-runtime:
+          - crun
+          - runc
+        monitor:
+          - conmon
+          - conmon-rs
+    name: ${{matrix.suite}} / ${{matrix.oci-runtime}} / ${{matrix.monitor}}
     runs-on: ubuntu-22.04
     steps:
+      - name: Checkout cri-tools
+        uses: actions/checkout@v4
+
       - name: Install go
         uses: actions/setup-go@v5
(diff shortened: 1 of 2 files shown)


.github/workflows/crio.yml
```diff
@@ -55,28 +48,38 @@ jobs:
...
+      - name: Configure CRI-O
         run: |
           sudo mkdir -p /etc/crio/crio.conf.d
           printf '[crio.runtime]\nlog_level = "debug"\n' | sudo tee /etc/crio/crio.conf.d/01-log-level.conf
-          printf '[crio.runtime]\nseccomp_use_default_when_empty = false\n' | sudo tee /etc/crio/crio.conf.d/02-seccomp.conf
+
+      - name: Configure CRI-O to use conmon-rs intead of the default conmon
+        if: ${{matrix.monitor == 'conmon-rs'}}
+        run: |
+          sudo sed -i -E 's;(monitor_path = ).*;\1"/usr/libexec/crio/conmonrs"\nruntime_type = "pod";g' /etc/crio/crio.conf.d/10-crio.conf
```
kwilczynski, line 63: I wonder whether it might be simpler to drop a drop-in file called `99-crio.conf` that has the properties we want to override.

Thoughts?

saschagrunert: It depends, that way I can just change one value by relying on the shipped defaults and easily verify everything manually in the "Show the CRI-O config drop-in" step. I think in the end it doesn't matter much.

saschagrunert:
Ready for review. PTAL @kubernetes-sigs/cri-tools-maintainers

PR #1553, opened 2024-08-06
Improve `crictl inspect[pi]` commands to allow filtering

saschagrunert:
#### What type of PR is this?

/kind feature

The commands now allow filtering based on various fields or just to inspect all pods, containers or images.

None

None

```release-note
Added filtering flags for `crictl` `inspect`, `inspectp` and `inspecti`. If no filter is provided, then the commands will now inspect all containers, pods or images.
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -502,19 +502,78 @@ var containerStatusCommand = &cli.Command{
 			Name:  "template",
 			Usage: "The template string is only used when output is go-template; The Template format is golang template",
 		},
+		&cli.StringFlag{
+			Name:  "name",
+			Usage: "Filter by container name regular expression pattern",
+		},
+		&cli.StringFlag{
+			Name:    "pod",
+			Aliases: []string{"p"},
+			Usage:   "Filter by pod id",
+		},
+		&cli.StringFlag{
+			Name:  "image",
+			Usage: "Filter by container image",
+		},
+		&cli.StringFlag{
+			Name:    "state",
+			Aliases: []string{"s"},
+			Usage:   "Filter by container state",
+		},
+		&cli.StringSliceFlag{
+			Name:  "label",
+			Usage: "Filter by key=value label",
+		},
+		&cli.BoolFlag{
+			Name:    "latest",
+			Aliases: []string{"l"},
+			Usage:   "Show the most recently created container (includes all states)",
+		},
+		&cli.IntFlag{
+			Name:    "last",
+			Aliases: []string{"n"},
+			Usage:   "Show last n recently created containers (includes all states). Set 0 for unlimited.",
+		},
 	},
 	Action: func(c *cli.Context) error {
-		if c.NArg() == 0 {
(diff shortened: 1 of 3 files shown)


cmd/crictl/container.go
```diff
@@ -508,19 +508,82 @@ var containerStatusCommand = &cli.Command{
 			Name:  "template",
 			Usage: "The template string is only used when output is go-template; The Template format is golang template",
 		},
+		&cli.StringFlag{
+			Name:  "name",
+			Value: "",
```
kwilczynski, line 513: Isn't an empty string the default value? It would be for a string variable.

saschagrunert: Yeah that's a copy paste leftover, fixed as suggested.

saschagrunert:

cmd/crictl/image.go
```diff
@@ -320,10 +317,25 @@ var imageStatusCommand = &cli.Command{
...
-			id := c.Args().Get(i)
+		ids := c.Args().Slice()
+
+		if len(ids) == 0 {
+			r, err := ListImages(imageClient, c.String("name"), c.StringSlice("filter"))
+			if err != nil {
+				return fmt.Errorf("listing images: %w", err)
+			}
+			for _, img := range r.GetImages() {
+				ids = append(ids, img.GetId())
```
SergeyKanzhelev, line 328: do we need to deduplicate here?

cmd/crictl/sandbox.go
```diff
@@ -322,7 +374,7 @@ var listPodCommand = &cli.Command{
 		if err != nil {
 			return err
 		}
-		if err = ListPodSandboxes(runtimeClient, opts); err != nil {
+		if err = OutputPodSandboxes(runtimeClient, opts); err != nil {
```
SergeyKanzhelev, line 377: if filtered by reference, what image reference will the output use as a main one? My understanding is that even with the filter, the output may show  different reference as a "main" name of an image and tag one filtered by in a list of alternative names. Is it right? I wonder this is actually the case and whether we can either improve output or make it clear in docs so not to confuse users

SergeyKanzhelev, line 377: in case of various filters, potential improvement over time may be to show filter-specific additional column. For since it may be time passed since that value in the filter, for example. This is just a nit.

cmd/crictl/image.go
saschagrunert: I'd say no, because right now it behaves in the same way to output two images when doing something like:

```
sudo crictl inspecti 324bc02ae1231 324bc02ae1231
```

cmd/crictl/sandbox.go
saschagrunert: Hm, pods are not filterable by reference, are we speaking about `crictl inspecti`? My main goal was to mimic closely as possible the existing `crictl` `pods`, `ps` and `images` subcommand flags to avoid confusion.

SergeyKanzhelev (approved):
/lgtm

PR #1603, opened 2024-09-20
crictl exec: allow running single command in multiple containers

saschagrunert:
#### What type of PR is this?

/kind feature

This patch adds the following filter flags to `crictl exec`:

```
--image value                    Exec command for all containers matching this container image filter
--label value [ --label value ]  Exec command for all containers matching this key=value label filter
--last value, -n value           Exec command for all last n containers, set to 0 for unlimited (default: 0)
--latest, -l                     Exec command for the most recently created container (default: false)
--name value                     Exec command for all containers matching this name filter regular expression pattern
--pod value, -p value            Exec command for all containers matching this pod ID filter
--state value                    Exec command for all containers matching this container state filter
```

If none of the above filters are set, then crictl will assume that the first provided argument is the container ID. If one of them is set, then crictl will filter and execute for all containers:

```
> sudo ./build/bin/linux/amd64/crictl exec --name pod ps aux
8f92b841d898828833d7dc55245e5df91c517cb0fce8a61f5def7cb9e432530c:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   2472  1344 ?        Ss   07:09   0:00 /bin/sleep 6000 6000
root          67  0.0  0.0   5536  2700 ?        Rs   07:38   0:00 ps aux
39f78a7b1835b8363a519424031ddbb370b743afad46cc401eeb73717c84e57d:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   2472  1500 ?        Ss   07:08   0:00 /bin/sleep 6000 6000
root         122  0.0  0.0   5536  2732 ?        Rs   07:38   0:00 ps aux
```

The container ID is printed when a filter command has been used to ensure existing backwards compatibility.

None

None

```release-note
crictl exec: allow running single command in multiple containers by using one of the new `--image`, `--label`, `--last`, `--latest`, `--name`, `--pod` or `--state` flags.
```

### DIFF
--- a/cmd/crictl/exec.go
+++ b/cmd/crictl/exec.go
@@ -41,14 +41,16 @@ const (
 	transportFlag      = "transport"
 	transportWebsocket = "websocket"
 	transportSpdy      = "spdy"
+
+	detachSequence = "ctrl-p,ctrl-q"
 )
 
-const detachSequence = "ctrl-p,ctrl-q"
-
 var runtimeExecCommand = &cli.Command{
-	Name:                   "exec",
-	Usage:                  "Run a command in a running container",
-	ArgsUsage:              "CONTAINER-ID COMMAND [ARG...]",
+	Name:      "exec",
+	Usage:     "Run a command in a running container",
+	ArgsUsage: "[CONTAINER-ID] COMMAND [ARG...]",
+	Description: `The CONTAINER-ID is only required if none of the following filter flags are set:
+--image, --label, --last, --latest, --name, --pod, --state`,
 	UseShortOptionHandling: true,
 	Flags: []cli.Flag{
 		&cli.BoolFlag{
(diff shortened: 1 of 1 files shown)


saschagrunert:

kwilczynski:
What use case does this solve?

saschagrunert:
> What use case does this solve?

Having a set of containers where the same command should be executed in, mostly for debugging and testing purposes.

kwilczynski:
> > What use case does this solve?
> 
> Having a set of containers where the same command should be executed in, mostly for debugging and testing purposes.

**An idea:**

Can we add an option, a command-line switch perhaps, to run whatever command has been given in parallel. This would take care of the use case, replacing the need to have either the `xargs` or `parallel` binaries available. Something that I often rely on while testing or troubleshooting something.

An ability to retry and ignore errors would also be nice.

saschagrunert:
> **An idea:**
> 
> Can we add an option, a command-line switch perhaps, to run whatever command has been given in parallel. This would take care of the use case, replacing the need to have either the `xargs` or `parallel` binaries available. Something that I often rely on while testing or troubleshooting something.
> 
> An ability to retry and ignore errors would also be nice.

Sounds like great ideas! Would that work as follow-up on that PR?

kwilczynski:
> > **An idea:**
> > Can we add an option, a command-line switch perhaps, to run whatever command has been given in parallel. This would take care of the use case, replacing the need to have either the `xargs` or `parallel` binaries available. Something that I often rely on while testing or troubleshooting something.
> > An ability to retry and ignore errors would also be nice.
> 
> Sounds like great ideas! Would that work as follow-up on that PR?

Yeah, sure.

PR #1622, opened 2024-10-08
crictl exec: add `--quiet/-q`, `--ignore-error/-e` and `--parallel` flags

saschagrunert:
#### What type of PR is this?

/kind feature

The flags can be used to further manipulate on the exec behavior. Follow-up on: https://github.com/kubernetes-sigs/cri-tools/pull/1603#issuecomment-2396328240

None

None

```release-note
crictl exec: add `--quiet/-q`, `--ignore-error/-e` and `--parallel/-x` flags.
```

### DIFF
--- a/cmd/crictl/exec.go
+++ b/cmd/crictl/exec.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"runtime"
+	"sync"
 	"time"
 
 	mobyterm "github.com/moby/term"
@@ -109,6 +111,21 @@ var runtimeExecCommand = &cli.Command{
 			Aliases: []string{"n"},
 			Usage:   "Exec command for all last n containers, set to 0 for unlimited",
 		},
+		&cli.BoolFlag{
+			Name:    "quiet",
+			Aliases: []string{"q"},
+			Usage:   "Do not print the container ID if multiple containers are selected",
+		},
+		&cli.BoolFlag{
+			Name:    "ignore-errors",
+			Aliases: []string{"e"},
+			Usage:   "Ignore errors during command execution",
+		},
+		&cli.BoolFlag{
+			Name:    "parallel",
+			Aliases: []string{"x"},
+			Usage:   "Run the command in parallel if multiple containers are selected",
+		},
 	},
 	Action: func(c *cli.Context) error {
 		if c.NArg() < 1 {
@@ -124,6 +141,12 @@ var runtimeExecCommand = &cli.Command{
 		ids := []string{c.Args().First()}
 		cmd := c.Args().Slice()[1:]
 		outputContainerID := false
+		ignoreErrors := c.Bool("ignore-errors")
+		quiet := c.Bool("quiet")
+
+		if ignoreErrors && !quiet {
+			logrus.Warn("The `--ignore-errors` option has been set all errors returned by the command will be ignored.")
+		}
 
 		// If any of the filter flags are set, then we assume that no
 		// CONTAINER-ID is provided as CLI parameter.
@@ -137,7 +160,7 @@ var runtimeExecCommand = &cli.Command{
 
 			ids = []string{}
 			cmd = c.Args().Slice()
-			outputContainerID = true
+			outputContainerID = !quiet && !c.Bool("parallel")
 
 			opts := &listOptions{
 				nameRegexp: c.String("name"),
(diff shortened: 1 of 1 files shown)


cmd/crictl/exec.go
```diff
@@ -178,37 +194,69 @@ var runtimeExecCommand = &cli.Command{
...
-			defer cancel()
-			err = Exec(ctx, runtimeClient, opts)
-			if err != nil {
-				return fmt.Errorf("execing command in container: %w", err)
-			}
+			return nil
+		})
 
+		if c.Bool("ignore-error") {
+			return nil
```
SergeyKanzhelev, line 230: Let's write a warning like "Ignoring errors: errors.Join"

cmd/crictl/exec.go
```diff
@@ -109,6 +111,20 @@ var runtimeExecCommand = &cli.Command{
 			Aliases: []string{"n"},
 			Usage:   "Exec command for all last n containers, set to 0 for unlimited",
 		},
+		&cli.BoolFlag{
+			Name:    "quiet",
```
kwilczynski, line 115: This is great! Makes it easier to pipe the output to something like `grep`, etc.

cmd/crictl/exec.go
```diff
@@ -109,6 +111,20 @@ var runtimeExecCommand = &cli.Command{
...
+			Aliases: []string{"q"},
+			Usage:   "Do not print the container ID if multiple containers are selected",
+		},
+		&cli.BoolFlag{
+			Name:    "ignore-error",
```
kwilczynski, line 120:
```suggestion
			Name:    "ignore-errors",
```

cmd/crictl/exec.go
```diff
@@ -109,6 +111,20 @@ var runtimeExecCommand = &cli.Command{
...
+			Aliases: []string{"e"},
+			Usage:   "Ignore any error during command execution",
```
kwilczynski, line 122:
```suggestion
			Usage:   "Ignore errors during command execution",
```

cmd/crictl/exec.go
kwilczynski: Also, it should be probably somewhere before the main execution loop, to let user know ahead of time.

saschagrunert: Fixed as suggested :+1:

cmd/crictl/exec.go
```diff
@@ -178,37 +196,74 @@ var runtimeExecCommand = &cli.Command{
...
-			}
+			return nil
+		})
+
+		errs := errors.Join(results...)
 
+		if c.Bool("ignore-errors") {
+			if !quiet {
+				logrus.Warnf("Ignoring errors: %v", errs)
+			}
```
kwilczynski, line 236: This can be moved to a debug log, if anything.

Why? The users should be told once, before the loop runs, that errors will be ignored. If someone had run this against 100 or so containers and the command failed (not in the `PATH` or such), then getting the warning that the errors were ignored 100 times is perhaps not so useful.

A small UI/UX change. Not quite the Unix philosophy, though.

saschagrunert: Sounds good, changed to become a debug message.

cmd/crictl/exec.go
```diff
@@ -178,37 +196,74 @@ var runtimeExecCommand = &cli.Command{
...
-			if err != nil {
-				return fmt.Errorf("execing command in container: %w", err)
-			}
+			return nil
+		})
+
+		errs := errors.Join(results...)
 
+		if c.Bool("ignore-errors") {
+			if !quiet {
```
kwilczynski, line 234: This might not be needed given the move to a debug log.

saschagrunert: True, I removed that one.

cmd/crictl/exec.go
kwilczynski: What I meant would be something akin to this:

```console
$ cictl exec -p -x --label='test=123' date
WARN[0000] The `--ignore-errors` option has been set. Errors following command execution will be ignored...
...
...
...
```

Perhaps something with better wording.

cmd/crictl/exec.go
SergeyKanzhelev: Is it a good practice to tell that there were errors?

kwilczynski: @SergeyKanzhelev, we do that by default. :smile: 

However, when the users passes `--ignore-errors`, then we... ignore errors (you can still see these if you are running debug log level, though). I also suggested warning user once at the beginning when this flag is used for clarity: https://github.com/kubernetes-sigs/cri-tools/pull/1622#discussion_r1793095129

Would that work here for this feature?

kwilczynski: Note, the current code is slightly different (updates were made).

cmd/crictl/exec.go
saschagrunert: Got it, added that additional log line.

cmd/crictl/exec.go
```diff
@@ -124,6 +141,11 @@ var runtimeExecCommand = &cli.Command{
 		ids := []string{c.Args().First()}
 		cmd := c.Args().Slice()[1:]
 		outputContainerID := false
+		ignoreErrors := c.Bool("ignore-errors")
+
+		if ignoreErrors {
```
kwilczynski, line 146:
```suggestion
		if ignoreErrors && !c.Bool("quiet") {
```

PR #1643, opened 2024-10-18
Simplify config parsing

saschagrunert:
#### What type of PR is this?

/kind cleanup

We can assume that the config is a flat map of keys and values and manually merge them into the `Config` struct. This allows to avoid manually parsing the YAML.

None
None

```release-note
None
```

### DIFF
--- a/pkg/common/file.go
+++ b/pkg/common/file.go
@@ -19,10 +19,12 @@ package common
 import (
 	"fmt"
 	"os"
-	gofilepath "path/filepath"
-	"strconv"
+	"path/filepath"
+	"reflect"
+	"slices"
 
-	"gopkg.in/yaml.v3"
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
 )
 
 // Config is the internal representation of the yaml that defines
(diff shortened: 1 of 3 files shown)


pkg/common/file.go
```diff
@@ -34,175 +35,123 @@ type Config struct {
...
-			if err != nil {
-				return nil, fmt.Errorf("parsing config option '%s': %w", name, err)
-			}
-		case "disable-pull-on-run":
-			config.DisablePullOnRun, err = strconv.ParseBool(value)
-			if err != nil {
-				return nil, fmt.Errorf("parsing config option '%s': %w", name, err)
-			}
-		default:
-			return nil, fmt.Errorf("Config option '%s' is not valid", name)
```
SergeyKanzhelev, line 125: so there will be no more check for unknown option? Cannot find it now

pkg/common/file.go
```diff
@@ -34,175 +35,123 @@ type Config struct {
...
+	debugKey             = "debug"
+	pullImageOnCreateKey = "pull-image-on-create"
+	disablePullOnRunKey  = "disable-pull-on-run"
+)
+
 // ReadConfig reads from a file with the given name and returns a config or
 // an error if the file was unable to be parsed.
-func ReadConfig(filepath string) (*Config, error) {
-	data, err := os.ReadFile(filepath)
+func ReadConfig(path string) (*Config, error) {
```
SergeyKanzhelev, line 51: maybe easier to review if there are some unit tests for this function.

saschagrunert: I agree, we should add them before this PR.

/hold

SergeyKanzhelev: Also those edge cases: https://github.com/kubernetes-sigs/cri-tools/pull/1646/files#r1811995991

I am curious if parsing to map would have troubles with the duplicate entries and whether it will keep which value will be used.

saschagrunert: Rebased the PR and fixed the logic accordingly to the failing test.s

pkg/common/file.go
```diff
@@ -34,177 +36,147 @@ type Config struct {
...
-	if len(yamlData.Content) == 0 ||
-		yamlData.Content[0].Content == nil {
-		return config, nil
-	}
-	contentLen := len(yamlData.Content[0].Content)
-
-	// YAML representation contains 2 yaml ScalarNodes per config option.
-	// One is config option name and other is the value of the option
-	// These ScalarNodes help preserve comments associated with
-	// the YAML entry
```
SergeyKanzhelev, line 95: I don't understand this comment. Why and where the comments associated with the YAML entry was ever used?

pkg/common/file.go
SergeyKanzhelev: thank you!

pkg/common/file.go
SergeyKanzhelev: besides this comment it is lgtm

saschagrunert: Looks like we cannot preserve the comments without the current logic. Closing this PR for now.

SergeyKanzhelev: I am not sure where we are preserving comments. Aren't config is just readonly?

If we do preserver comments in some scenario - it will be great to have a test for it so it is not just a comment.

saschagrunert: We preserve the comments but not the newlines right now, I added an e2e test scenario for the whole `crictl config` subcommand in https://github.com/kubernetes-sigs/cri-tools/pull/1659

PR #1659, opened 2024-10-25
Add `crictl config` e2e tests

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
Adding e2e tests for the config get / set.
#### Which issue(s) this PR fixes:

Follow-up on https://github.com/kubernetes-sigs/cri-tools/pull/1643
#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #1661, opened 2024-10-28
Fix `crictl config --set` if the YAML defines entries multiple times

saschagrunert:
#### What type of PR is this?

/kind bug

Only the first entry would be written before applying this patch. This means that a config like this:

```yaml
timeout: 5
timeout: 10
timeout: 20
runtime-endpoint: ""
image-endpoint: ""
debug: false
pull-image-on-create: false
disable-pull-on-run: false
```

Would return the right value (the last one):

```
> crictl --config ./crictl.yaml config --get timeout
20
```

But setting the config will only result in setting the first item:

```
> crictl --config ./crictl.yaml config --set timeout=30
> cat crictl.yaml
timeout: 30
timeout: 10
timeout: 20
runtime-endpoint: ""
image-endpoint: ""
debug: false
pull-image-on-create: false
disable-pull-on-run: false
```

And therefore also a wrong result of `crictl config --get`:

```
> crictl --config ./crictl.yaml config --get timeout
20
```

We now change that behavior and set all values without de-duplicating them:

```
> crictl --config ./crictl.yaml config --set timeout=30
> cat crictl.yaml
timeout: 30
timeout: 30
timeout: 30
runtime-endpoint: ""
image-endpoint: ""
debug: false
pull-image-on-create: false
disable-pull-on-run: false
```

e2e test cases around that scenario have been added as well.

None
Follow-up on https://github.com/kubernetes-sigs/cri-tools/pull/1659#discussion_r1817010069

```release-note
Fixed `crictl config --set` if the configuration YAML defines entries multiple times.
```

### DIFF
--- a/pkg/common/file.go
+++ b/pkg/common/file.go
@@ -182,9 +182,9 @@ func setConfigOption(configName, configValue string, yamlData *yaml.Node) {
 	for indx := 0; indx < contentLen-1; {
 		name := yamlData.Content[0].Content[indx].Value
 		if name == configName {
+			// Set the value, even if we have the option defined multiple times.
 			yamlData.Content[0].Content[indx+1].Value = configValue
 			foundOption = true
-			break
 		}
 		indx += 2
 	}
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev (approved):
/lgtm

PR #1662, opened 2024-10-28
Clarify that `crictl config` without args refers to `set`

saschagrunert:
#### What type of PR is this?

/kind documentation

It's not clear to the end user that not specifying `--list`, `--set` or `--get` will result in a configuration set. We now clarify that in the docs as well as logs of the option.

None
None

```release-note
Update `crictl config` documentation to clarify that `crictl config` without args refers to `--set`.
```

### DIFF
--- a/cmd/crictl/config.go
+++ b/cmd/crictl/config.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev (approved):
/lgtm

PR #1681, opened 2024-11-11
Use `slices.Compact` instead of own logic

saschagrunert:
#### What type of PR is this?

/kind cleanup

The `slices` package can be used to remove duplicates.

None
None

```release-note
None
```

### DIFF
--- a/pkg/validate/image.go
+++ b/pkg/validate/image.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"slices"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/validate/image.go
```diff
@@ -221,7 +220,7 @@ func pullImageList(c internalapi.ImageManagerService, imageList []string, podCon
 	for _, imageName := range imageList {
 		ids = append(ids, framework.PullPublicImage(c, imageName, podConfig))
 	}
-	return ids
+	return slices.Compact(ids)
```
SergeyKanzhelev, line 223: should we sort before compact? It looks like docs are saying it is deduping consequetive values, not all unique values

saschagrunert: That's tricky, the tests expect the ID's in the same order as the images, but I guess we can sort the image list and dedup the IDs then.

saschagrunert: Yeah, that depends on the test. I'll fix that up.

SergeyKanzhelev (approved):
/lgtm
/appeove

PR #1724, opened 2024-12-14
feat: retrieve specific log stream of a container

knight42:
#### What type of PR is this?

/kind feature

Per https://github.com/kubernetes/kubernetes/pull/127360#issuecomment-2465312757, I am going to make crictl be able to fetch a specific log stream of a container. 

BTW it turns out that crictl reads the log file directly (without any communication with apiserver or kubelet), so this change actually doesn't depend on k8s 1.32 APIs.

None

```release-note
`crictl logs` now is able to fetch a specific log stream of a container with the introducted `--stream` flag.
```


cmd/crictl/logs.go
```diff
@@ -75,6 +81,10 @@ var logsCommand = &cli.Command{
 			Aliases: []string{"t"},
 			Usage:   "Show timestamps",
 		},
+		&cli.StringFlag{
+			Name:  "stream",
```
saschagrunert, line 85: Do we want to add an alias?

```suggestion
			Name:  "stream",
			Aliases: []string{"s"},
```

knight42: I think it is great to add one.

feiskyer:
@knight42 could you fix the lint errors?

knight42:
> @knight42 could you fix the lint errors?

Sure! Please take another look.

knight42:
@feiskyer The test failure appears to be irrelevant. Could you please rerun the test? It seems I am unable to do so on my end.

feiskyer (approved):
Thanks addressing the lint issues

Rerun works now

/lgtm

PR #1749, opened 2025-01-16
When use crictl rmp -a -f, Avoid excessive coroutines causing context…

jokemanfire:
There were some issues when creating approximately 400 containers and 200 pods.When I use crictl rmp -a -f.
The error occur
`removing the pod sandbox "5c963da93d21cb6af42513ddbb682491aac7c91fd687f8e3bbf4d51a7dfaffed": rpc error: code = DeadlineExceeded desc = context deadline exceeded`

#### What type of PR is this?
/kind optimize

Increase the maximum number of co processes limit.

### DIFF
--- a/cmd/crictl/sandbox.go
+++ b/cmd/crictl/sandbox.go
@@ -166,8 +166,13 @@ var removePodCommand = &cli.Command{
 		}
 
 		funcs := []func() error{}
+		// set max concurrency to 10, to avoid too many goroutines
+		maxConcurrency := 10
+		sem := make(chan struct{}, maxConcurrency)
 		for _, id := range ids {
 			funcs = append(funcs, func() error {
+				sem <- struct{}{}
+				defer func() { <-sem }()
 				resp, err := InterruptableRPC(nil, func(ctx context.Context) (*pb.PodSandboxStatusResponse, error) {
 					return runtimeClient.PodSandboxStatus(ctx, id, false)
 				})


kwilczynski:
@jokemanfire have you tried to increase the default timeout, which is rather short, instead?

jokemanfire:
> @jokemanfire have you tried to increase the default timeout, which is rather short, instead?

It could , But is it more user-friendly to increase the concurrency limit on goroutines here？Otherwise, I would have to try a better threshold on timeout. Thanks

saschagrunert:
> But is it more user-friendly to increase the concurrency limit on goroutines here？

I would not say it's more user friendly, because it may be intentional to have more than `10` pods to be removed at a specific time.

jokemanfire:
> > But is it more user-friendly to increase the concurrency limit on goroutines here？
> 
> I would not say it's more user friendly, because it may be intentional to have more than `10` pods to be removed at a specific time.

In my opinion, the design should be more inclined towards thread pools or database connection pools (given a limit, after all, containerd processing requests also have their own concurrency). If 1000 pods are deleted and 1000 coroutines or threads are started at the same time, I think this is a problem in the design, after all, crictl is not a tool for testing concurrency limit. This is just my opinion.

saschagrunert:
Closing this one for now since it looks like we have no consensus on a solution.

PR #1768, opened 2025-02-10
Simplify auth handling

saschagrunert:
#### What type of PR is this?

/kind cleanup

The `auth` string can be empty and will be omitted during serialization in that case. We also extend the credential parsing error message to slightly more information.

None
None

```release-note
None
```

### DIFF
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -632,7 +632,7 @@ func getAuth(creds, auth, username string) (*pb.AuthConfig, error) {
 	if creds != "" {
 		username, password, err := parseCreds(creds)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("parse credentials: %w", err)
 		}
 
 		return &pb.AuthConfig{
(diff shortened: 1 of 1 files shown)


cmd/crictl/image.go
```diff
@@ -641,13 +641,7 @@ func getAuth(creds, auth, username string) (*pb.AuthConfig, error) {
...
 	}
 
-	if auth != "" {
-		return &pb.AuthConfig{
-			Auth: auth,
-		}, nil
-	}
-
-	return nil, nil
+	return &pb.AuthConfig{Auth: auth}, nil
```
SergeyKanzhelev, line 644: hm, this will result in a different path here: https://github.com/containerd/containerd/blob/59c8cf6ea5f4175ad512914dd5ce554942bf144f/internal/cri/server/images/image_pull.go#L103

Or serialization of empty object will result in `nil`?

saschagrunert: Yes this should result in an empty object.

SergeyKanzhelev: so will the comparison with `nil` be true for an empty object? I worried that the behavior of a pull will change as contianerd seems to take some host auth if nil was passed.

SergeyKanzhelev: @mikebrow do you know this part?

saschagrunert:
Closing this for now, does not actually seem to be required.

PR #1817, opened 2025-04-08
chore: add and migrate to sigs.k8s.io/yaml

wcrum:
#### What type of PR is this?
/kind cleanup

/kind deprecation

Move away from depreciated requirements and standard usage of sigs.k8s.io/yaml

This PR does not fix any open issue.

```release-note

```


vendor/modules.txt
```diff
@@ -600,3 +600,4 @@ sigs.k8s.io/structured-merge-diff/v4/value
 ## explicit; go 1.12
 sigs.k8s.io/yaml
 sigs.k8s.io/yaml/goyaml.v2
```
SergeyKanzhelev, line 602: do we need both? Or this is transitive?

wcrum: I believe this is just transitive, it was introduced when bumping from 1.3.0 to 1.4.0 which as well has goyaml.v3.

SergeyKanzhelev: so we need both in vendor folder? What uses it?

wcrum: Seems that way.

https://github.com/kubernetes-sigs/yaml/blob/b9a9b1cb9d17f91e923529e5dcd3df1a6500b6a3/yaml.go#L27

The primary module references and utilizes the `sigs.k8s.io/yaml/goyaml.v2` and `sigs.k8s.io/yaml/goyaml.v3`.

SergeyKanzhelev: /lgtm

saschagrunert:
CI fails because of https://github.com/kubernetes-sigs/cri-tools/issues/1821

saschagrunert:
Please rebase to fix the CI.

saschagrunert (approved):

PR #1834, opened 2025-04-29
add local imports to goimports formatting settings

tariq1890:
#### What type of PR is this?

/kind cleanup

This enforces the local imports ordering when checking for goimports formatting

None

No

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -162,6 +162,9 @@ formatters:
         - standard
         - default
         - localmodule
+    goimports:
+      local-prefixes:
+        - sigs.k8s.io/cri-tools
   exclusions:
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -162,6 +162,9 @@ formatters:
         - standard
         - default
         - localmodule
+    goimports:
+      local-prefixes:
+        - sigs.k8s.io/cri-tools
```
saschagrunert, line 167: Do you mind elaborating why this is necessary when we have no code changes?

saschagrunert: Thank you for the explanation!

PR #1845, opened 2025-05-16
Fix containerd CI tests

saschagrunert:
#### What type of PR is this?

/kind failing-test

Fix the containerd tests in GitHub actions.

None
None

```release-note
None
```


saschagrunert:

.github/workflows/containerd.yml
```diff
@@ -321,6 +321,11 @@ jobs:
           # Remove possibly existing containerd configuration
           sudo rm -rf /etc/containerd
           sudo PATH=$PATH bash -c "/usr/local/bin/containerd -a ${BDIR}/c.sock -root ${BDIR}/root -state ${BDIR}/state -log-level debug &> ${BDIR}/containerd-cri.log &"
+
+          while ! grep -q 'containerd successfully booted' ${BDIR}/containerd-cri.log; do
+              sleep 1
+          done
```
kwilczynski, line 327: The challenge with this approach is... if for some reason the specific log line will never appear, then you will have an infinite loop... eventually the entire build would time out, I suppose.

I wonder if there is a better, safer, way to handle this...

PR #1850, opened 2025-06-02
Add --all flag to crictl stop command.

kenneth-dsouza:
#### What type of PR is this?
With this flag we can stop all containers at once.

/kind feature

During troubleshooting one needs to stop all containers on the node, instead of using regex and other commands we can simply use --all flag with stop command.

We are already supporting --all flag for other sub commands like rmi and rm.

Yes

~~~
# crictl stop --all 
1813a6972abbd043c737fa645bef6eb63dac3cc18e98569ec81731007375fba6
94e56fd327d579eed8501c5d82059b9b7ada8b1f5b9a485c997c3f10b162e20d
be17961753d029bab5d6b965013a6ed378f7d5abeb9c832fc8262de7e2dd1968
5bb2d026c301126eef4982623137d86f8efe1a6a1f07c72fe9c0f70cc3c05b08
2dcd498b542499c5660c45284e6f52550a9854b904c2f9f4a3c78683fedb688f
b62ad9a6025c19c2816f3104be21eb4109aafb76110bc1fb2c2a1cdb96988807
578bbc25e454ea374c1c7f163b91c20b9aa74cca1ae529aeed6fe20ddf80de0b
1280f939f04f4f5376791fd096812002445936213b6462b4e84f43c21dc4061a
59e1ada61ace37490970ff9247c87e3c01462a1f64e30c9810079b79e48698a3
3a943a5670abf68e358da4ec6fb6e9c0ea4f2a9819f414c8edc7abd4a775074e
a7a7729852437458f8f8aedd8f00299669d9f48d316459b7213851af8a875500
~~~

```release-note

```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -367,18 +367,52 @@ var stopContainerCommand = &cli.Command{
 			Aliases: []string{"t"},
 			Usage:   "Seconds to wait to kill the container after a graceful stop is requested",
 		},
+		&cli.BoolFlag{
+			Name:    "all",
+			Aliases: []string{"a"},
+			Usage:   "Stop all running containers",
+		},
 	},
 	Action: func(c *cli.Context) error {
-		if c.NArg() == 0 {
-			return errors.New("ID cannot be empty")
-		}
 		runtimeClient, err := getRuntimeService(c, 0)
 		if err != nil {
 			return err
 		}
 
-		for i := range c.NArg() {
-			containerID := c.Args().Get(i)
+		if c.Bool("all") && c.NArg() > 0 {
+			return cli.Exit("cannot specify CONTAINER-ID(s) when using --all", 1)
+		}
+
+		var containerIDs []string
+
+		if c.Bool("all") {
+			opts := &listOptions{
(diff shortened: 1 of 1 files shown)


cmd/crictl/container.go
```diff
@@ -367,18 +367,52 @@ var stopContainerCommand = &cli.Command{
...
-		}
 		runtimeClient, err := getRuntimeService(c, 0)
 		if err != nil {
 			return err
 		}
 
-		for i := range c.NArg() {
-			containerID := c.Args().Get(i)
+		if c.Bool("all") && c.NArg() > 0 {
+			return cli.Exit("cannot specify CONTAINER-ID(s) when using --all", 1)
```
SergeyKanzhelev, line 383: perhaps e2e test can be written for some of the error conditions like this in https://github.com/kubernetes-sigs/cri-tools/tree/master/test/e2e

kenneth-dsouza: I felt that one would use -all to stop running containers than the one which are already stopped.

kenneth-dsouza:
@mrunalp @derekwaynecarr can you please review this PR?

PR #1445, opened 2024-06-05
Update e2e test suite

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Make the e2e tests to expect an installed runtime version + endpoint and move them to the CRI-O CI to test the latest version.

#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #1918, opened 2025-09-22
docs: updated the typo in CRI-O badge image URL

grinish21:
#### What type of PR is this?
/kind documentation

 The PR fixes the staus badge which had the wrong URL 
 Before: 
image

After:
image

NO

```release-note

```


SergeyKanzhelev:
This was the rename: https://github.com/kubernetes-sigs/cri-tools/pull/1445/

@saschagrunert should we rename it for containerd as well?

grinish21:
> > This was the rename: #1445
> > @saschagrunert should we rename it for containerd as well?
> 
> Yeah, let's keep both names in sync, but this can also be a follow up.
> 
> Thank you for the PR!

The suggestion has been committed , the names are in sync now. Thank you for such a wonderful CLI.

PR #2034, opened 2026-03-24
local dev docker-based containerd environment

SergeyKanzhelev:
#### What type of PR is this?

/kind cleanup

This allows to ask AI agent to run tests without the need to grant them `sudo`.

```release-note
NONE
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,12 @@ To build `crictl` run:
 make crictl
 ```
 
+To build the `crictl` e2e tests binary, run:
+
+```bash
+make crictl-e2e
+```
+
 ### Linter rules
 
 To install the recommended linter (`golangci-lint`), run:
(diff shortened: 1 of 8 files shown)


AGENTS.md
```diff
@@ -107,6 +113,42 @@ This is how tests will run on CI/CD. In order to run a single test while debuggi
 sudo PATH="$PATH:$(pwd)/build/bin/<GOOS>/<GOARCH>" ./build/bin/<GOOS>/<GOARCH>/critest --ginkgo.vv --ginkgo.focus="<regex_to_match_test_description>"
 ```
 
+#### Running Tests in a Container
+
+If you don't have a local container runtime installed, you can run the tests in a container using Docker. This will build a local image with `containerd` and run the tests inside it. This approach doesn't require `sudo` if your user has access to Docker.
+
+**Note:** AppArmor tests must be skipped as the containerized environment does not support them. The `make` targets below already include the skip flag.
```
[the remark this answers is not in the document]
SergeyKanzhelev: leftover, removed

images/containerd-local-test/.dockerignore
```diff
@@ -0,0 +1,5 @@
+*
```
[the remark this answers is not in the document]
SergeyKanzhelev: thanks for the pointer! Renamed

images/containerd-local-test/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# Copy setup scripts
+COPY hack/setup-containerd.sh /usr/local/bin/setup-containerd.sh
+COPY hack/wait-for-containerd.sh /usr/local/bin/wait-for-containerd.sh
+RUN chmod +x /usr/local/bin/setup-containerd.sh /usr/local/bin/wait-for-containerd.sh
+
+# Initial setup
+RUN setup-containerd.sh
+
+# Copy entrypoint
+COPY images/containerd-local-test/entrypoint.sh /entrypoint.sh
```
saschagrunert, line 41: `RUN setup-containerd.sh` during the image build is unnecessary since the entrypoint re-runs it at container start anyway (to pick up the `RUNTIME` env var). This just adds an extra image layer that gets overwritten at runtime. Consider removing this line.

hack/run-e2e-container.sh
```diff
@@ -0,0 +1,68 @@
...
+OPTIONAL_MOUNTS=()
+if [ -d "/lib/modules" ]; then
+    OPTIONAL_MOUNTS+=(-v "/lib/modules:/lib/modules:ro")
+fi
+if [ -d "/etc/apparmor.d" ]; then
+    OPTIONAL_MOUNTS+=(-v "/etc/apparmor.d:/etc/apparmor.d:ro")
+fi
+
+docker run --rm --privileged \
+    -e "RUNTIME=${RUNTIME:-io.containerd.runc.v2}" \
```
saschagrunert, line 62: Mounting the host's `/tmp` into the container means host temp files leak into the test environment and concurrent runs could conflict. Since containerd runs inside the container, the tests should work fine with the container's own `/tmp`. Is there a specific reason for this mount?

saschagrunert (commented):
Previous comments addressed (Dockerfile.dockerignore rename, AppArmor skip removal, set -euo pipefail in entrypoint). A couple of remaining items inline.

images/containerd-local-test/Dockerfile
SergeyKanzhelev: yes, fixed

hack/run-e2e-container.sh
SergeyKanzhelev: it was a leftover of experimenting. Fixed

saschagrunert (approved):

PR #2039, opened 2026-04-07
added image volume critest

SergeyKanzhelev:
#### What type of PR is this?

/kind feature

cc @iholder101

image volume 

```release-note
New critest conformance tests for image volumes.
```

### DIFF
--- a/pkg/validate/image_volume.go
+++ b/pkg/validate/image_volume.go
@@ -1,0 +1,469 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validate
+
+import (
+	"runtime"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	internalapi "k8s.io/cri-api/pkg/apis"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"sigs.k8s.io/cri-tools/pkg/framework"
+)
+
+const (
+	imageVolumeTestPath = "/mnt/image-volume"
+)
+
+// skipIfImageVolumeUnsupported checks the error from CreateContainer and skips
+// the test if the runtime does not support image volumes yet.
+// Runtimes may signal lack of support in different ways:
+//   - gRPC Unimplemented: the runtime explicitly does not implement image volumes.
+//   - gRPC Unknown with "failed to mkdir" or similar: the runtime ignores the
+//     Image field in Mount and fails while processing an empty host path
+//     (observed with containerd 1.7).
+//
+// When expectError is true the caller expects CreateContainer to fail (negative
+// tests). In that case the function only checks for "unsupported" patterns and
+// returns without asserting no error.
+func skipIfImageVolumeUnsupported(err error, expectError bool) {
+	if err == nil {
+		return
+	}
+
+	if s, ok := status.FromError(err); ok {
+		if s.Code() == codes.Unimplemented {
+			Skip("Image Volumes are not yet supported by the runtime (Unimplemented).")
+		}
+
+		// containerd 1.7 does not recognise the Image field in Mount and
+		// produces errors like: failed to mkdir "": mkdir : no such file
+		// or directory. Treat these as "unsupported".
+		if s.Code() == codes.Unknown {
+			msg := s.Message()
+			if strings.Contains(msg, "failed to mkdir") ||
+				strings.Contains(msg, "failed to generate spec") {
+				Skip("Image Volumes are not yet supported by the runtime: " + msg)
+			}
+		}
+	}
+
+	if !expectError {
+		framework.ExpectNoError(err, "failed to create container")
+	}
+}
+
+var _ = framework.KubeDescribe("Image Volume [Feature:ImageVolume]", func() {
+	f := framework.NewDefaultCRIFramework()
+
+	var (
+		rc internalapi.RuntimeService
+		ic internalapi.ImageManagerService
+	)
+
+	skipOnWindows := func() {
+		if runtime.GOOS == framework.OSWindows && !framework.TestContext.IsLcow {
+			Skip("Image Volume tests are not supported on Windows")
+		}
+	}
+
+	BeforeEach(func() {
+		skipOnWindows()
+
+		rc = f.CRIClient.CRIRuntimeClient
+		ic = f.CRIClient.CRIImageClient
+	})
+
+	It("should return image digest in ContainerStatus mounts", func(ctx SpecContext) {
+		imageName := testImageWithTag
+
+		By("Pulling image: " + imageName)
+		imageRef := framework.PullPublicImage(ctx, ic, imageName, testImagePodSandbox)
+
+		By("Creating pod sandbox")
+
+		podID, podConfig := framework.CreatePodSandboxForContainer(ctx, rc)
+
+		defer func() {
+			By("stop PodSandbox")
+			Expect(rc.StopPodSandbox(ctx, podID)).NotTo(HaveOccurred())
+			By("delete PodSandbox")
+			Expect(rc.RemovePodSandbox(ctx, podID)).NotTo(HaveOccurred())
+		}()
+
+		By("Creating container with Image Volume")
+
+		containerName := "image-volume-digest-test-" + framework.NewUUID()
+		containerPath := imageVolumeTestPath
+
+		containerConfig := &runtimeapi.ContainerConfig{
+			Metadata: framework.BuildContainerMetadata(containerName, framework.DefaultAttempt),
+			Image: &runtimeapi.ImageSpec{
+				Image:              imageRef,
+				UserSpecifiedImage: imageName,
+			},
+			Command: pauseCmd,
+			Mounts: []*runtimeapi.Mount{
+				{
+					ContainerPath: containerPath,
+					Readonly:      true,
+					Image: &runtimeapi.ImageSpec{
+						Image:              imageRef,
+						UserSpecifiedImage: imageName,
+					},
+				},
+			},
+		}
+
+		containerID, err := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
+		skipIfImageVolumeUnsupported(err, false)
+
+		defer func() {
+			By("delete Container")
+			Expect(rc.RemoveContainer(ctx, containerID)).NotTo(HaveOccurred())
+		}()
+
+		By("Checking ContainerStatus for image volume digest")
+
+		statusResp, err := rc.ContainerStatus(ctx, containerID, false)
+		framework.ExpectNoError(err, "failed to get container status")
+
+		var found bool
+
+		for _, m := range statusResp.GetStatus().GetMounts() {
+			if m.GetContainerPath() == containerPath {
+				found = true
+				// Check if the new ImageSpec.image_ref (tag 20) field is populated (KEP-5365).
+				if m.GetImage() == nil || m.GetImage().GetImageRef() == "" {
+					Skip("Image Volume digest reporting (ImageSpec.image_ref tag 20) is not yet supported by the runtime.")
+				}
+
+				framework.Logf("Found Image Volume mount: Image.Image=%q, Image.ImageRef=%q",
+					m.GetImage().GetImage(), m.GetImage().GetImageRef())
+
+				Expect(m.GetImage().GetImageRef()).To(Equal(imageRef),
+					"ImageSpec.image_ref (tag 20) in Mount should match the image digest")
+
+				break
+			}
+		}
+
+		if !found {
+			Skip("Image Volume mount not found in ContainerStatus; runtime might not support Image Volumes.")
+		}
+	})
+
+	It("should correctly mount image and enforce read-only", func(ctx SpecContext) {
+		imageName := testImageWithTag
+
+		By("Pulling image: " + imageName)
+		imageRef := framework.PullPublicImage(ctx, ic, imageName, testImagePodSandbox)
+
+		By("Creating pod sandbox")
+
+		podID, podConfig := framework.CreatePodSandboxForContainer(ctx, rc)
+
+		defer func() {
+			By("stop PodSandbox")
+			Expect(rc.StopPodSandbox(ctx, podID)).NotTo(HaveOccurred())
+			By("delete PodSandbox")
+			Expect(rc.RemovePodSandbox(ctx, podID)).NotTo(HaveOccurred())
+		}()
+
+		By("Creating container with Image Volume")
+
+		containerName := "image-volume-mount-test-" + framework.NewUUID()
+		containerPath := imageVolumeTestPath
+
+		containerConfig := &runtimeapi.ContainerConfig{
+			Metadata: framework.BuildContainerMetadata(containerName, framework.DefaultAttempt),
+			Image: &runtimeapi.ImageSpec{
+				Image:              imageRef,
+				UserSpecifiedImage: imageName,
+			},
+			Command: pauseCmd,
+			Mounts: []*runtimeapi.Mount{
+				{
+					ContainerPath: containerPath,
+					Readonly:      true,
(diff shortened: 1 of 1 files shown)


saschagrunert (commented):
Thanks for adding image volume conformance tests, this is a great addition! Left some inline comments below.

go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/cri-tools
 go 1.26.3
 
 require (
+	github.com/blang/semver/v4 v4.0.0
```
saschagrunert: This vendors ~975 lines of code just to compare two-segment versions like `"2.1"`. A simple `major.minor` comparison with `strings.Cut` + `strconv.Atoi` would avoid the new dependency entirely.

.github/workflows/containerd.yml
```diff
@@ -81,6 +81,7 @@ jobs:
           path: ${{ github.workspace }}/src/github.com/containerd/containerd
           repository: containerd/containerd
           ref: ${{ format(matrix.version, '/') }}
+          fetch-depth: 0
```
saschagrunert: This fetches the full containerd git history for every CI run. Presumably needed so containerd's build derives its version via `git describe` for the new `requireContainerdVersion` check. Worth a comment explaining the rationale, or splitting into its own commit.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+		if resp.GetRuntimeName() != "containerd" {
+			return
+		}
+
+		// Throw away any version suffix which just causes semver parsing issues.
+		version, _, _ := strings.Cut(resp.GetRuntimeVersion(), "-")
+
+		parsedVersion, err := semver.ParseTolerant(version)
+		if err != nil {
+			framework.Logf("Skipping version check: failed to parse CRI runtime version %q: %v", version, err)
```
saschagrunert: Seven `Skip()` calls across tests tagged `[Conformance]`. Conformance tests define what a compliant runtime *must* implement. If these features are still optional, the tests should not carry `[Conformance]` yet, use the `[Feature:ImageVolume]` tag (already on the Describe) and promote to conformance later.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+
+		if parsedVersion.LT(requiredVersion) {
+			Skip(fmt.Sprintf("runtime containerd is version %q, but test requires at least %q", resp.GetRuntimeVersion(), givenVersion))
+		}
+	}
+
+	// skipOnWindows skips the test if running on Windows.
+	skipOnWindows := func() {
+		if runtime.GOOS == "windows" {
+			Skip("Image Volume tests are not supported on Windows")
```
saschagrunert, line 86: If `semver.ParseTolerant` fails, this logs and returns without skipping. The test then runs against an unparseable version and may produce confusing failures. Safer to `Skip()` when the version cannot be parsed.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+		}
+	}
+
+	// skipOnCrio skips the test if the runtime is CRI-O.
+	skipOnCrio := func(ctx context.Context, message string) {
```
saschagrunert: Hard-coding a runtime name in conformance tests couples the suite to specific implementations. If CRI-O doesn't support this behavior, either the test isn't conformance-ready or CRI-O should fail conformance.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	BeforeEach(func(ctx SpecContext) {
+		skipOnWindows()
+
+		rc = f.CRIClient.CRIRuntimeClient
+		ic = f.CRIClient.CRIImageClient
+
+		requireContainerdVersion(ctx, "2.1")
+	})
```
saschagrunert, line 95: This image pull result is discarded. The image is pulled again on line 119. Same pattern in the subPath tests. The first pull can be removed.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+		}
+
+		containerID, err := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
+		if err != nil {
+			// Check if the runtime explicitly rejects the image volume field (KEP-4639).
+			errStr := err.Error()
+			if strings.Contains(errStr, "Unimplemented") ||
+				strings.Contains(errStr, "unknown field") ||
+				strings.Contains(errStr, "not supported") ||
+				strings.Contains(errStr, "is not supported") {
```
saschagrunert: Matching against human-readable error strings is brittle. For gRPC errors, checking `status.Code(err) == codes.Unimplemented` is more reliable. The other strings (`"unknown field"`, `"not supported"`) have no stable contract.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+			Image: &runtimeapi.ImageSpec{
+				Image:              imageRef,
+				UserSpecifiedImage: imageName,
+			},
+			Command: pauseCmd,
+			Mounts: []*runtimeapi.Mount{
+				{
+					ContainerPath: containerPath,
+					Readonly:      true,
+					Image: &runtimeapi.ImageSpec{
```
saschagrunert, line 209: nit: the `ContainerConfig` here omits the `Command` field, while all other tests use `pauseCmd`. Inconsistent, and could mask issues if a runtime unexpectedly accepts this request.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+		// The test image is busybox-based, so /bin/ls should be executable.
+		cmd = []string{containerPath + "/bin/ls", containerPath}
+		_, stderr, err = rc.ExecSync(ctx, containerID, cmd, 0)
+		framework.ExpectNoError(err, "failed to execute binary from image volume (might be noexec): stderr: %s", stderr)
+		framework.Logf("Binary execution from image volume succeeded as expected")
+	})
+
+	It("should reject Image Volume with Readonly set to false [Conformance]", func(ctx SpecContext) {
+		skipOnCrio(ctx, "does not yet reject Image Volume with Readonly=false. "+
+			"This is a temporary skip and will be required for conformance in future versions.")
```
saschagrunert: The subPath tests don't have the unsupported-runtime skip logic (error string checks) that the earlier tests have. If a runtime doesn't support image volumes at all, these will hard-fail instead of skipping.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,464 @@
...
+			By("stop PodSandbox")
+			Expect(rc.StopPodSandbox(ctx, podID)).NotTo(HaveOccurred())
+			By("delete PodSandbox")
+			Expect(rc.RemovePodSandbox(ctx, podID)).NotTo(HaveOccurred())
+		}()
+
+		By("Creating container with Image Volume and Readonly=false")
+
+		containerName := "image-volume-readonly-test-" + framework.NewUUID()
+		containerPath := imageVolumeTestPath
```
saschagrunert, line 289: nit: "should fail if subPath in volume is not existing" reads awkwardly. Suggest: "should fail when subPath does not exist in the image".

.github/workflows/containerd.yml
SergeyKanzhelev: yes, it was the reason. More I think about more I realize we should not version-check and instead just declare incompatibility via `Skip`.  Let me play with it

SergeyKanzhelev:
I think I fixed all comments

saschagrunert (commented):
Thanks for addressing the previous feedback. A few more items, mainly two container leaks in the error-path tests.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,419 @@
...
+					Image: &runtimeapi.ImageSpec{
+						Image:              imageName,
+						UserSpecifiedImage: imageName,
+					},
+				},
+			},
+		}
+
+		_, err := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
+		if err != nil {
```
saschagrunert: When `CreateContainer` succeeds (`err == nil`), the container is never cleaned up because the ID is discarded via `_`. Capture `containerID`, defer `RemoveContainer`, then `Skip`.

For example:
```go
containerID, err := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
```

Then before the `Skip` on line 284:
```go
if err == nil {
	defer func() {
		Expect(rc.RemoveContainer(ctx, containerID)).NotTo(HaveOccurred())
	}()
	Skip("Runtime accepted Image Volume with Readonly=false; rejection is not yet enforced.")
}
```

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,419 @@
...
+							Image:              imageName,
+							UserSpecifiedImage: imageName,
+						},
+						ImageSubPath: subPath,
+					},
+				},
+			}
+
+			_, err := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
+			if err != nil {
```
saschagrunert: Same container leak here. If `CreateContainer` unexpectedly succeeds, the container is orphaned. Capture `containerID` and defer cleanup before the assertion:

```go
containerID, err := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
if err == nil {
	defer func() {
		Expect(rc.RemoveContainer(ctx, containerID)).NotTo(HaveOccurred())
	}()
}
```

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,419 @@
...
+
+		err = rc.StartContainer(ctx, containerID)
+		framework.ExpectNoError(err, "failed to start container")
+
+		By("Verifying image content is visible at the mount point")
+		// We use /bin/sh which is a standard binary present in the test image.
+		cmd := []string{"ls", "-A", containerPath + "/bin/sh"}
+
+		stdout, stderr, err := rc.ExecSync(ctx, containerID, cmd, 0)
+		if err != nil {
```
saschagrunert, line 235: All other `ExecSync` callers in the codebase pass `time.Duration(defaultExecSyncTimeout)*time.Second`. Passing `0` is inconsistent and its semantics may vary across runtimes. Same applies to the other ExecSync calls in this file (lines 213, 222, 230, 353, 362).

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	framework.ExpectNoError(err, "failed to create container")
+}
+
+var _ = framework.KubeDescribe("Image Volume [Feature:ImageVolume]", func() {
+	f := framework.NewDefaultCRIFramework()
+
+	var rc internalapi.RuntimeService
+
```
saschagrunert: nit: group into a single `var()` block to match the pattern in `container.go`, `pod.go`, etc.

```go
var (
	rc internalapi.RuntimeService
	ic internalapi.ImageManagerService
)
```

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,419 @@
...
+	// skipOnWindows skips the test if running on Windows.
+	skipOnWindows := func() {
+		if runtime.GOOS == "windows" {
+			Skip("Image Volume tests are not supported on Windows")
+		}
+	}
+
+	BeforeEach(func(ctx SpecContext) {
+		skipOnWindows()
```
saschagrunert, line 91: nit: this `BeforeEach` only calls `skipOnWindows()` and assigns variables, none of which need a context. The existing convention (e.g. `container.go:75`) uses bare `func()` for this.

go.mod
SergeyKanzhelev: this is gone now

.github/workflows/containerd.yml
SergeyKanzhelev: removed

pkg/validate/image_volume.go
SergeyKanzhelev: yes, now all skips are based on outcome, not the version

pkg/validate/image_volume.go
SergeyKanzhelev: fixed

pkg/validate/image_volume.go
SergeyKanzhelev: removed context

pkg/validate/image_volume.go
SergeyKanzhelev: merged

pkg/validate/image_volume.go
SergeyKanzhelev: fixed

pkg/validate/image_volume.go
SergeyKanzhelev: yep, addressed

pkg/validate/image_volume.go
SergeyKanzhelev: done

SergeyKanzhelev:
This seems to be ready. Containerd 1.7 behavior is strange, but we only need it for another half a year or so

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,472 @@
...
+
+var _ = framework.KubeDescribe("Image Volume [Feature:ImageVolume]", func() {
+	f := framework.NewDefaultCRIFramework()
+
+	var (
+		rc internalapi.RuntimeService
+		ic internalapi.ImageManagerService
+	)
+
+	// skipOnWindows skips the test if running on Windows.
```
saschagrunert: This uses a raw `"windows"` string while every other check in `pkg/validate/` uses `framework.OSWindows`. It also unconditionally skips on Windows, but LCOW (Linux Containers on Windows) can run Linux image volumes. The established pattern is:

```go
if runtime.GOOS == framework.OSWindows && !framework.TestContext.IsLcow {
    Skip("Image Volume tests are not supported on Windows")
}
```

saschagrunert (commented):
Thanks for the updates, most of the previous feedback is addressed. Four remaining items below.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,472 @@
...
+			Metadata: framework.BuildContainerMetadata(containerName, framework.DefaultAttempt),
+			Image: &runtimeapi.ImageSpec{
+				Image:              imageRef,
+				UserSpecifiedImage: imageName,
+			},
+			Command: pauseCmd,
+			Mounts: []*runtimeapi.Mount{
+				{
+					ContainerPath: containerPath,
+					Readonly:      true,
```
saschagrunert, line 129: Per the proto comment on `Mount.image`:

> The kubelet does a PullImage RPC and evaluates the returned PullImageResponse.image_ref value, which is then set to the ImageSpec.image field.

So the mount's `Image.Image` should be `imageRef` (the digest returned by PullImage), not `imageName` (the tag reference), to match what kubelet actually sends. Same applies to all other mount ImageSpecs in this file (lines 174, 280, 339, 430).

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,472 @@
...
+		imageName := testImageWithTag
+
+		By("Pulling image: " + imageName)
+		imageRef := framework.PullPublicImage(ctx, ic, imageName, testImagePodSandbox)
+
+		By("Creating pod sandbox")
+
+		podID, podConfig := framework.CreatePodSandboxForContainer(ctx, rc)
+
+		defer func() {
```
saschagrunert, line 279: If a runtime doesn't support image volumes but returns a generic error (e.g. containerd 1.7's `"failed to mkdir"`), the test falls through to line 304 and reports "Verified that Runtime rejected invalid Readonly=false configuration". That's misleading because the runtime didn't validate `Readonly=false`, it just doesn't support image volumes at all.

This should filter through `skipIfImageVolumeUnsupported` (or equivalent checks) before treating the error as a correct rejection of the Readonly constraint.

pkg/validate/image_volume.go
```diff
@@ -0,0 +1,472 @@
...
+			containerConfig := &runtimeapi.ContainerConfig{
+				Metadata: framework.BuildContainerMetadata(containerName, framework.DefaultAttempt),
+				Image: &runtimeapi.ImageSpec{
+					Image:              imageRef,
+					UserSpecifiedImage: imageName,
+				},
+				Command: pauseCmd,
+				Mounts: []*runtimeapi.Mount{
+					{
+						ContainerPath: containerPath,
```
saschagrunert, line 443: Same issue as the Readonly=false test: a runtime that doesn't support image volumes at all but returns a generic error would be mistakenly credited with correctly rejecting a non-existing subPath. Should also filter through `skipIfImageVolumeUnsupported` before treating the error as a valid subPath rejection.

pkg/validate/image_volume.go
SergeyKanzhelev: fixed

pkg/validate/image_volume.go
SergeyKanzhelev: fixed

pkg/validate/image_volume.go
SergeyKanzhelev: right, changed

pkg/validate/image_volume.go
SergeyKanzhelev: yes, fixed

saschagrunert (approved):

PR #2042, opened 2026-04-09
refactor: move linux-specific validation code to separate files

SergeyKanzhelev:
#### What type of PR is this?

/kind cleanup

Random clean up to remove warnings on build for "unused" things.

Also pre-compiled the regex in `verifyLogContentsRe ` so all logs are matched against precompiled version of it. If patter is wrong, the function will fail now.

```release-note
NONE
```

### DIFF
--- a/pkg/validate/consts.go
+++ b/pkg/validate/consts.go
@@ -198,27 +198,21 @@ const (
 	defaultDNSOption       string = "ndots:8"
 	webServerContainerPort int32  = 80
 	// The following host ports must not be in-use when running the test.
-	webServerHostPortForPortMapping        int32 = 12000
-	webServerHostPortForPortForward        int32 = 12001
-	webServerHostPortForHostNetPortFroward int32 = 12002
-	// The port used in hostNetNginxImage (See images/hostnet-nginx/).
-	webServerHostNetContainerPort int32 = 12003
+	webServerHostPortForPortMapping int32 = 12000
+	webServerHostPortForPortForward int32 = 12001
 
 	// Linux defaults.
-	webServerLinuxImage        = framework.DefaultRegistryE2ETestImagesPrefix + "nginx:1.14-2"
-	hostNetWebServerLinuxImage = registry + "hostnet-nginx-" + runtime.GOARCH
+	webServerLinuxImage = framework.DefaultRegistryE2ETestImagesPrefix + "nginx:1.14-2"
 
 	// Windows defaults.
-	webServerWindowsImage        = webServerLinuxImage
-	hostNetWebServerWindowsImage = webServerLinuxImage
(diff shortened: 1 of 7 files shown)


SergeyKanzhelev:
this is from macos linter issues. Not critical, but nice to have

pkg/validate/consts_linux.go
```diff
@@ -0,0 +1,31 @@
...
+
+const (
+	// The port used in hostNetNginxImage (See images/hostnet-nginx/).
+	webServerHostNetContainerPort int32 = 12003
+
+	// The host port for hostnet port-forward tests; only used on Linux.
+	webServerHostPortForHostNetPortForward int32 = 12002
+)
+
+var hostNetWebServerImage = registry + "hostnet-nginx-" + runtime.GOARCH
```
saschagrunert:
```suggestion
const hostNetWebServerImage = registry + "hostnet-nginx-" + runtime.GOARCH
```

saschagrunert (commented):
Just one nit, otherwise LGTM. The function `verifyLogContentsRe` changes in logic a bit, probably worth to mention in the PR description.

pkg/validate/consts_linux.go
SergeyKanzhelev: fixed

SergeyKanzhelev:
> Just one nit, otherwise LGTM. The function `verifyLogContentsRe` changes in logic a bit, probably worth to mention in the PR description.

Updated, it is just a small optimization as I was on it

pkg/validate/consts.go
```diff
@@ -198,27 +198,21 @@ const (
...
 
 	// Linux defaults.
-	webServerLinuxImage        = framework.DefaultRegistryE2ETestImagesPrefix + "nginx:1.14-2"
-	hostNetWebServerLinuxImage = registry + "hostnet-nginx-" + runtime.GOARCH
+	webServerLinuxImage = framework.DefaultRegistryE2ETestImagesPrefix + "nginx:1.14-2"
 
 	// Windows defaults.
-	webServerWindowsImage        = webServerLinuxImage
-	hostNetWebServerWindowsImage = webServerLinuxImage
+	webServerWindowsImage = webServerLinuxImage
```
haircommander, line 208: I think we should compress these if we're otherwise separating files

SergeyKanzhelev: do you suggest separate Windows as well? I can do it as a follow up. I just have this annoying warning now when trying it on mac that I want to resolve

SergeyKanzhelev: actually I wouldn't be able to split up Windows away because of a support of Linux on Windows: `framework.TestContext.IsLcow`

saschagrunert (approved):

PR #2048, opened 2026-04-13
Makefile: Don't explicitly pass GOFLAGS to go commands

chewi:
#### What type of PR is this?

/kind bug

go automatically checks GOFLAGS and filters unknown flags for you, e.g. it will drop the `go run` -exec flag when doing `go build`. Explicitly passing GOFLAGS breaks this filtering.

None

FYI @dvzrv

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ VERSION ?= $(shell git describe --tags --dirty --always | sed 's/^v//')
 CGO_ENABLED ?= 0
 GOFLAGS ?= -trimpath
 GO_LDFLAGS := $(GO_LDFLAGS) -X $(PROJECT)/pkg/version.Version=$(VERSION)
+export GOFLAGS
 
 BUILD_PATH := $(shell pwd)/build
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
> go automatically checks GOFLAGS and filters unknown flags for you, e.g. it will drop the `go run` -exec flag when doing `go build`. Explicitly passing GOFLAGS breaks this filtering.

can you explain which flags are breaking which command and why they are set on your environment? Overwriting the variable feels like unintended consequence of running the build that we need to avoid

chewi:
I am a maintainer for Gentoo Linux, and we were adding `-exec` to `GOFLAGS` when cross-compiling because that is what the option is for. As I mentioned, this option is for `go run` and it breaks when explicitly passed to `go build`. We have since taken a different approach for an unrelated reason, but I still consider this a perfectly legitimate thing to do.

SergeyKanzhelev (approved):
/lgtm

sounds like well-motivated change. I thought the export might leak, but it looks like it only works for child processes

PR #2051, opened 2026-04-14
Fix crictl images not filtering by repository:tag

vigneshakaviki:
## Summary

- Add client-side image name filtering in `ListImages` as a fallback when the CRI runtime does not honor the `ListImagesRequest` filter
- Some runtimes return all images regardless of the `ImageFilter.Image` spec, causing `crictl images repo:tag` to show unrelated images
- The fix filters results after the CRI call, matching by `RepoTags` and `RepoDigests`

Fixes #2008

## Root Cause

`crictl images REPOSITORY:TAG` passes the filter to the CRI runtime via `ListImagesRequest.Filter.Image`, but some runtimes (e.g. certain containerd configurations) ignore this filter and return all images. There was no client-side fallback filtering.

## Changes

- Added `filterByName()` in `cmd/crictl/image.go` — filters image list by matching `RepoTags` or `RepoDigests` against the user-provided name filter
- Added `matchesNameFilter()` helper — handles exact `repo:tag` match, repo-only match (when no tag specified), and `repo@digest` match
- Applied after CRI response in `ListImages()`, before sorting and condition filters

## Test plan

- [x] Added 5 ginkgo test cases for `filterByName`: exact repo:tag, repo-only, no match, digest match, tag-specific match
- [x] All 31 existing tests pass

```
Ran 31 of 31 Specs in 0.002 seconds
SUCCESS! -- 31 Passed | 0 Failed | 0 Pending | 0 Skipped
```

- [x] Tested on minikube (containerd runtime):

```
=== Patched: filter by exact repo:tag ===
IMAGE                   TAG                 IMAGE ID            SIZE
registry.k8s.io/pause   3.10.1              cd073f4c5f6a8       736kB

=== Patched: filter by repo only ===
IMAGE                   TAG                 IMAGE ID            SIZE
registry.k8s.io/pause   3.10.1              cd073f4c5f6a8       736kB

=== Patched: filter by non-existent tag ===
IMAGE               TAG                 IMAGE ID            SIZE

=== Patched: no filter (all images) ===
IMAGE                                      TAG        IMAGE ID       SIZE
gcr.io/k8s-minikube/storage-provisioner    v5         6e38f40d628db  31.5MB
...
```

### DIFF
--- a/cmd/crictl/image.go
+++ b/cmd/crictl/image.go
@@ -806,6 +806,13 @@ func ListImages(ctx context.Context, client internalapi.ImageManagerService, nam
 	resp := &pb.ListImagesResponse{Images: res}
 	logrus.Debugf("ListImagesResponse: %v", resp)
 
+	// The CRI runtime may not honor the image name filter, so apply it
+	// client-side to ensure correct results. See
+	// https://github.com/kubernetes-sigs/cri-tools/issues/2008
+	if nameFilter != "" && len(resp.GetImages()) > 0 {
+		resp.Images = filterByName(nameFilter, resp.GetImages())
+	}
+
 	slices.SortFunc(resp.GetImages(), func(a, b *pb.Image) int {
 		if len(a.GetRepoTags()) > 0 && len(b.GetRepoTags()) > 0 {
 			return cmp.Compare(a.GetRepoTags()[0], b.GetRepoTags()[0])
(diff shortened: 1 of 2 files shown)


cmd/crictl/image.go
```diff
@@ -895,6 +902,58 @@ func filterByBeforeSince(filterValue string, imageList []*pb.Image) []*pb.Image
...
+// matchesNameFilter checks if an image reference matches the name filter.
+// If the filter contains a tag (":") or digest ("@"), it must match exactly.
+// Otherwise, only the repository part is compared.
+func matchesNameFilter(imageRef, nameFilter string) bool {
+	if imageRef == nameFilter {
+		return true
+	}
+
+	// If the filter has no tag/digest, match by repository name only.
+	if !strings.Contains(nameFilter, ":") && !strings.Contains(nameFilter, "@") {
```
saschagrunert, line 943: This condition treats any `:` in the filter as a tag separator, but registry URLs commonly contain ports (e.g. `localhost:5000/myimage`, `registry.example.com:8443/app`). Filtering by `localhost:5000/myimage` would contain `:`, skip the repo-only matching block, and return `false` for every image.

`github.com/distribution/reference` is already a dependency. Using `reference.Parse` to split the filter into its components would handle ports, tags, and digests correctly.

cmd/crictl/image.go
```diff
@@ -895,6 +902,58 @@ func filterByBeforeSince(filterValue string, imageList []*pb.Image) []*pb.Image
...
+		repo := imageRef
+		if idx := strings.LastIndex(repo, ":"); idx != -1 {
```
saschagrunert, line 945: The `@` check needs to come before the `:` check. For a RepoDigest like `docker.io/library/busybox@sha256:abc123`, `LastIndex(repo, ":")` finds the `:` inside the digest hash, producing `docker.io/library/busybox@sha256` instead of `docker.io/library/busybox`. This breaks repo-only matching against RepoDigests (relevant for images pulled by digest that have no RepoTags).

```suggestion
		if idx := strings.LastIndex(repo, "@"); idx != -1 {
			repo = repo[:idx]
		} else if idx := strings.LastIndex(repo, ":"); idx != -1 {
```

Though this becomes moot if you switch to `reference.Parse` per the comment above.

cmd/crictl/image.go
```diff
@@ -895,6 +902,58 @@ func filterByBeforeSince(filterValue string, imageList []*pb.Image) []*pb.Image
...
+// not honor the ListImages filter.
+func filterByName(nameFilter string, imageList []*pb.Image) []*pb.Image {
+	var filtered []*pb.Image
+
+	for _, img := range imageList {
+		for _, repoTag := range img.GetRepoTags() {
+			if matchesNameFilter(repoTag, nameFilter) {
+				filtered = append(filtered, img)
+
+				goto next
```
saschagrunert, line 917: Nit: a labeled `continue` is more idiomatic Go for this pattern:

```go
outer:
	for _, img := range imageList {
		for _, repoTag := range img.GetRepoTags() {
			if matchesNameFilter(repoTag, nameFilter) {
				filtered = append(filtered, img)
				continue outer
			}
		}
		// ...
	}
```

saschagrunert (changes requested):
Two correctness issues in `matchesNameFilter` and a style nit.

SergeyKanzhelev:
@vigneshakaviki do you still plan to work on this?

PR #2055, opened 2026-04-23
crictl: replace docker/docker timestamp dependency

timonwong:
#### What type of PR is this?

/kind cleanup

Replace `github.com/docker/docker` dependency with a local implementation.

- `github.com/docker/docker` introduces too much CVE noise in dependency and security scanning results.
- The old `docker/docker` module path is effectively superseded by `moby/moby`, but the timestamp is now internal package so we cannot use it.

Reference implementations:
- nerdctl timestamp implementation: https://github.com/containerd/nerdctl/blob/9b4248721a8e2c2b173bb5fdcad79df3bd4819dd/pkg/timestamp/timestamp.go
- container-libs timetype implementation: https://github.com/containers/container-libs/blob/4089c55104bf8fd73fbf921fcdbf5e8ad49f6140/common/pkg/timetype/timestamp.go

```release-note
NONE
```


pkg/timetype/timestamp.go
```diff
@@ -0,0 +1,134 @@
+/*
```
SergeyKanzhelev, line 17: Kubernetes copyright needs to be on top. Both copyrights can be preserved if this is a direct copy

pkg/timetype/timestamp.go
```diff
@@ -0,0 +1,134 @@
...
+Originally from https://github.com/moby/moby/blob/v2.0.0-beta.9/client/internal/timestamp/timestamp.go
+Copyright (C) Docker/Moby authors.
+Licensed under the Apache License, Version 2.0
+NOTICE: https://github.com/moby/moby/blob/v2.0.0-beta.9/NOTICE
+*/
+
+package timetype
```
SergeyKanzhelev, line 24: you beat me to it. Did you check if K8s has anything similar that we can use?

timonwong: After a quick search of the ripgrep  within kubectl repo, it implements similar logic:

In ：[https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/logs/logs.go](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/logs/logs.go)

- Timestamp Parsing: The [ParseRFC3339](https://github.com/kubernetes/kubectl/blob/ef67d87dbf47049dbbed5fe8d9309a72fe157c43/pkg/util/util.go#L31-L40) function is used for the `--since-time` flag.
- Duration Handling: Native time.ParseDuration is used for the `--since` flag.

However, as observed, their behavior differs significantly from that of crictl.

SergeyKanzhelev (approved):
/lgtm
/approve

we are replacing code with the newer copy that is slightly different. The difference is cosmetic, not functional.

PR #2077, opened 2026-05-13
ci: declare contents: read across the 4 remaining workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to read-only on the four workflows still inheriting org defaults:

- `build.yml` — image build + linter matrix.
- `containerd.yml` — runs the CRI test suite against containerd shims.
- `crio.yml` — runs e2e + critest against CRI-O.
- `release.yml` — `make release-notes release` and `ncipollo/release-action` authenticate via `secrets.GH_TOKEN` (a dedicated PAT), so the default `GITHUB_TOKEN` only needs read.

YAML validated locally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,9 @@ on:
     branches:
       - master
   pull_request:
+permissions:
+  contents: read
+
 jobs:
   images:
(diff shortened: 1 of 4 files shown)


saschagrunert (approved):
Thanks!

PR #2083, opened 2026-05-19
crictl: extract shared helpers and fix ContainerStats context

saschagrunert:
#### What type of PR is this?
/kind cleanup
/kind bug

- Fixes `ContainerStats` to propagate the CLI context instead of using `context.TODO()`, so cancellation and timeouts work correctly.
- Extracts `truncateCount` and `collectIDs[T]` helpers to deduplicate identical list-truncation and ID-collection logic in container and sandbox commands.
- Adds unit tests for both new helpers.

None

Image removal is intentionally left alone since it uses a different pattern (`map[string]bool` for prune filtering).

```release-note
NONE
```

### DIFF
--- a/cmd/crictl/util.go
+++ b/cmd/crictl/util.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/invopop/jsonschema"
 	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/protoadapt"
 	"google.golang.org/protobuf/runtime/protoiface"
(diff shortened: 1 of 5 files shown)


saschagrunert:

cmd/crictl/util.go
```diff
@@ -666,3 +667,48 @@ func AggregateGoroutines(funcs ...func() error) error {
...
+	}
+
+	if last > 0 {
+		n = last
+	}
+
+	return min(n, total)
+}
+
+func collectIDs[T interface{ GetId() string }](
```
SergeyKanzhelev, line 686: I don't see templates every day. With this templatized, maybe worth adding a small unit test for it?

SergeyKanzhelev (approved):
/lgtm
/hold

holding in case you want to add a unit test for `collect`.

cmd/crictl/util.go
saschagrunert: Sure, pushed a unit test in aae32920.

saschagrunert:
@SergeyKanzhelev PTAL again

/unhold

SergeyKanzhelev (approved):
/lgtm

PR #2092, opened 2026-05-22
align local containerized test with CI

SergeyKanzhelev:
#### What type of PR is this?

/kind cleanup

I want to test with containerd 1.7 locally

```release-note
NONE
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -145,6 +145,34 @@ Or for `crictl` e2e:
 make test-crictl-e2e-containerd TESTFLAGS='--ginkgo.focus="pull"'
 ```
 
+##### Selecting the containerd version
+
+By default the image builds containerd from source at the `main` git ref,
+matching CI's primary matrix entry (NRI enabled, `--parallel=8`). Pass
+`CONTAINERD_VERSION` to test another ref, and `RUNC_FLAVOR` / `RUNTIME` to
+change the runtime:
+
+```bash
+make test-critest-containerd CONTAINERD_VERSION=release/1.7
+make test-critest-containerd CONTAINERD_VERSION=main RUNC_FLAVOR=crun
+```
+
+critest runs with `--parallel=8` by default (mirroring CI); override with
+`make test-critest-containerd PARALLEL=4`.
+
+The built image is tagged per version (`containerd-local-test:<version>`, with
+`-<flavor>` appended for a non-default `RUNC_FLAVOR`, e.g.
+`containerd-local-test:main-crun`) and each version gets its own data volume,
+so images are cached and reused across runs — only a changed
+`CONTAINERD_VERSION`/`RUNC_FLAVOR` triggers a rebuild. Set `FORCE_REBUILD=1` to
+force a fresh build that bypasses the Docker layer cache (`--no-cache --pull`),
+which is needed to refetch a moving ref such as `main`. To delete the cached
+images and volumes so they are regenerated:
+
+```bash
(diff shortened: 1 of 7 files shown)


hack/run-e2e-container.sh
```diff
@@ -24,24 +24,84 @@ cd "${REPO_ROOT}"
...
+        -t "${IMAGE_NAME}" \
+        -f images/containerd-local-test/Dockerfile .
+fi
 
 # If no command is provided, default to critest
 if [ $# -eq 0 ]; then
     set -- /usr/local/bin/critest-tools/critest --runtime-endpoint=unix:///run/containerd/containerd.sock
 fi
 
+# When NRI is enabled, inject --nri-socket into critest arguments so the
```
saschagrunert: This injects `--nri-socket` into all commands passed to the script, but `--nri-socket` is a flag registered only by `critest` (`pkg/framework/test_context.go:188`). The `crictl-e2e` binary uses a separate framework (`test/framework/framework.go`) that only defines `-crictl-binary-path` and `-crictl-runtime-endpoint`.

Since the default `CONTAINERD_VERSION=main` sets `ENABLE_NRI=true`, running `make test-crictl-e2e-containerd` would fail with `flag provided but not defined: -nri-socket`.

The old code handled this correctly by putting `--nri-socket` only on the `test-critest-containerd` Makefile target. Consider either guarding the injection on the binary name, or moving `--nri-socket` back into the Makefile target as a regular argument.

images/containerd-local-test/entrypoint.sh
```diff
@@ -43,10 +42,12 @@ CONTAINERD_PID=$!
...
+# Execute the command passed to the container. Capture the exit code without
+# tripping `set -e` so the containerd cleanup below always runs even when the
+# test command fails (mirrors CI's `|| TEST_RC=$?` pattern).
 echo "Executing: $@"
-"$@"
-EXIT_CODE=$?
+EXIT_CODE=0
+"$@" || EXIT_CODE=$?
 
 # Cleanup containerd
```
saschagrunert, line 52: Good fix. The old code (`"$@"` followed by `EXIT_CODE=$?`) would abort at the failing command under `set -e`, skipping the containerd cleanup below. This pattern correctly captures the exit code while letting cleanup proceed.

hack/run-e2e-container.sh
SergeyKanzhelev: fixed, thanks. I also renamed the cleanup target, I forgot about crictl =)

AGENTS.md
```diff
@@ -145,6 +145,34 @@ Or for `crictl` e2e:
...
+`-<flavor>` appended for a non-default `RUNC_FLAVOR`, e.g.
+`containerd-local-test:main-crun`) and each version gets its own data volume,
+so images are cached and reused across runs — only a changed
+`CONTAINERD_VERSION`/`RUNC_FLAVOR` triggers a rebuild. Set `FORCE_REBUILD=1` to
+force a fresh build that bypasses the Docker layer cache (`--no-cache --pull`),
+which is needed to refetch a moving ref such as `main`. To delete the cached
+images and volumes so they are regenerated:
+
+```bash
+make clean-critest-containerd-images
```
saschagrunert: The Makefile target is `clean-containerd-test-images`, not `clean-critest-containerd-images`.

SergeyKanzhelev: fixed, thank you

saschagrunert (approved):

PR #2099, opened 2026-05-26
CRI List Streaming Operations

bitoku:
#### What type of PR is this?

/kind feature

- https://github.com/kubernetes/enhancements/issues/5825

This PR makes crictl and critest use streaming RPCs by default.

Assisted-by: Claude

```release-note
Use streaming RPCs by default in crictl and critest.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ CLAUDE.md
 !.github/
 !.skills/
 
+crictl
--- a/cmd/crictl/config.go
+++ b/cmd/crictl/config.go
@@ -53,7 +53,8 @@ CRICTL OPTIONS:
 	 debug:	Enable debug output (default: false)
 	 pull-image-on-create:	Enable pulling image on create requests (default: false)
 	 disable-pull-on-run:	Disable pulling image on run requests (default: false)
-	 max-retries:	Max retries for connecting to an explicitly set endpoint (default: 3, 0 to disable, negative for infinite)`,
+	 max-retries:	Max retries for connecting to an explicitly set endpoint (default: 3, 0 to disable, negative for infinite)
+	 enable-streaming:	Enable streaming API for list operations (default: false)`,
 	UseShortOptionHandling: true,
 	Flags: []cli.Flag{
 		&cli.StringFlag{
@@ -99,6 +100,8 @@ CRICTL OPTIONS:
 				fmt.Println(config.DisablePullOnRun)
 			case common.MaxRetries:
 				fmt.Println(config.MaxRetries)
+			case common.EnableStreaming:
+				fmt.Println(config.EnableStreaming)
 			default:
 				return fmt.Errorf("no configuration option named %s", get)
 			}
@@ -134,6 +137,7 @@ CRICTL OPTIONS:
 			display.AddRow([]string{common.PullImageOnCreate, strconv.FormatBool(config.PullImageOnCreate)})
 			display.AddRow([]string{common.DisablePullOnRun, strconv.FormatBool(config.DisablePullOnRun)})
 			display.AddRow([]string{common.MaxRetries, strconv.Itoa(config.MaxRetries)})
+			display.AddRow([]string{common.EnableStreaming, strconv.FormatBool(config.EnableStreaming)})
 			display.ClearScreen()
 			display.Flush()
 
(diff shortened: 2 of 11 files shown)


cmd/crictl/config.go
```diff
@@ -53,7 +53,8 @@ CRICTL OPTIONS:
 	 debug:	Enable debug output (default: false)
 	 pull-image-on-create:	Enable pulling image on create requests (default: false)
 	 disable-pull-on-run:	Disable pulling image on run requests (default: false)
-	 max-retries:	Max retries for connecting to an explicitly set endpoint (default: 3, 0 to disable, negative for infinite)`,
+	 max-retries:	Max retries for connecting to an explicitly set endpoint (default: 3, 0 to disable, negative for infinite)
+	 enable-streaming:	Enable streaming API for list operations (default: false)`,
```
SergeyKanzhelev, line 57: this will be a temporary setting, right? We will switch to streaming everywhere. Do you see this as "try streaming first" setting or like a "force-streaming"?

In general I see no use for this configuration parameter as crictl users should not care

pkg/validate/streaming_list.go
```diff
@@ -0,0 +1,348 @@
...
+	BeforeEach(func(ctx SpecContext) {
+		client, err := framework.LoadCRIClientWithStreaming()
+		Expect(err).NotTo(HaveOccurred())
+
+		rc = client.CRIRuntimeClient
+		ic = client.CRIImageClient
+
+		version, err := rc.Version(ctx, "")
+		Expect(err).NotTo(HaveOccurred())
+		// TODO: remove once containerd supports streaming list APIs
```
SergeyKanzhelev, line 46: let's not skip based on runtime name and will skip based on whether it works or not

pkg/validate/streaming_list.go
```diff
@@ -0,0 +1,348 @@
...
+)
+
+var _ = framework.KubeDescribe("Streaming List", func() {
+	var (
+		rc internalapi.RuntimeService
+		ic internalapi.ImageManagerService
+	)
+
+	BeforeEach(func(ctx SpecContext) {
+		client, err := framework.LoadCRIClientWithStreaming()
```
SergeyKanzhelev, line 38: as far as I remember setting streaming to true will still allow to fallback to non-streaming. Is it right? If this is true, is only option for this test is to go to gRPC level?

SergeyKanzhelev (commented):
when I am thinking about tests for Streaming, I am thinking of a conformance tests that will exceed the size of a regular List calls. Introducing the conformance test that ensures that a certain number of pods/images/etc. is supported will be a good way to validate that List is working. It may be hard though and requires some experimentation

bitoku:
@SergeyKanzhelev That makes sense. Thank you for the review.
I'll redesign it to use streaming by default and add a conformance test to verify if it hits grpc size limit.

.gitignore
```diff
@@ -45,3 +45,4 @@ CLAUDE.md
 !.github/
 !.skills/
 
+crictl
```
saschagrunert, line 48: `/crictl` at line 37 already ignores the root binary. This entry (without the leading slash) is either redundant or unintentionally broader in scope. Should be dropped.

pkg/framework/util.go
```diff
@@ -160,6 +160,36 @@ func LoadCRIClient() (*InternalAPIClient, error) {
 	}, nil
 }
 
+// LoadCRIClientWithStreaming creates an InternalAPIClient with streaming
+// list operations explicitly enabled, regardless of the --enable-streaming flag.
+func LoadCRIClientWithStreaming() (*InternalAPIClient, error) {
```
saschagrunert, line 165: This is a near-exact duplicate of `LoadCRIClient()` above, differing only in the final boolean argument. Consider extracting a shared helper:

```go
func loadCRIClient(streaming bool) (*InternalAPIClient, error) { ... }
func LoadCRIClient() (*InternalAPIClient, error) { return loadCRIClient(false) }
func LoadCRIClientWithStreaming() (*InternalAPIClient, error) { return loadCRIClient(true) }
```

If streaming becomes the default per the redesign, this helper may not be needed at all.

saschagrunert (commented):
Two minor items on top of the existing review feedback.

SergeyKanzhelev (approved):
/approve
/lgtm

bitoku:
@SergeyKanzhelev Could you rerun [critest containerd / main / windows-2025 / containerd-shim-runhcs-v1 / runc (pull_request)](https://github.com/kubernetes-sigs/cri-tools/actions/runs/34366336947/job/102670536911?pr=2099) ?

PR #2102, opened 2026-05-27
enable zizmor

SergeyKanzhelev:
#### What type of PR is this?

/kind cleanup

zizmor seems to be used for supply chain attack prevention thru GHA. Addressed most of the findings

I am not sure how trustworthy the zizmor is. We can discuss in this PR or I can start an issue.

Some AI was used

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,13 +16,16 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
+          persist-credentials: false
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: go.mod
+          cache: false
       - run: make release-notes release
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
-      - uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0
+      # ncipollo/release-action handles create-or-update atomically
+      - uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0 # zizmor: ignore[superfluous-actions]
         with:
           allowUpdates: true
           artifacts: _output/releases/*
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,6 +91,29 @@ To automatically fix formatting issues, run:
 make prettier-fix
 ```
 
+### Zizmor
+
+Zizmor is a static analyzer for GitHub Actions workflows. It runs as part
+of `make verify` (and the `linters` CI matrix), and new findings will fail
+the build.
+
+To install zizmor locally, run:
+
+```bash
+make install.zizmor
+```
+
+To run zizmor against `.github/workflows/`, run:
+
+```bash
+make verify-zizmor
+```
+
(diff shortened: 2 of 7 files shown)


AGENTS.md
```diff
@@ -91,6 +91,30 @@ To automatically fix formatting issues, run:
...
+To run zizmor against `.github/workflows/`, run:
+
+```bash
+make verify-zizmor
+```
+
+There is no auto-fix target — findings must be fixed by hand. When a
+finding is a confirmed false positive, suppress it with an inline
+`# zizmor: ignore[<rule>]` comment on the offending line, and put a short
+justification on the line above (see `.github/workflows/containerd.yml`
```
saschagrunert: nit: containerd.yml has no `zizmor: ignore` comments. The only ignore is in `release.yml`. Should this reference point there instead?

.github/workflows/release.yml
```diff
@@ -16,13 +16,15 @@ jobs:
...
+          persist-credentials: false
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: go.mod
+          cache: false
       - run: make release-notes release
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
-      - uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0
+      - uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0 # zizmor: ignore[superfluous-actions] ncipollo/release-action handles create-or-update atomically
```
saschagrunert: nit: The AGENTS.md guidance added in this PR says to "put a short justification on the line above," but here the justification is on the same line, making it ~150 chars. Consider splitting:
```yaml
      # ncipollo/release-action handles create-or-update atomically
      - uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0 # zizmor: ignore[superfluous-actions]
```

Makefile
```diff
@@ -200,6 +203,32 @@ verify-dependencies: $(BUILD_BIN_PATH)/zeitgeist ## Verify third party dependenc
...
 
+.PHONY: verify-zizmor
+verify-zizmor: $(ZIZMOR) ## Run zizmor on .github/workflows/.
+	@if [ -x "$(ZIZMOR)" ]; then \
+		$(ZIZMOR) .github/workflows/; \
+	else \
+		echo "Skipping verify-zizmor: no zizmor binary for $(GOOS)/$(GOARCH)."; \
+	fi
+
+$(ZIZMOR): $(BUILD_BIN_PATH)
```
saschagrunert, line 214: nit: On unsupported platforms (e.g. FreeBSD) this recipe exits 0 without creating the binary, so make will re-run the recipe on every `make verify` invocation. Minor, but could be avoided by touching a sentinel file in the unsupported-platform case, or by moving the platform guard into `verify-zizmor` itself and dropping the `$(ZIZMOR)` prerequisite.

saschagrunert (commented):
LGTM, just three nits inline.

Makefile
SergeyKanzhelev: added touch

.github/workflows/release.yml
SergeyKanzhelev: right, fixed

AGENTS.md
SergeyKanzhelev: removed the see comment

saschagrunert (approved):

Issue #2046, opened 2026-04-10
critest for NRI integration


SergeyKanzhelev:
NRI is being used more and more. I am trying to understand where the NRI integration testing must reside. Does it belong to NRI repository or we should define some runtime conformance using critest.

I special case NRI because it is becoming increasingly part of a Pod lifecycle. CNI behavior may be considered the same way. And it will be a fair question, so let's discuss in this issue.

Examples of tests:

### NRI is actually called

Validating that NRI plugin is actually called with the proper arguments. This seems to be something that NRI may test for already.

### failure handling

Various tests where NRI plugin fails on sandbox and container creation as well as removal. Keeping the sandbox around if removal failed seems like a good conformance test. Same as propagating NRI errors up to creation calls. The key here is what runtime returns on NRI failure.

BTW, is there an ordering guarantee and whether NRI plugin which ACK-ed the creation of sandbox will be notified that the sandbox is NOT being created because of a second NRI plugin.

### allowing the retry

The retry must be allowed by runtime immidiately after create sandgox has failed with NRI error. This should ensure that sandbox is cleaned up before create sandbox CRI returned.

###  re-entrance consistency/invalid arguments handling

NRI must not be called when invalid arguments were passed in creation commands. Like creating the "same" sandbox that already exists or some bad names. Container creation for non-existing sandbox should not result in NRI call.

BTW, we do not have a good set of critests for bad arguments handling today. It feels like a good candidate for additional testing.

### NRI must block the sandbox creation for real

Long running NRI plugin on sandbox creation should be treated as sandbox was not created yet. So it should not be returned when you list sandboxes not creation of containers should not be allowed on it.

### ordering of things created

This is tricky to validate things in https://github.com/containerd/nri/pull/286/ - like whether cgroup hierarchy already created and such. This may be part of NRI testing. But we should seriously consider what parts of it are ok to adopt as conformance

SergeyKanzhelev:
@samuelkarp @mikebrow @aojea @haircommander @saschagrunert

haircommander:
this is a cool idea! I find myself wondering if this is tying the CRI too close to an external API that technically has nothing to do with kube (though in practice it certainly does). I don't know if we need to go through the work to like move NRI into k8s sigs, for instance, but  I wonder what's the best way to make it feel officially part of k8s and not an impl detail of CRI impls

mikebrow:
> this is a cool idea! 

nod

[...]

PR #2104, opened 2026-05-28
StopPodSandbox first test case

SergeyKanzhelev:
#### What type of PR is this?

/kind feature

Continue on https://github.com/kubernetes-sigs/cri-tools/issues/2046. I think after this one all tests will have some difference in behavior or errors.

Some AI was used

```release-note
An NRI test case for StopPodSandbox.
```

### DIFF
--- a/pkg/validate/nri_linux.go
+++ b/pkg/validate/nri_linux.go
@@ -411,4 +411,170 @@ var _ = framework.KubeDescribe("NRI", func() {
 			}
 		})
 	})
+
+	Context("StopPodSandbox contract", Serial, func() {
+		var (
+			testStub    *NRITestStub
+			podID       string
+			podConfig   *runtimeapi.PodSandboxConfig
+			containerID string
+		)
+
+		AfterEach(func(ctx SpecContext) {
+			// Stop the stub first to unblock any hooks that may be holding
+			// a StopPodSandbox call, allowing it to complete.
+			if testStub != nil {
+				testStub.Cleanup()
+			}
+
+			if containerID != "" {
+				if err := rc.StopContainer(ctx, containerID, 0); err != nil {
+					framework.Logf("AfterEach: StopContainer(%s) failed: %v", containerID, err)
+				}
+
+				if err := rc.RemoveContainer(ctx, containerID); err != nil {
+					framework.Logf("AfterEach: RemoveContainer(%s) failed: %v", containerID, err)
+				}
+			}
+
+			if podID != "" {
+				if err := rc.StopPodSandbox(ctx, podID); err != nil {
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
@saschagrunert @haircommander whenever you have a chance to take a look at this one.

@aojea @mikebrow as well

pkg/validate/nri_linux.go
```diff
@@ -411,4 +411,167 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+
+			By("verifying all workload containers were stopped before the hook fired")
+
+			containers, listErr := rc.ListContainers(ctx, &runtimeapi.ContainerFilter{
+				PodSandboxId: podID,
+			})
+			Expect(listErr).NotTo(HaveOccurred(), "ListContainers during StopPodSandbox hook")
+
+			for _, c := range containers {
+				Expect(c.GetState()).To(Equal(runtimeapi.ContainerState_CONTAINER_EXITED),
```
saschagrunert, line 553: If the runtime removes the container (rather than just stopping it) before the hook fires, `ListContainers` returns an empty slice and this loop never executes, so the assertion passes vacuously without actually verifying anything.

Consider adding a non-empty check before the loop:
```suggestion
			Expect(containers).NotTo(BeEmpty(),
				"workload containers should still be listed (in EXITED state) during StopPodSandbox hook")

			for _, c := range containers {
```

SergeyKanzhelev: it makes sense. Added.

pkg/validate/nri_linux.go
```diff
@@ -411,4 +411,170 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+
+			// Configure stub to block during StopPodSandbox so the main
+			// goroutine can inspect runtime state while the hook is active.
+			// sync.Once guards against the hook being invoked multiple times
+			// (e.g., idempotent stop redelivery), which would otherwise panic
+			// on a double close of hookReached.
+			testStub.Plugin.OnStopPodSandbox = func(hookCtx context.Context, _ *nri.PodSandbox) error {
+				firstInvocation := false
+
+				hookOnce.Do(func() { firstInvocation = true })
```
haircommander, line 474: why not wrap alll the cleanup in the hookOnce, rather than tracking firstInvocation?

SergeyKanzhelev: do you mean this as a code style change or something more than that? Just trying to clarify if I need to check for anything once this fixed.

SergeyKanzhelev: Actually, making it as a variable is better as it will allow for the second call to complete without waiting while first is still blocked

SergeyKanzhelev:
@haircommander ok to merge this?

PR #2106, opened 2026-05-28
Replace Fatal calls with proper error returns in crictl

saschagrunert:
#### What type of PR is this?

/kind cleanup

Replaces all 8 `log.Fatalf`/`logrus.Fatal` calls in `cmd/crictl/` with proper error returns. Extracts `run() error` from `main()` so deferred cleanup (CPU profiling, tracer shutdown) runs before exit.

None

`convertContainerState`, `convertPodState`, and `jsonBuiltinTmplFunc` now return `(T, error)`.

```release-note
NONE
```

### DIFF
--- a/cmd/crictl/container.go
+++ b/cmd/crictl/container.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 	goruntime "runtime"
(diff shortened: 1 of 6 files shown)


cmd/crictl/container.go
```diff
@@ -1222,7 +1221,7 @@ func ListContainers(ctx context.Context, runtimeClient internalapi.RuntimeServic
 			st.State = pb.ContainerState_CONTAINER_UNKNOWN
 			filter.State = st
 		default:
-			log.Fatalf("--state should be one of created, running, exited or unknown")
```
SergeyKanzhelev, line 1225: I wonder if linter can be configured to not allow these. It is so much better with a clear error path instead of fatalf

SergeyKanzhelev (approved):
/lgtm

it will be great to have a linter for this

PR #2108, opened 2026-05-28
Add --max-retries flag for connection retry with backoff

saschagrunert:
#### What type of PR is this?

/kind feature

Adds `--max-retries` flag (default: 3) with exponential backoff (500ms-5s) for crictl connections to explicitly configured endpoints. Also available via config file and `crictl config`. Set to 0 to disable retries.

None

Retry only applies to explicit endpoints, not the deprecated default-endpoint fallback. The `newCrictlConfig` refactor into two functions satisfies the nestif linter.

```release-note
Add --max-retries flag for crictl connection retry with exponential backoff (default: 3).
```

### DIFF
--- a/cmd/crictl/config.go
+++ b/cmd/crictl/config.go
@@ -52,7 +52,8 @@ CRICTL OPTIONS:
 	 timeout:	Timeout of connecting to server (default: 2)
 	 debug:	Enable debug output (default: false)
 	 pull-image-on-create:	Enable pulling image on create requests (default: false)
-	 disable-pull-on-run:	Disable pulling image on run requests (default: false)`,
+	 disable-pull-on-run:	Disable pulling image on run requests (default: false)
+	 max-retries:	Max retries for connecting to an explicitly set endpoint (default: 3, 0 to disable, negative for infinite)`,
 	UseShortOptionHandling: true,
 	Flags: []cli.Flag{
 		&cli.StringFlag{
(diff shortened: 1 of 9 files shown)


cmd/crictl/config.go
```diff
@@ -52,7 +52,8 @@ CRICTL OPTIONS:
 	 timeout:	Timeout of connecting to server (default: 2)
 	 debug:	Enable debug output (default: false)
 	 pull-image-on-create:	Enable pulling image on create requests (default: false)
-	 disable-pull-on-run:	Disable pulling image on run requests (default: false)`,
+	 disable-pull-on-run:	Disable pulling image on run requests (default: false)
+	 max-retries:	Max retries for connecting to the server (default: 3, 0 to disable)`,
```
SergeyKanzhelev: should we reserve negative values to "forever"?

SergeyKanzhelev: also docs should say that this only works when endpoint is set explicitly. Do we need arguments validation for this?

saschagrunert: Done, negative values now mean retry forever (only stopped by context cancellation).

Updated all docs (CLI flag, config help, docs/crictl.md) to clarify this only works with explicitly set endpoints. No argument validation needed since any integer is valid: positive = finite retries, 0 = disabled, negative = infinite.

SergeyKanzhelev (approved):
/lgtm

PR #2109, opened 2026-05-28
linter to block fatalf

SergeyKanzhelev:
#### What type of PR is this?

/kind cleanup

Follow up from https://github.com/kubernetes-sigs/cri-tools/pull/2106 @saschagrunert wdyt?

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,7 @@ linters:
     - exhaustive
     - exptostd
     - fatcontext
+    - forbidigo
     - forcetypeassert
     - funcorder
     - ginkgolinter
@@ -101,7 +102,6 @@ linters:
     # - depguard
     # - err113
     # - exhaustruct
-    # - forbidigo
     # - funlen
     # - gochecknoglobals
     # - gochecknoinits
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -124,6 +124,15 @@ linters:
...
+    forbidigo:
+      analyze-types: false
+      forbid:
+        - pattern: ^logrus\.Fatal.*$
+          msg: "use error returns instead of logrus.Fatal*; let main() print and exit with non-zero (see PR #2106)"
+        - pattern: ^log\.Fatal.*$
+          msg: "use error returns instead of log.Fatal*; let main() print and exit with non-zero (see PR #2106)"
+        - pattern: ^klog\.Fatal.*$
+          msg: "use error returns instead of klog.Fatal*; let main() print and exit with non-zero (see PR #2106)"
     gocognit:
```
saschagrunert, line 138: Worth adding `os.Exit` too? It has the same problem (skips deferred cleanup). There are two existing calls in non-test code (`main.go:62` and `util.go:79`), both intentional, so they'd need an exclusion or `//nolint` annotation.

```suggestion
        - pattern: ^klog\.Fatal.*$
          msg: "use error returns instead of klog.Fatal*; let main() print and exit with non-zero (see PR #2106)"
        - pattern: ^os\.Exit.*$
          msg: "use error returns instead of os.Exit; let main() handle the exit (see PR #2106)"
```

SergeyKanzhelev: added

saschagrunert (approved):
Love it, thank you!

PR #2118, opened 2026-06-04
container lifecycle NRI tests

SergeyKanzhelev:
#### What type of PR is this?

/kind feature

```release-note
Basic test in critest for container-related NRI callbacks.
```

### DIFF
--- a/pkg/validate/nri_linux.go
+++ b/pkg/validate/nri_linux.go
@@ -157,6 +157,178 @@ var _ = framework.KubeDescribe("NRI", func() {
 		})
 	})
 
+	Context("container lifecycle", Serial, func() {
+		var (
+			testStub    *NRITestStub
+			podID       string
+			podConfig   *runtimeapi.PodSandboxConfig
+			containerID string
+		)
+
+		BeforeEach(func(ctx SpecContext) {
+			var err error
+
+			testStub, err = StartNRITestStub("cri-test-nri-ctr-lifecycle", "00")
+			Expect(err).NotTo(HaveOccurred(), "failed to start NRI test stub")
+
+			// Ensure test image is available
+			framework.PullPublicImage(ctx, ic, framework.TestContext.TestImageList.DefaultTestContainerImage, nil)
+		})
+
+		AfterEach(func(ctx SpecContext) {
+			if containerID != "" {
+				if err := rc.StopContainer(ctx, containerID, 0); err != nil {
+					framework.Logf("AfterEach: StopContainer(%s) failed: %v", containerID, err)
+				}
+
+				if err := rc.RemoveContainer(ctx, containerID); err != nil {
+					framework.Logf("AfterEach: RemoveContainer(%s) failed: %v", containerID, err)
+				}
+			}
+
+			if podID != "" {
+				if err := rc.StopPodSandbox(ctx, podID); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,163 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+					stopEvent = &containerEvents[i]
+				case EventRemoveContainer:
+					removeEvent = &containerEvents[i]
+				case EventRunPodSandbox, EventStopPodSandbox, EventRemovePodSandbox:
+					// Pod events not verified in this test. Events are expected, but test ignores them
+				}
+			}
+
+			Expect(createEvent).NotTo(BeNil(), "CreateContainer event not received")
+			// SPEC_DISCREPANCY: CRI-O does not populate container name in NRI CreateContainer event metadata
```
SergeyKanzhelev: filed https://github.com/cri-o/cri-o/issues/10011

cc @haircommander

SergeyKanzhelev:

pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,163 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			if createEvent.ContainerName == "" {
+				Skip("spec discrepancy: runtime does not populate container name in NRI CreateContainer event metadata")
```
saschagrunert: Ginkgo's `Skip()` terminates the entire `It` block, so this also skips the `startEvent`/`stopEvent`/`removeEvent` nil checks, their `ContainerID` validation, and the timestamp ordering checks (lines 289-310). None of those depend on `ContainerName` being populated.

Consider guarding only the name assertion so the rest of the test still runs on CRI-O:

```go
if createEvent.ContainerName != "" {
    Expect(createEvent.ContainerName).To(Equal(containerName))
}
```

pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,163 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+				case EventCreateContainer:
+					createEvent = &containerEvents[i]
+				case EventStartContainer:
+					startEvent = &containerEvents[i]
+				case EventStopContainer:
+					stopEvent = &containerEvents[i]
+				case EventRemoveContainer:
+					removeEvent = &containerEvents[i]
+				case EventRunPodSandbox, EventStopPodSandbox, EventRemovePodSandbox:
+					// Pod events not verified in this test. Events are expected, but test ignores them
```
saschagrunert: Dead code: `containerEvents` is built by filtering for `ContainerID != ""` (line 253), and `recordPodEvent()` never sets `ContainerID` (only `recordContainerEvent()` does). Pod events cannot appear in `containerEvents`, so this case is unreachable.

The analogous pod lifecycle test (line 130) calls `Fail()` on unexpected event types. Consider removing this case and adding a `default` with `Fail()` to catch truly unexpected events.

saschagrunert (commented):
Overall this is solid test code that follows the existing patterns well. Two observations below.

pkg/validate/nri_linux.go
SergeyKanzhelev: fixed

pkg/validate/nri_linux.go
SergeyKanzhelev: added Fail

saschagrunert (approved):

PR #2125, opened 2026-06-15
container lifecycle failed container

SergeyKanzhelev:
#### What type of PR is this?

/kind feature

Similar to https://github.com/kubernetes-sigs/cri-tools/pull/2118 but check that Stop event will be send to container stopped by itself.

Some AI was used

```release-note
NRI test in critest to ensure that Stop event will be sent for container that exited by itself.
```

### DIFF
--- a/pkg/validate/nri_linux.go
+++ b/pkg/validate/nri_linux.go
@@ -329,6 +329,205 @@ var _ = framework.KubeDescribe("NRI", func() {
 		})
 	})
 
+	Context("container lifecycle (failed container)", Serial, func() {
+		var (
+			testStub    *NRITestStub
+			podID       string
+			podConfig   *runtimeapi.PodSandboxConfig
+			containerID string
+		)
+
+		BeforeEach(func(ctx SpecContext) {
+			var err error
+
+			testStub, err = StartNRITestStub("cri-test-nri-ctr-fail", "00")
+			Expect(err).NotTo(HaveOccurred(), "failed to start NRI test stub")
+
+			// Ensure test image is available
+			framework.PullPublicImage(ctx, ic, framework.TestContext.TestImageList.DefaultTestContainerImage, nil)
+		})
+
+		AfterEach(func(ctx SpecContext) {
+			if containerID != "" {
+				if err := rc.StopContainer(ctx, containerID, 0); err != nil {
+					framework.Logf("AfterEach: StopContainer(%s) failed: %v", containerID, err)
+				}
+
+				if err := rc.RemoveContainer(ctx, containerID); err != nil {
+					framework.Logf("AfterEach: RemoveContainer(%s) failed: %v", containerID, err)
+				}
+			}
+
+			if podID != "" {
+				if err := rc.StopPodSandbox(ctx, podID); err != nil {
+					framework.Logf("AfterEach: StopPodSandbox(%s) failed: %v", podID, err)
+				}
+
+				if err := rc.RemovePodSandbox(ctx, podID); err != nil {
+					framework.Logf("AfterEach: RemovePodSandbox(%s) failed: %v", podID, err)
+				}
+			}
+
+			if testStub != nil {
+				testStub.Cleanup()
+			}
+		})
+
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:

saschagrunert (commented):
Thanks for adding this, the test design is solid.

pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,186 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+					framework.Logf("AfterEach: RemovePodSandbox(%s) failed: %v", podID, err)
+				}
+			}
+
+			if testStub != nil {
+				testStub.Cleanup()
+			}
+		})
+
+		It("should receive CreateContainer, StartContainer, and RemoveContainer (and StopContainer when the runtime delivers it) for a container that exits with a non-zero code", func(ctx SpecContext) {
```
saschagrunert: The title says StopContainer is conditional ("when the runtime delivers it"), but the test requires it via `Eventually(...).Should(BeTrue())`. Consider:
```suggestion
		It("should receive CreateContainer, StartContainer, StopContainer, and RemoveContainer for a container that exits with a non-zero code", func(ctx SpecContext) {
```

pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,186 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			status := getContainerStatus(ctx, rc, containerID)
+			Expect(status.GetExitCode()).To(BeEquivalentTo(1),
+				`container should have failed with exit code 1 (from sh -c "exit 1")`)
+
+			By("waiting for Create, Start, and Stop NRI events before removing the container")
+			// The runtime must deliver StopContainer on its own for a
+			// self-exited container, without an explicit CRI RemoveContainer
+			// call. Wait for all three events before calling RemoveContainer
+			// to prove that StopContainer is not a side effect of removal.
+			var containerEvents []NRIEvent
```
saschagrunert: Nit: `containerEvents` is populated inside the Eventually closure but never read after it returns. Can be removed.

pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,186 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+					}
+				}
+
+				return createEvent != nil && startEvent != nil && stopEvent != nil
+			}, 10*time.Second, 50*time.Millisecond).Should(BeTrue(),
+				"NRI stub did not receive Create, Start, and Stop events for container %s before removal", containerID)
+
+			By("verifying CreateContainer event fired with correct metadata")
+			Expect(createEvent.ContainerID).To(Equal(containerID))
+
```
saschagrunert, line 470: These three ContainerID assertions are tautological since events are already filtered by `e.ContainerID != containerID` in the Eventually loop above. Consider verifying `ContainerName` instead (like #2118 does, with the CRI-O skip for the known spec discrepancy) to get actual verification value here.

pkg/validate/nri_linux.go
```diff
@@ -157,6 +157,198 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			// event metadata. Record it here and skip at the very end so the remaining
+			// assertions (ContainerID, Start/Stop/Remove, ordering) still run.
+			skipForMissingName := false
+			if createEvent.ContainerName == "" {
+				skipForMissingName = true
+			} else {
+				Expect(createEvent.ContainerName).To(Equal(containerName))
+			}
+
+			Expect(createEvent.ContainerName).To(Equal(containerName))
```
saschagrunert: This assertion runs unconditionally, so it will fail on CRI-O where `ContainerName` is empty (the spec discrepancy the `skipForMissingName` mechanism above is designed to handle). Looks like a leftover from reworking the `ContainerID` checks.

```suggestion
```

SergeyKanzhelev: indeed the leftover, thanks

saschagrunert (approved):

saschagrunert:
Ah this one needs a rebase now.

saschagrunert (approved):

PR #2126, opened 2026-06-15
StopPodSandbox contract

SergeyKanzhelev:
#### What type of PR is this?

/kind feature

Stop pod sandbox contact with NRI

Some AI was used

```release-note
New cirtest to verify NRI callbacks during Pod sandbox stop API call handling
```

### DIFF
--- a/pkg/validate/nri_linux.go
+++ b/pkg/validate/nri_linux.go
@@ -576,5 +576,118 @@ var _ = framework.KubeDescribe("NRI", func() {
 				"RemovePodSandbox should succeed after stop")
 			podID = ""
 		})
+
+		It("should handle StopPodSandbox idempotently and never reuse sandbox", func(ctx SpecContext) {
+			// This test validates two spec guarantees:
+			// 1. StopPodSandbox is idempotent - calling it multiple times succeeds without error.
+			// 2. After Stop, the sandbox is never reused - CreateContainer fails.
+			var err error
+
+			testStub, err = StartNRITestStub("cri-test-nri-stop-idempotent", "00")
+			Expect(err).NotTo(HaveOccurred(), "failed to start NRI test stub")
+
+			By("creating a pod sandbox")
+
+			podSandboxName := "nri-test-stop-idempotent-" + framework.NewUUID()
+			uid := framework.DefaultUIDPrefix + framework.NewUUID()
+			namespace := framework.DefaultNamespacePrefix + framework.NewUUID()
+			podConfig = &runtimeapi.PodSandboxConfig{
+				Metadata: framework.BuildPodSandboxMetadata(podSandboxName, uid, namespace, framework.DefaultAttempt),
+				Linux: &runtimeapi.LinuxPodSandboxConfig{
+					CgroupParent: common.GetCgroupParent(ctx, rc),
+				},
+				Labels: framework.DefaultPodLabels,
+			}
+			podID = framework.RunPodSandbox(ctx, rc, podConfig)
+			Expect(podID).NotTo(BeEmpty())
+
+			By("calling StopPodSandbox the first time")
+			Expect(rc.StopPodSandbox(ctx, podID)).NotTo(HaveOccurred(),
+				"First StopPodSandbox call should succeed")
+
+			By("verifying the StopPodSandbox NRI hook fired exactly once")
+			// Poll for the StopPodSandbox event rather than assuming a fixed
+			// global event count, which would break if a future runtime/NRI
+			// version emits additional lifecycle events.
+			var stopPodSandboxEvents []NRIEvent
+
+			Eventually(func() []NRIEvent {
+				stopPodSandboxEvents = nil
+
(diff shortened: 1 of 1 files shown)


pkg/validate/nri_linux.go
```diff
@@ -576,5 +576,115 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			// CreateContainer on a stopped sandbox MUST fail per spec.
+			ctrID, createErr := rc.CreateContainer(ctx, podID, containerConfig, podConfig)
+			if createErr == nil {
+				// SPEC_DISCREPANCY: containerd allows CreateContainer on a stopped
+				// sandbox instead of rejecting it. Hand the unexpectedly created
+				// container to AfterEach for cleanup, then skip the non-reuse
+				// assertion (spec says the sandbox should never be reused after Stop).
+				containerID = ctrID
+
+				Skip("spec discrepancy: containerd allows CreateContainer on a stopped sandbox; " +
```
SergeyKanzhelev, line 671: bug for this: https://github.com/containerd/containerd/issues/13599

SergeyKanzhelev:

pkg/validate/nri_linux.go
```diff
@@ -576,5 +576,115 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			podID = framework.RunPodSandbox(ctx, rc, podConfig)
+			Expect(podID).NotTo(BeEmpty())
+
+			By("calling StopPodSandbox the first time")
+			Expect(rc.StopPodSandbox(ctx, podID)).NotTo(HaveOccurred(),
+				"First StopPodSandbox call should succeed")
+
+			By("verifying the StopPodSandbox NRI hook fired exactly once")
+			// Wait for RunPodSandbox + StopPodSandbox events from the first call.
+			events, err := testStub.Plugin.WaitForEventCount(2, 10*time.Second)
```
saschagrunert: `WaitForEventCount(2, ...)` assumes RunPodSandbox and StopPodSandbox are the only two events. If a future runtime/NRI version emits additional lifecycle events, this could return before the StopPodSandbox event arrives.

The pod lifecycle test at line 114 already accounts for this with `BeNumerically(">=" , 3)`. Consider polling `Events()` until a StopPodSandbox event for this pod appears (with a timeout), instead of relying on a global count of exactly 2.

pkg/validate/nri_linux.go
```diff
@@ -576,5 +576,115 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+				// sandbox instead of rejecting it. Hand the unexpectedly created
+				// container to AfterEach for cleanup, then skip the non-reuse
+				// assertion (spec says the sandbox should never be reused after Stop).
+				containerID = ctrID
+
+				Skip("spec discrepancy: containerd allows CreateContainer on a stopped sandbox; " +
+					"spec says sandbox should never be reused after Stop")
+			}
+
+			Expect(createErr).To(HaveOccurred(),
```
saschagrunert: This assertion is unreachable when `createErr == nil` (`Skip` exits the test) and tautologically true when `createErr != nil` (guaranteed by the if-guard above). Could just remove it -- the `if createErr == nil { Skip(...) }` already communicates the contract clearly.

saschagrunert (commented):
Nice test! Two small suggestions below.

pkg/validate/nri_linux.go
SergeyKanzhelev: added comment to not forget to remove it once containerd is fixed

pkg/validate/nri_linux.go
SergeyKanzhelev: good point. Fixed

pkg/validate/nri_linux.go
```diff
@@ -576,5 +576,119 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			containerEvents := 0
+
+			for _, e := range FilterEventsByPodID(allEventsAfterCreate, podID) {
+				if e.Type == EventCreateContainer {
+					containerEvents++
+				}
+			}
+
+			Expect(containerEvents).To(Equal(0),
+				"Failed CreateContainer on a stopped sandbox MUST NOT generate an NRI CreateContainer event")
```
saschagrunert: This uses `time.Sleep` + manual polling, but the identical pattern at line 635 (verifying no second StopPodSandbox event) already uses `Consistently`. Consider using the same approach here for consistency:

```suggestion
		Consistently(func() int {
			count := 0

			for _, e := range FilterEventsByPodID(testStub.Plugin.Events(), podID) {
				if e.Type == EventCreateContainer {
					count++
				}
			}

			return count
		}, 2*time.Second, 200*time.Millisecond).Should(Equal(0),
			"Failed CreateContainer on a stopped sandbox MUST NOT generate an NRI CreateContainer event")

SergeyKanzhelev: applied

saschagrunert (approved):

PR #2137, opened 2026-06-30
Late-joining plugin is synchronized with existing pods and containers

SergeyKanzhelev:
#### What type of PR is this?

/kind feature

This seems to be not handled at all in containerd, but I think it will be important for consistency of NRI plugins.

I checked on containerd - it fails so I mentioned it in spec discrepancy. CI will validate on CRI-O

https://github.com/kubernetes-sigs/cri-tools/issues/2046

Some AI was used

```release-note
Test Synchronize config and race conditions
```

### DIFF
--- a/pkg/validate/nri_linux.go
+++ b/pkg/validate/nri_linux.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/validate/nri_util_linux.go
```diff
@@ -66,8 +66,20 @@ type NRITestPlugin struct {
 	ready     chan struct{}
 	readyOnce sync.Once
 
+	// syncPods/syncContainers capture the pod and container IDs passed to the
+	// most recent Synchronize call, so tests can assert that a late-joining
+	// plugin is reconciled with the runtime's existing state.
+	syncPods       []string
```
saschagrunert, line 72: `Reset()` (called by `Cleanup()`) clears `events` but not these new fields. Consider adding `p.syncPods = nil; p.syncContainers = nil` to `Reset()` so state is fully cleared.

Not a bug today since each test creates a fresh stub, but would bite if a stub is ever reused across sub-specs.

pkg/validate/nri_linux.go
```diff
@@ -1061,4 +1062,580 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+			// spec-compliant; what is forbidden is losing it entirely.
+			//
+			// SPEC_DISCREPANCY: containerd (as of main / 2.x) may lose containers
+			// created while a late-joining plugin's Synchronize is in progress —
+			// they appear in neither the Synchronize set nor as a CreateContainer
+			// callback. The NRI spec requires that no container is lost during
+			// plugin initialization, but containerd does not yet implement this
+			// guarantee. Poll and Skip rather than hard-failing so the remaining
+			// test suite still runs.
+			pluginSeesContainer := func() bool {
```
saschagrunert: Nit: tests 2 and 3 duplicate the poll-then-Skip pattern (and the `syncReached`/`syncRelease`/`syncOnce` coordination setup). The two poll loops differ slightly (single condition vs per-item), but a small helper could still cut the repetition. Not blocking, just something to consider if more sync tests get added.

pkg/validate/nri_linux.go
```diff
@@ -1061,4 +1062,580 @@ var _ = framework.KubeDescribe("NRI", func() {
...
+
+				select {
+				case <-timeout:
+					break poll
+				case <-ticker.C:
+				}
+			}
+
+			if len(lostContainers) > 0 {
+				Skip(fmt.Sprintf("spec discrepancy: containerd lost %d container(s) created "+
```
saschagrunert, line 1920: When the test passes (no containers lost), nothing is logged. If containerd fixes this, CI won't show the change and nobody will know the `Skip` can be removed. A small log on the success path would help:

```go
framework.Logf("all %d containers reached the late-joining plugin", len(allIDs))
```

SergeyKanzhelev:
addressed all comments

SergeyKanzhelev:
- Containerd: https://github.com/containerd/containerd/issues/13698
- CRI-O: https://github.com/cri-o/cri-o/issues/10090

Since both runtimes do not implement it, we can discuss if this indeed is a requirement. I believe it should be, but offering an alternative solution for NRI plugin authors to reconcile the running sandboxes and containers would also be acceptable solution

saschagrunert (approved):

PR #2144, opened 2026-07-08
fix: allow crictl config to create a missing explicit file

### LINKED ISSUE #2143: crictl config fails to create a missing explicit config file
#### What happened:

`crictl --config /tmp/crictl.yaml config --set debug=true` fails if the file does not exist yet, even though the `config` command is supposed to bootstrap it.

The command should create the file and write the requested setting.

```bash
tmpdir=$(mktemp -d)
crictl --config "$tmpdir/crictl.yaml" config --set debug=true
```

Current result is a `get server config` error before the `config` command runs.

This is a normal local CLI workflow. No special scale or boundary conditions are needed.

#### Environment:

Any host where `crictl` is run with a custom config path that does not exist yet.

immanuwell:
#### What type of PR is this?

/kind bug

Fixes a small but real `crictl config` paper cut.

If you pass a custom `--config` path and the file is missing, `config --set` dies in startup before the `config` command can create the file. That makes the bootstrap flow kinda busted.

This patch lets `config --set` and the positional `config key value` path create the missing file, and keeps `config --get` and `config --list` read-only.

Repro:

```bash
tmpdir=$(mktemp -d)
crictl --config "$tmpdir/crictl.yaml" config --set debug=true
```

Before this PR, that fails with a `get server config` error.

Fixes #2143

Checked with the built `crictl` binary and the existing `crictl: config` e2e suite. This bug is easy to hit in normal local CLI use, no weird limits or edge conditions needed.

```release-note
`crictl config --set` and `crictl config <key> <value>` now create a missing explicit `--config` file instead of failing during startup.
```

### DIFF
--- a/.github/workflows/crio.yml
+++ b/.github/workflows/crio.yml
@@ -141,6 +141,7 @@ jobs:
 
           set +o errexit
           sudo -E PATH=$PATH make test-e2e \
+            GINKGO_FLAGS="--flake-attempts=3" \
             TESTFLAGS="-crictl-runtime-endpoint=unix://var/run/crio/crio.sock"
           TEST_RC=$?
           set -o errexit
--- a/cmd/crictl/config.go
+++ b/cmd/crictl/config.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -72,14 +73,26 @@ CRICTL OPTIONS:
 	Action: func(c *cli.Context) error {
 		configFile := c.String("config")
 		if _, err := os.Stat(configFile); err != nil {
-			if err := common.WriteConfig(nil, configFile); err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
 				return err
 			}
+
+			if shouldCreateConfigFile(c) {
+				if err := common.WriteConfig(nil, configFile); err != nil {
+					return err
+				}
+			}
 		}
-		// Get config from file.
-		config, err := common.ReadConfig(configFile)
-		if err != nil {
-			return fmt.Errorf("load config file: %w", err)
+
+		config := &common.Config{}
+		if _, err := os.Stat(configFile); err == nil {
+			// Get config from file.
+			config, err = common.ReadConfig(configFile)
+			if err != nil {
+				return fmt.Errorf("load config file: %w", err)
+			}
+		} else if !errors.Is(err, os.ErrNotExist) {
+			return err
 		}
 
 		if c.IsSet("get") {
@@ -155,6 +168,18 @@ CRICTL OPTIONS:
 
 		return common.WriteConfig(config, configFile)
 	},
+}
+
+func shouldCreateConfigFile(c *cli.Context) bool {
+	if c.IsSet("set") {
+		return true
+	}
+
+	if c.IsSet("get") || c.Bool("list") {
+		return false
+	}
+
+	return c.Args().First() != ""
 }
 
 func setValue(key, value string, config *common.Config) error {
--- a/cmd/crictl/config_command_test.go
+++ b/cmd/crictl/config_command_test.go
@@ -1,0 +1,125 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+)
+
+var runWithArgsForTestMu sync.Mutex
+
+func TestConfigCommandCreatesMissingExplicitConfigOnSet(t *testing.T) {
+	t.Parallel()
+
+	configPath := filepath.Join(t.TempDir(), "nested", "crictl.yaml")
+
+	err := runWithArgsForTest(
+		t,
+		"crictl",
+		"--config",
+		configPath,
+		"config",
+		"--set",
+		"debug=true",
+	)
+	if err != nil {
+		t.Fatalf("run config --set: %v", err)
+	}
+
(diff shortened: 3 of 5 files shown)


cmd/crictl/main.go
SergeyKanzhelev, line 229: the config is nil here if file was not found. Is it safe?

cmd/crictl/main.go
```diff
@@ -213,9 +214,11 @@ func run() error {
 			return fmt.Errorf("get executable path: %w", err)
 		}
 
-		if config, err = common.GetServerConfigFromFile(context.String("config"), exePath); err != nil {
-			if context.IsSet("config") {
-				return fmt.Errorf("get server config: %w", err)
+		if !shouldSkipConfigLoad(context) {
```
SergeyKanzhelev: maybe a better fix is to create an empty file here if config set command is called. And keep all other logic the same. This will avoid a lot of additional file existence checks.

SergeyKanzhelev:
> Fixes a small but real `crictl config` paper cut.

So the scenario is downloading crictl, and use config set commands to set all properties? I wonder if we need a new command that creates an empty config file instead. I think the `crictl config` were supposed to do it. Is it a regression? I wonder if I induced it when introduced global span for tracing. Would you be willing to see if this is a regression?

cmd/crictl/main.go
immanuwell: changed the implementation so the file is created before GetServerConfigFromFile runs for config write operations

cmd/crictl/main.go
immanuwell: reworked it to create the missing explicit file before the existing config-loading logic, only for config --set and positional config key-value writes

immanuwell:
@SergeyKanzhelev addressed your suggestions, please take a look

yong-jie-gong:
- --help/-h: shoud not throw error simply

saschagrunert (commented):
A few findings, mostly around the manual arg parsing and the split between `main.go` and `config.go`.

cmd/crictl/main.go
```diff
@@ -316,3 +323,31 @@ func run() error {
...
+	if !ctx.IsSet("config") || len(args) == 0 || args[0] != configCommand.Name {
+		return false
+	}
+
+	for _, arg := range args[1:] {
+		switch {
+		case arg == "--get" || strings.HasPrefix(arg, "--get="):
+			return false
+		case arg == "--list" || strings.HasPrefix(arg, "--list="):
+			return false
```
saschagrunert, line 338: The switch returns on the first matching arg. `--list=false` matches `strings.HasPrefix(arg, "--list=")` and returns `false`, even if `--set` follows later in the arg list. So `crictl --config /missing config --list=false --set debug=true` would not create the file and fail in `Before`.

The `--list` flag is a `BoolFlag`, so `--list=false` is valid urfave/cli v2 syntax meaning "don't list." The same issue applies to `--get=<val>` preceding `--set`.

Consider scanning all args before deciding, or restructuring so this manual parsing isn't needed (e.g. letting `Before` skip the `GetServerConfigFromFile` error when the subcommand is `config`, and having the Action handle file creation entirely).

cmd/crictl/main.go
```diff
@@ -213,6 +214,12 @@ func run() error {
 			return fmt.Errorf("get executable path: %w", err)
 		}
 
+		if shouldCreateMissingExplicitConfigFile(context) {
```
saschagrunert, line 217: This duplicates the write-vs-read decision that `shouldCreateConfigFile` in `config.go:173` also makes, but with a different mechanism (manual string matching vs `cli.Context`). The two can drift if flags are added or renamed.

One way to avoid the duplication: have `Before` skip the `GetServerConfigFromFile` error when the subcommand is `config`, and let the config command's Action decide whether to create the file or fail. That keeps all the logic in one place with access to the proper `cli.Context`.

cmd/crictl/config.go
```diff
@@ -72,14 +73,26 @@ CRICTL OPTIONS:
 	Action: func(c *cli.Context) error {
 		configFile := c.String("config")
 		if _, err := os.Stat(configFile); err != nil {
-			if err := common.WriteConfig(nil, configFile); err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
 				return err
 			}
+
+			if shouldCreateConfigFile(c) {
```
saschagrunert, line 80: For the explicit `--config` case this block is unreachable: write ops already have the file created by `Before` (`main.go:217-221`), so `os.Stat` at line 75 succeeds and this branch is skipped. Read ops fail in `Before` at `main.go:224-226`, so the Action never runs.

This only fires for the default config path (where `Before` swallows the missing-file error). If that's intentional, a comment would help. Otherwise, consider whether the `config.go` changes are needed at all given the `main.go` fix.

cmd/crictl/config.go
```diff
@@ -72,14 +73,26 @@ CRICTL OPTIONS:
...
+				}
+			}
 		}
-		// Get config from file.
-		config, err := common.ReadConfig(configFile)
-		if err != nil {
-			return fmt.Errorf("load config file: %w", err)
+
+		config := &common.Config{}
+		if _, err := os.Stat(configFile); err == nil {
```
saschagrunert, line 88: nit: this re-stats the same `configFile` that was just checked at line 75. If line 80-83 created the file, this always succeeds. If it didn't, this repeats the existence check. Threading a bool from the first block (or restructuring into a single flow) would avoid the redundant syscall and make the control flow clearer.

cmd/crictl/config_command_test.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		t.Fatalf("read config file: %v", err)
+	}
+
+	if !strings.Contains(string(content), "debug: true") {
+		t.Fatalf("expected config file to contain debug=true, got:\n%s", string(content))
+	}
+}
+
+func TestConfigCommandListDoesNotCreateMissingExplicitConfig(t *testing.T) {
```
saschagrunert, line 86: Worth adding a `TestConfigCommandGetDoesNotCreateMissingExplicitConfig` counterpart. `--get` and `--list` take separate code paths in `shouldCreateMissingExplicitConfigFile` (`main.go:335-336` vs `337-338`), so testing only one leaves the other unexercised.

.github/workflows/crio.yml
```diff
@@ -141,6 +141,7 @@ jobs:
 
           set +o errexit
           sudo -E PATH=$PATH make test-e2e \
+            GINKGO_FLAGS="--flake-attempts=3" \
```
saschagrunert, line 144: This and the `GINKGO_FLAGS` Makefile changes (`Makefile:63,271`) are unrelated to the config fix. Splitting them into a separate commit or PR would keep the review scope focused and git history clean.

Issue #2127, opened 2026-06-17
Backward compatibility is broken because of cri-api change


bitoku:
Due to this recent change (https://github.com/kubernetes/kubernetes/pull/139168#issuecomment-4501163149), crictl no longer accepts environment variables in plain text; it now requires them to be base64-encoded.

Here is a reproducer: https://gist.github.com/bitoku/749684a85b44018e7d5cd4c77cd47202

The question is: do we want to fix this incompatibility or just leave it as is?
The downside of keeping the compatibility is that it may require some custom unmarshalling and that we won't be able to test non-UTF8 binaries using crictl.

bitoku:
cc @SergeyKanzhelev @saschagrunert

SergeyKanzhelev:
can we actually fix it? Isn't the problem on the wire format?

bitoku:
@SergeyKanzhelev It's caused by the change of the wire format, but the problem itself is more of a parser. The parser now only accepts base64-encoded variables for the field because it's changed from `string` to `[]byte`.
I'd say it's not impossible, but it needs special treatment only for that field.

SergeyKanzhelev:
I didn't look close. After the cherry-picks in k/k, will we need to just bump dependencies? Are you suggesting fixing it and release a patch version for crictl?

I am open to have a special treatment for the field since it is a regression

locker95:
@SergeyKanzhelev's guess was right — a dependency bump was enough, and it has already landed on `master`. The chain:

1. kubernetes/kubernetes#139964 ("Restore string JSON encoding of cri-api KeyValue", merged 2026-06-24) added custom `MarshalJSON`/`UnmarshalJSON` to `KeyValue`, restoring the pre-1.34 plaintext semantics. Backported via #139967/#139966/#139965.
2. That shipped in `k8s.io/cri-api` v0.36.3.
3. cri-tools bumped v0.36.2 → v0.36.3 on 2026-07-24 (`895f236b`, #2152), vendoring `vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api_json.go`.

Verified by loading the same JSON config through `loadContainerConfig()` with only the cri-api version changed:

```
### cri-api v0.36.2 (as reported here) ###
  input="hello"    -> ERROR: illegal base64 data at input byte 4
  input="aGVsbG8=" -> OK Value="hello"
  input="a=b"      -> ERROR: illegal base64 data at input byte 1

[...]

PR #2154, opened 2026-07-25
crictl: Do not advertise base64 env values in JSON schema

### LINKED ISSUE #2153: crictl jsonschema still advertises base64 encoding for env values
#### What happened:

`crictl create jsonschema --container` (and `crictl run jsonschema --container`) advertises the `envs[].value` field as base64 encoded:

```json
"KeyValue": {
  "properties": {
    "key": {
      "type": "string"
    },
    "value": {
      "type": "string",
      "contentEncoding": "base64"
    }
  },
  "additionalProperties": false,
  "type": "object"
}
```

The container config parser (`loadContainerConfig`, via `sigs.k8s.io/yaml` → `encoding/json`) does not accept base64 there. Since the bump to `k8s.io/cri-api` v0.36.3 in 895f236b, `KeyValue` implements custom `MarshalJSON`/`UnmarshalJSON` (added upstream in kubernetes/kubernetes#139964 to fix #2127) which restores the pre-1.34 plain string encoding for `value`.

So the machine readable schema crictl publishes now contradicts crictl's own parser. A user or code generator that follows the schema and base64 encodes the value ends up with a container whose env var literally contains the base64 text:

```
value written in config: "aGVsbG8="   ->  parsed env value: "aGVsbG8="   (expected "hello")
value written in config: "hello"      ->  parsed env value: "hello"
```

No error is reported, the env var is just silently wrong — which is exactly the bad path #2127 complained about.

The cause is that the custom `MarshalJSON`/`UnmarshalJSON` on `KeyValue` only affects the `encoding/json` path. `printJSONSchema` uses `github.com/invopop/jsonschema`, which reflects over the Go struct and therefore still renders the underlying `[]byte` as `contentEncoding: base64`.

`envs[].value` in the generated JSON schema should be a plain `{"type": "string"}` with no `contentEncoding`, matching what the config parser accepts.

```console
$ go build -o ./crictl ./cmd/crictl
$ ./crictl create jsonschema --container | grep -A4 '"value"'
      "value": {
        "type": "string",
        "contentEncoding": "base64"
      }
    },
```

Then feed a config that follows that schema, e.g. `envs: [{key: GREETING, value: aGVsbG8=}]`, and observe that the resulting env value is the literal string `aGVsbG8=` rather than `hello`.

Only the reflection based `jsonschema` output is affected. `KeyValue` is an input-only message for crictl (it only appears in `ContainerConfig.envs`), so crictl never `protojson`-marshals it and the `protojson` path in `protobufObjectToJSON` has no user visible problem here.

This mismatch does not exist in the released v1.36.0 (which pins cri-api v0.36.0, where both the parser and the schema use base64), so it currently only affects `master`.

I have a fix ready and will send a PR shortly.

#### Environment:

- Container runtime or hardware configuration: not runtime specific, `crictl ... jsonschema` and config parsing only
- OS (e.g: `cat /etc/os-release`): not OS specific
- Kernel (e.g. `uname -a`): not kernel specific
- Others: cri-tools `master` at 7e2aa2ec, `k8s.io/cri-api` v0.36.3, go1.26.4

locker95:
#### What type of PR is this?

/kind bug

`crictl create jsonschema --container` and `crictl run jsonschema --container` publish a machine readable schema that contradicts crictl's own config parser: the schema says `envs[].value` is base64, the parser accepts plain text.

**Evidence 1 — what the schema says on `master`:**

```console
$ go build -o ./crictl ./cmd/crictl
$ ./crictl create jsonschema --container | grep -A4 '"value"'
      "value": {
        "type": "string",
        "contentEncoding": "base64"
      }
    },
```

**Evidence 2 — what the parser actually does on `master`** (`loadContainerConfig` on a config with a single env var):

```
config value: "hello"      ->  parsed env value: "hello"
config value: "aGVsbG8="   ->  parsed env value: "aGVsbG8="     # expected "hello"
```

Base64 is not rejected, it is taken literally. So a user or a code generator that trusts the published schema and encodes the value gets a container whose env var contains the base64 text, with no error — which is exactly the bad path #2127 complains about.

Root cause: `KeyValue` regained a plain string `value` encoding through the custom `MarshalJSON`/`UnmarshalJSON` added upstream in kubernetes/kubernetes#139964, which landed here with the `k8s.io/cri-api` v0.36.3 bump in 895f236b (#2152). Those methods only affect the `encoding/json` path used by `sigs.k8s.io/yaml` when parsing configs. `printJSONSchema` goes through `github.com/invopop/jsonschema`, which reflects over the Go struct and still renders the underlying `[]byte` as `contentEncoding: base64`. The parser side was fixed, the reflection side was missed.

This PR removes `contentEncoding` from the reflected `KeyValue` definition, so the schema again matches what the parser accepts. The rest of the schema is untouched — the only diff in `crictl create jsonschema --container` output is:

```diff
       "value": {
-        "type": "string",
-        "contentEncoding": "base64"
+        "type": "string"
       }
```

Fixes #2153
Relates to #2127

**Scope.** Only the reflection based `jsonschema` output is changed. There is a second base64 producing path in `protobufObjectToJSON` (`protojson`), but `KeyValue` is input-only for crictl — it appears solely in `ContainerConfig.envs`, which crictl never `protojson`-marshals — so that path has no user visible problem here and is deliberately left alone.

**Implementation choice.** `invopop/jsonschema` offers `Reflector.Mapper` for type specific overrides, but a `Mapper` hit returns before the type is registered, which inlines `KeyValue` at its use site and drops the `#/$defs/KeyValue` entry — a needless, larger change to a published artifact that code generators consume. Clearing `ContentEncoding` on the reflected definition (all public API: `Schema.Definitions`, `Properties.Get`, `Schema.ContentEncoding`) keeps `$defs` intact and reduces the output diff to the two lines above. Happy to switch to `Mapper` if you prefer that even with the `$defs` change.

**Release impact.** The mismatch is not in v1.36.0, which pins cri-api v0.36.0 where the parser and the schema agree on base64. It only exists on `master` since 895f236b, so this lands before it can ship.

**Testing.**

- New `TestPrintJSONSchemaEnvValueIsPlainString` in `cmd/crictl/util_test.go` asserts the printed schema describes `KeyValue.value` as a plain string with no `contentEncoding`. It fails on `master` and passes with this change:

  ```console
  # before (fix reverted)
  --- FAIL: TestPrintJSONSchemaEnvValueIsPlainString (0.00s)
      util_test.go:454: expected empty content encoding for KeyValue value, got "base64"

  # after
  --- PASS: TestPrintJSONSchemaEnvValueIsPlainString (0.04s)
  ```

- `go test ./cmd/crictl/... -count=1`, `go build ./...`, `gofmt`, and `golangci-lint run ./cmd/crictl/...` (for both the host GOOS and `GOOS=linux`) are clean.
- `critest` and the `crictl` e2e suites were not run locally: they need a live containerd/CRI-O, which this change does not touch.

/cc @SergeyKanzhelev @bitoku

```release-note
crictl now describes container environment variable values as plain strings in the output of `crictl create jsonschema --container` and `crictl run jsonschema --container`, instead of incorrectly advertising them as base64 encoded.
```

### DIFF
--- a/cmd/crictl/util.go
+++ b/cmd/crictl/util.go
@@ -235,8 +235,29 @@ func loadContainerConfig(path string) (*pb.ContainerConfig, error) {
 	return &config, nil
 }
 
+// stripKeyValueContentEncoding removes the base64 content encoding from the
+// reflected pb.KeyValue schema.
+//
+// pb.KeyValue stores the value as []byte, which reflection renders as a base64
+// encoded string. The CRI API implements custom JSON marshaling for
+// pb.KeyValue to keep encoding the value as a plain string, which is what the
+// config parsers in this package accept as well. Advertising base64 in the
+// schema would point users and code generators to an encoding crictl does not
+// understand.
+func stripKeyValueContentEncoding(schema *jsonschema.Schema) {
+	definition, ok := schema.Definitions[reflect.TypeFor[pb.KeyValue]().Name()]
+	if !ok || definition.Properties == nil {
+		return
+	}
+
+	if value, ok := definition.Properties.Get("value"); ok {
+		value.ContentEncoding = ""
+	}
+}
+
 func printJSONSchema(value any) error {
 	schema := jsonschema.Reflect(value)
+	stripKeyValueContentEncoding(schema)
 
 	data, err := json.MarshalIndent(schema, "", "  ")
 	if err != nil {
--- a/cmd/crictl/util.go
+++ b/cmd/crictl/util.go
@@ -236,14 +236,8 @@ func loadContainerConfig(path string) (*pb.ContainerConfig, error) {
 }
 
 // stripKeyValueContentEncoding removes the base64 content encoding from the
-// reflected pb.KeyValue schema.
-//
-// pb.KeyValue stores the value as []byte, which reflection renders as a base64
-// encoded string. The CRI API implements custom JSON marshaling for
-// pb.KeyValue to keep encoding the value as a plain string, which is what the
-// config parsers in this package accept as well. Advertising base64 in the
-// schema would point users and code generators to an encoding crictl does not
-// understand.
+// reflected pb.KeyValue schema so it matches the custom JSON marshaling that
+// keeps values as plain strings.
 func stripKeyValueContentEncoding(schema *jsonschema.Schema) {
 	definition, ok := schema.Definitions[reflect.TypeFor[pb.KeyValue]().Name()]
 	if !ok || definition.Properties == nil {
--- a/cmd/crictl/util_test.go
+++ b/cmd/crictl/util_test.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
 	"io"
@@ -27,6 +28,7 @@ import (
 
 	. "github.com/onsi/gomega"
 	"github.com/urfave/cli/v2"
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
 func TestGetSortedKeys(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cmd/crictl/util.go
```diff
@@ -235,8 +235,29 @@ func loadContainerConfig(path string) (*pb.ContainerConfig, error) {
...
 
+// stripKeyValueContentEncoding removes the base64 content encoding from the
+// reflected pb.KeyValue schema.
+//
+// pb.KeyValue stores the value as []byte, which reflection renders as a base64
+// encoded string. The CRI API implements custom JSON marshaling for
+// pb.KeyValue to keep encoding the value as a plain string, which is what the
+// config parsers in this package accept as well. Advertising base64 in the
+// schema would point users and code generators to an encoding crictl does not
+// understand.
```
saschagrunert: Nit: the comment is 9 lines for a single idea. The godoc summary is enough; the rationale is well covered in the PR description and commit message.

```suggestion
// stripKeyValueContentEncoding removes the base64 content encoding from the
// reflected pb.KeyValue schema so it matches the custom JSON marshaling that
// keeps values as plain strings.
```

saschagrunert:
LGTM other than the inline comment.

locker95:
trimmed the godoc down to the summary as suggested

saschagrunert (approved):

PR #2159, opened 2026-08-08
crictl: allow config to create a missing --config file

### LINKED ISSUE #2143: crictl config fails to create a missing explicit config file

locker95:
`crictl --config /path/to/new.yaml config --set debug=true` used to die in `app.Before` with `get server config` when the file was not there yet. The config subcommand already knows how to create the file; this just lets that path through when the error is `ErrNotExist`.

Other commands still require an existing `--config` path.

Fixes #2143

### DIFF
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -216,9 +216,8 @@ func run() error {
 
 		if config, err = common.GetServerConfigFromFile(context.String("config"), exePath); err != nil {
 			if context.IsSet("config") {
-				// `crictl config` can create a missing file; allow that path through.
-				// Other subcommands still require an existing --config path.
-				isConfigCmd := context.Args().Present() && context.Args().First() == "config"
+				// crictl config can create a missing file; let it through.
+				isConfigCmd := context.Args().First() == "config"
 				if !isConfigCmd || !errors.Is(err, os.ErrNotExist) {
 					return fmt.Errorf("get server config: %w", err)
 				}
--- a/cmd/crictl/main.go
+++ b/cmd/crictl/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


saschagrunert (commented):
Two small nits, see inline suggestion.

cmd/crictl/main.go
```diff
@@ -215,7 +216,12 @@ func run() error {
 
 		if config, err = common.GetServerConfigFromFile(context.String("config"), exePath); err != nil {
 			if context.IsSet("config") {
-				return fmt.Errorf("get server config: %w", err)
+				// `crictl config` can create a missing file; allow that path through.
+				// Other subcommands still require an existing --config path.
+				isConfigCmd := context.Args().Present() && context.Args().First() == "config"
```
saschagrunert: Nit: the second comment line restates the `!isConfigCmd` condition below, and `context.Args().Present()` is redundant since `First()` returns `""` on empty args (see `args.Get(0)` in urfave/cli), so `"" == "config"` is already false.

```suggestion
				// crictl config can create a missing file; let it through.
				isConfigCmd := context.Args().First() == "config"
```

locker95:
took the nit — dropped the redundant Present() check and the extra comment line.

saschagrunert (approved):

PR #2181, opened 2026-09-10
fix: add eventual polling on inspection test to avoid race condition

### LINKED ISSUE #2179: "crictl: inspect" test fails because `crictl run` doesnt wait for container to exit
#### What happened:

The `crictl: inspect` "should succeed" e2e test intermittently fails with the following error: 
```[FAILED] Expected
<string>:
(a full container JSON dump)
to contain substring
<string>: nothing found per filter
In [It] at: test/e2e/inspect_test.go:71
```

In `test/e2e/inspect_test.go`, the `BeforeEach` starts a container via `crictl run` using the default entrypoint (`sh`, no stdin/tty). The `It("should succeed")` test then immediately asserts that a plain `crictl inspect` (without `--all`) finds nothing, with the comment "since container is exited".

However, `RunContainer` (cmd/crictl/container.go) only creates the sandbox, creates the container, and calls `StartContainer`. It never waits for the container process itself to exit, so the very next `inspect` call races against the container actually transitioning from `CONTAINER_RUNNING` to `CONTAINER_EXITED`. On a busy CI runner the container is sometimes still `CONTAINER_RUNNING` a few dozen ms later, so the default `inspect` finds and prints it instead of returning "nothing found per filter", failing the assertion.

This test should pass with the rest of the e2e tests consistently. 

Since this is timing dependent, it's not a consistently reproducible thing, but I have found it happens about 1/3 of the time running `make test-e2e GINKGO_FOCUS="crictl: inspect"`
Suggested fix: Poll/`Eventually` for the container to actually reach `CONTAINER_EXITED` before asserting `inspect` finds nothing 
Happy to raise the PR myself if maintainers agree with it.

#### Environment:

- run on: GitHub Actions hosted `ubuntu-22.04` runner (2 vCPU / 7GB, standard GH-hosted spec)
- OS (e.g: `cat /etc/os-release`): ubuntu 22.04.5
- containerd: 2.3.4
- Kernel (e.g. `uname -a`): 
- Others:

ethandcosta:
#### What type of PR is this?
/kind bug
/kind flake

This PR adds an eventual polling to the check that the container exited in the `inspect-test` to prevent a flake that can happen due to a race condition where the container hasnt finished exiting when the empty inspection test is asserted. 

Fixes https://github.com/kubernetes-sigs/cri-tools/issues/2179

```release-note

```

```release-note
NONE
```


ethandcosta:
Hey @tallclair @derekwaynecarr @SergeyKanzhelev any thoughts on the changes in this PR? should be a relatively small fix and wont cause feature changes.

PR #2185, opened 2026-09-18
Remove containerd 1.7 support (EOL)

SergeyKanzhelev:
containerd 1.7 has reached end of life. Remove it from the CI matrix, drop all version-specific workarounds (NRI gating, CNI install branching, image_volume.go skip paths), and enable NRI unconditionally for all supported containerd versions (2.x+).

/kind cleanup

k/k dropped testing of 1.7 and we should do the same.

/cc @samuelkarp @mikebrow 

```release-note
Containerd 1.7 is not tested any longer. Users of containerd 1.7 should use older releases for compatibility.
```


samuelkarp:
> containerd 1.7 has reached end of life

Almost, not quite. End of the month.

samuelkarp:
Can this wait until either the end of September or until we have changed the status on https://containerd.io/releases/#current-state-of-containerd-releases to "End of Life"?

SergeyKanzhelev:
> Can this wait until either the end of September or until we have changed the status on https://containerd.io/releases/#current-state-of-containerd-releases to "End of Life"?

Do you expect the EOL to change? We already removed it from k/k. Doing the same here early in preparation for 1.38.

I am not feeling string about it, but I do not see many reasons to wait

samuelkarp:
> Do you expect the EOL to change? We already removed it from k/k.

No, but we're still cherry-picking a few patches to 1.7 and we depend on critest in containerd's CI.

SergeyKanzhelev:
makes sense.

/hold

mikebrow:
note.. the 1.7 branch of containerd is way back on cri-tools version 1.29 https://github.com/containerd/containerd/blame/release/1.7/script/setup/critools-version

I don't think we'll be migrating up as that would also require migrating up to a later cri version.

